PR #3, opened 2026-08-05
test: merge policy (should be rejected)

lenoxys:
Temporary. Verifying that a non-beta branch is refused.


lenoxys:
Verified: the merge policy refuses a non-beta source branch. Closing the test.

PR #7, opened 2026-08-07
setPollInterval(): change the cloud poll interval at runtime

max246:
`pollMs` was a constructor-only option (default 600000 = 10 min, `0` disables). A host that lets the user tune "how often to poll the cloud" needs to change it **after login**, without rebuilding the client.

## Change
- **`setPollInterval(ms)`** — updates `opts.pollMs`, cancels the pending tick and re-arms `schedulePoll()` at the new interval; `0` disables (leaves the loop cancelled). Takes effect immediately.
- **`pollIntervalMs`** getter — the effective value (configured `pollMs` or `DEFAULT_POLL_MS`), so a host can read it back to display the current setting.

`schedulePoll()` already re-reads `opts.pollMs` on every re-arm, so a mutation is naturally picked up next cycle; the explicit `cancel()` + reschedule just makes it immediate.

## Why
Backs a bridge/HA feature: exposing the poll cadence as a Home Assistant option so the user can trade freshness vs cloud traffic from the UI.

Builds clean (`tsc`).


lenoxys (approved):
lgtm

PR #8, opened 2026-08-07
The property manifest honours per-member `available`

max246:
## Problem
A HomeBase was advertising camera-only params (`microphone`, `speaker`, `audioRecording`) it can't use, and a camera was advertising the hub-only `hubAlarmTone`.

`propertiesOf()` runs at **module load** with no device context, so each capability's property list carries **every** member regardless of its `available` gate. The `audio` capability is shared (a hub has it for alarm/prompt volume), so the hub inherited the camera-only members. The `available` predicate was only ever applied to the fluent getter/setter surface — never to the manifest a host (bridge / HA) builds entities from.

## Fix
- **`mergeProperties(caps, ctx?)`** — with a context, drop any property whose backing member has an `available` gate that fails for it, cross-referenced from the module's `members` by property name (`m.property ?? key`). A throwing gate keeps the property (fail-open). No `ctx` → no filtering (back-compat, existing callers/tests unaffected).
- **`resolveDevice`** — builds a lightweight resolve-time context (`codec`, `deviceType`, `model`, `category`, `capabilities`) and passes it. The `available` predicates only read those (`isCameraCodec`→codec, `isHomeBase`→deviceType, `hasCapability`→capabilities).

## Result (verified)
- Camera → `microphone`, `speaker`, `speakerVolume`, `audioRecording` (no `hubAlarmTone`).
- HomeBase → `hubAlarmTone` (no mic/speaker/recording).
- No context → everything (back-compat).

## Tests
Added camera-vs-hub-vs-no-context coverage to `mergeProperties`. Full suite: **2018 passed, 1 skipped**, builds clean.

Fixes the HA-side reports of a HomeBase showing mic/speaker/audio-recording controls.


src/model/capabilities/index.ts
```diff
@@ -158,21 +158,46 @@ export const CAPABILITY_MODULES = Object.fromEntries(MODULES.map((m) => [m.capab
...
   const merged: PropertySpec[] = [];
   for (const cap of caps) {
     const module = getCapabilityModule(cap);
     if (!module) continue;
     for (const spec of module.properties) {
       if (seen.has(spec.name)) continue;
+      // A property whose member is family-gated (`available`) belongs on the manifest only where
+      // the gate holds — otherwise a shared capability leaks camera-only params onto a HomeBase
+      // (e.g. `audio` gives the hub its alarm volume, but not `microphone`/`speaker`).
+      if (ctx && !memberAvailable(module.members, spec.name, ctx)) continue;
```
lenoxys: This adds a second interpreter for the `members` table by reverse-joining `module.properties` to members by name. That can drift from `propertiesOf()` and `bindMembers()`, and `bindMembers()` currently does not apply `available` to getters at all. Please derive the contextual property list directly in `members.ts` so the getter, setter, and manifest use one availability decision.

src/model/capabilities/index.ts
```diff
@@ -158,21 +158,46 @@ export const CAPABILITY_MODULES = Object.fromEntries(MODULES.map((m) => [m.capab
...
+  if (!members) return true;
+  for (const [key, m] of Object.entries(members)) {
+    const name = "property" in m && m.property ? m.property : key;
+    if (name !== propName) continue;
+    const gate = "available" in m ? m.available : undefined;
+    if (!gate) return true;
+    try {
+      return gate(ctx);
+    } catch {
+      return true;
```
lenoxys: Please do not fail open here. `installs()` invokes the same predicate without catching, so a thrown gate can advertise a property in the manifest and then fail while binding the action surface. Let the error propagate, or define one shared failure policy used by every projection.

src/model/registry.ts
```diff
@@ -210,7 +210,19 @@ export function resolveDevice(rec: CloudRecord): ResolvedDevice {
...
+  const gateCtx = {
+    channel: 0,
+    codec,
+    deviceType: rec.deviceType,
+    model: rec.model,
+    category: rec.category,
+    capabilities: new Set(capabilities),
+    paramIds: new Set<number>(),
+  } as unknown as CommandContext;
+  const properties = mergeProperties(capabilities, gateCtx);
```
lenoxys: `as unknown as CommandContext` is hiding the wrong abstraction: this is not a command context, and `channel: 0` plus an empty `paramIds` set are fabricated values. Please introduce or use a narrower resolution/availability context whose required fields can be populated truthfully from `CloudRecord`.

src/model/registry.ts
```diff
@@ -210,7 +210,19 @@ export function resolveDevice(rec: CloudRecord): ResolvedDevice {
...
```
lenoxys, line 214: Making properties depend on `deviceType` and `category` introduces a re-resolution bug: `Device.reresolve()` only adopts the next resolution when capabilities are gained. If one of these fields changes while the capability set stays the same, the old property manifest remains. Please cover and update that path so contextual properties are adopted independently of capability growth.

src/model/__tests__/model.spec.ts
```diff
@@ -331,4 +331,27 @@ describe("mergeProperties — shared props dedupe across capabilities", () => {
     expect(detectCapabilities({ params: { 1550: "1" } }, "sensor")).not.toContain("video");
     expect(detectCapabilities({ model: "cam" }, "camera")).toContain("video");
   });
+
+  it("honours per-member `available` — a hub doesn't inherit camera-only audio props", () => {
+    const names = (ctx: object | undefined) => mergeProperties(["audio"], ctx as never).map((p) => p.name);
```
lenoxys: Please test the real resolution path (`resolveDevice()` or `Device.fromRecord()`) rather than bypassing it with `ctx as never`. The camera assertions should also cover `speakerVolume` (the renamed `property` case) and `audioRecording`, since both are part of the claimed result.

src/model/__tests__/model.spec.ts
```diff
@@ -331,4 +331,27 @@ describe("mergeProperties — shared props dedupe across capabilities", () => {
...
+
+    const hubNames = names(hub);
+    expect(hubNames).not.toContain("microphone");
+    expect(hubNames).not.toContain("speaker");
+    expect(hubNames).not.toContain("audioRecording");
+    expect(hubNames).toContain("hubAlarmTone");
+  });
+
+  it("without a context, nothing is gated (back-compat)", () => {
+    const all = mergeProperties(["audio"]).map((p) => p.name);
```
lenoxys: `mergeProperties()` is internal and the package is pre-1.0, so preserving the known-invalid unfiltered behavior is not a compatibility requirement. Please make the context explicit for device manifests instead of locking this fallback in with a test.

lenoxys (changes requested):
Please address the inline comments.

max246:
Reworked per the review — all six points addressed (pushed `b57857e`; the `enumValuesFor`/workingMode work from #10 is folded in since it rides the same mechanism, so there aren't two interpreters again).

1. **Second interpreter (reverse-join in index.ts)** → gone. `propertiesOf(members, ctx?)` is now the single interpreter: it applies each member's `available` gate and resolves `enumValuesFor` straight from the table. `mergeProperties` calls it, `bindMembers` gates getters on the same `available`, and `installs()` gates setters — one decision, three surfaces.
2. **Fail-open** → removed. A throwing gate propagates, same policy as `installs()`; no try/catch swallowing.
3. **`as unknown as CommandContext`** → replaced with a real `AvailabilityContext` (codec/deviceType/model/category/capabilities; `hasP2p` optional for the command path). `registry.resolveProperties()` is the one place it's built, populated truthfully from the `CloudRecord` — no fabricated `channel`/`paramIds`.
4. **`reresolve` adoption bug** → `Device.reresolve` recomputes the manifest for the UNION capability set and adopts when it changed even with no capability gained (and no longer drops a retained capability's props). Covered by a category-flip test (same caps, different manifest).
5. **Tests bypassing resolution** → rewritten to go through `resolveDevice` / `Device.fromRecord` (no `ctx as never`); the camera case now asserts `speakerVolume` and `audioRecording` too.
6. **Locking in the unfiltered fallback** → that test is gone; the device-manifest path always passes a context.

`npm run verify` green (2019 passed, 1 skipped). Ready for another look.

src/model/capabilities/index.ts
```diff
@@ -158,13 +159,18 @@ export const CAPABILITY_MODULES = Object.fromEntries(MODULES.map((m) => [m.capab
...
   const merged: PropertySpec[] = [];
   for (const cap of caps) {
     const module = getCapabilityModule(cap);
     if (!module) continue;
-    for (const spec of module.properties) {
+    // With a device context, re-derive the property list from the SAME `propertiesOf` the module's
+    // static list comes from — now applying that device's `available` gates and per-model enums. This
+    // keeps one interpreter of the members table (getter, setter and manifest agree); without a
+    // context the precomputed static list is used unchanged.
+    const props = ctx && module.members ? propertiesOf(module.members, ctx) : module.properties;
```
lenoxys, line 172: A read-only A/B run against the same live fleet confirms this PR fixes the intended `available` filtering. Only aggregate device classes and counts were emitted; no serials, names, account details, or other identifiers were logged.

```text
                         base 17568c2   PR 4652ffd
fleet devices                     14            14
audio cameras                      7             7
camera hubAlarmTone leaks          7             0
HomeBase camera-property leaks     1             0
HomeBase with hubAlarmTone          1             1
workingMode manifest mismatches    4             4
```

The audio regression is fixed. The four `workingMode` mismatches are unchanged from the base: battery-capable sensors without reported param `1246` still advertise `workingMode` in `Device.properties` while omitting the bound getter. That is a separate `requires`-filtering gap rather than a regression in this PR, so it should be handled as follow-up work.

lenoxys (approved):
Approved. A read-only A/B run against the same 14-device fleet confirms the intended family-gated manifest fix: camera hubAlarmTone leaks fell from 7 to 0, and HomeBase camera-property leaks fell from 1 to 0. No PII was logged. The existing workingMode/requires mismatch was unchanged from the base and is follow-up work, not a regression in this PR.

src/model/capabilities/index.ts
max246: Good catch from the live run — the manifest/getter gap on `workingMode` is real. But it can't be fixed by gating `propertiesOf`/`Device.properties` directly, because that list does double duty and the naive gate breaks both jobs (I tried it):

1. **Decode** — `Device` builds `specByName`/`byParam` from `resolved.properties`. Reads-gating a property out at resolve time means an inbound realtime param for it can't be decoded (no spec) — it lands as `unknown_…`. `raw_dp_reads.spec` ("a realtime report reaches the typed getters") goes red.
2. **Schema completeness** — `resolveDevice().properties` is asserted to be the full *capability* schema regardless of reported params: `vacuum battery is DP 163` and `doorbell … 1702-1719` both `resolveDevice({model})` with **no params** and expect those specs present.

So `Device.properties` is "what this model *can* report" (feeds decode), while the getter is the "what it *has* reported" evidence gate — a superset/subset by design, not a contradiction. The thing a host over-builds from is a *host-facing manifest*, which today is just `Device.properties` unfiltered.

Two clean fixes — happy to do whichever you prefer:

- **A (my rec):** keep `resolved.properties` full for decode, and apply the reported-param + `available` gate to the **host-facing enumeration** (`describeCapabilities` / a `describe()`-level manifest) so what a host consumes matches the installed getters. No decode regression; existing schema tests stand.
- **B:** make `Device.properties` itself the gated view and decouple decode onto a separate full spec table (`mergeProperties(caps)` with no ctx), updating the schema-completeness tests to assert against that full table.

Which direction do you want? The `available`-family gating in this PR is unaffected either way.

PR #9, opened 2026-08-07
feat(media): complete fragmented recordings

lenoxys:
## Summary

- Carry station-declared audio metadata and mux AAC-LC/AAC-ELD tracks into fragmented recordings.
- Drain keyframe-aligned audio/video prebuffer with capture-time continuity.
- Expose battery-budget extension and explicit stop through a caller-owned recording handle.
- Keep raw readable output video-only and document keyframe-bounded fragment timing.

## Validation

- `npm run verify`: 2028 passed, 1 skipped.
- Read-only fleet validation covered every camera path, AAC-LC audio, and silent video-only sources.


max246 (commented):
Reviewed the media pass. The audio muxing, capture-time timeline alignment, and the caller-owned `FragmentRecording` handle are well-structured, and the test coverage (capture-time durations, prebuffer replay, budget forwarding, ADTS strip, video-only fallback) is genuinely thorough. A few things worth addressing before merge — inline. The main one: `pushAudio` trusts that the ADTS header parses but never checks its declared sample-rate/channels, so a non-16 kHz or stereo source would be muxed as 16 kHz mono (wrong pitch/speed) even though `isSupportedAdts()` already exists for this. The rest are minor / nits.

src/transport/p2p/fmp4.ts
```diff
@@ -51,63 +55,131 @@ export class Fmp4Muxer {
...
+      this.lastAudioTimestampMs === undefined
+        ? 0
+        : Math.round((timestampMs - this.lastAudioTimestampMs) * (AAC_SAMPLE_RATE / 1000));
+    this.lastAudioTimestampMs = timestampMs;
+    if (this.timelineAligned && this.audioSamples.length === 0 && this.timelineOriginMs !== undefined) {
+      this.audioBaseDecodeTime = Math.round((timestampMs - this.timelineOriginMs) * (AAC_SAMPLE_RATE / 1000));
+    } else if (measured > 0 && this.audioSamples.length) {
+      this.audioSamples[this.audioSamples.length - 1].duration = measured;
+    }
+    this.audioSamples.push({
```
max246, line 168: The muxer accepts any parseable ADTS frame but never checks the header's `frequencyIndex` / `channels` — it always writes a 16 kHz-mono `esds` + audio timescale. A source at 44.1 kHz or stereo would be muxed as 16 kHz mono → wrong pitch/speed. `adts.ts` already has `isSupportedAdts()` (AAC-LC / 16 k / mono) for exactly this; since it's LC-only and you also support ELD, consider a param check here (`frequencyIndex === FREQ_INDEX_16K && channels === CHANNELS_MONO`) and disabling audio (like the g711a path) when it doesn't match, rather than silently mis-muxing.

Also: AAC-ELD frames are 512 samples, not `AAC_SAMPLES_PER_FRAME` (1024). The measured-timing correction hides most of it, but the last sample of each fragment keeps the 1024 default → a small per-fragment tail error for ELD.

src/transport/p2p/fmp4.ts
```diff
@@ -51,63 +55,131 @@ export class Fmp4Muxer {
...
+    if (!this.audioRequested || this.audioDisabled) return undefined;
+    if (frame.codec === "g711a") {
+      if (this.initSent) return undefined;
+      this.audioDisabled = true;
+      if (!this.params) return undefined;
+      this.initSent = true;
+      return { init: this.buildInit(), data: Buffer.alloc(0), keyframe: false };
+    }
+    if (this.audioCodec && this.audioCodec !== frame.codec) {
+      throw new Error(`fMP4 muxer: audio codec changed from ${this.audioCodec} to ${frame.codec}`);
```
max246: Throwing here tears down the whole recording (`FragmentRecording.ingest` catches it → `fail()` → ends). But the `AudioCodec` doc + `live-media.md` say the station may re-declare the codec per frame, so an AAC-LC ↔ ELD flip mid-stream would kill an otherwise-fine recording. Consider disabling audio and continuing video-only (as the g711a branch does) instead of throwing — or, if the throw is intended, document that a mid-stream AAC-profile switch deliberately ends the recording.

src/transport/p2p/fmp4.ts
```diff
@@ -355,12 +514,20 @@ function annexbToAvcc(annexb: Buffer): Buffer {
 }
 
 /** Locate the trun's data_offset field inside a freshly built moof and set it. */
-function patchTrunDataOffset(moof: Buffer, offset: number): void {
+function patchTrunDataOffsets(moof: Buffer, offsets: number[]): void {
```
max246, line 532: Nit: locating each `trun` via `moof.indexOf('trun')` can false-match the bytes `74 72 75 6e` inside a preceding trun's sample table (durations/sizes/flags). Low probability, but now that there can be two truns the risk is real (a wrong `data_offset` would corrupt the fragment). Walking the moof boxes by size (moof→traf→trun) would be robust.

src/transport/p2p/index.ts
```diff
@@ -2,14 +2,7 @@ export * from "./p2p-session.js";
...
 export * from "./media.js";
-// `Consumer` is what `SharedLiveSource.attach` returns, so a caller has to be able to name it —
-// exported under `LiveConsumer`, since `Consumer` alone is too generic to claim at the root barrel.
-export {
-  SharedLiveSource,
-  type SharedLiveState,
-  type SharedLiveSourceOptions,
-  type Consumer as LiveConsumer,
-} from "./shared-live-source.js";
+export * from "./shared-live-source.js";
```
max246: Switching to `export *` re-exposes the generic `Consumer` at the root barrel and drops the deliberate `LiveConsumer` alias — a breaking rename for anyone importing `LiveConsumer`, and it re-claims the generic name the previous curated export intentionally avoided (per the comment removed here). Suggest keeping the explicit re-export block and just adding the newly-needed `TimedMediaFrame` (and `attachWithPrebuffer`'s return type if a caller needs to name it).

src/transport/p2p/fragment-recording.ts
```diff
@@ -0,0 +1,110 @@
...
+    }
+  }
+
+  private async attach(sourcePromise: Promise<SharedLiveSource>): Promise<void> {
+    try {
+      const source = await sourcePromise;
+      if (this.ended) return;
+      const attached = source.attachWithPrebuffer(this.opts.preBufferSeconds ?? 0);
+      this.consumer = attached.consumer;
+      this.consumer.on("video", (frame) => this.ingest({ kind: "video", frame, timestampMs: Date.now() }));
```
max246: Minor: live frames are stamped with `Date.now()` at ingest, while buffered frames carry the source's ring arrival time. This is only correct because the consumer emits synchronously when the source delivers (ingest time ≈ arrival time). Worth a one-line comment noting that assumption — or thread the arrival timestamp through the consumer's `video`/`audio` event so live and buffered use the same clock.

max246:
Re-reviewed after `927a98db` (address recording review feedback) — all five points from my earlier review are genuinely resolved, not just marked:

1. **ADTS params** — `pushAudio` now checks `frequencyIndex`/`channels` (`ADTS_FREQUENCY_INDEX_16K` / `ADTS_CHANNELS_MONO`) and `disableAudio()`s a non-16 kHz / non-mono source instead of muxing it as 16 kHz mono. ✅
2. **Mid-stream codec change** — no longer throws; `return this.disableAudio()` keeps the recording alive video-only. ✅
3. **`patchTrunDataOffsets`** — walks the moof boxes by size now, so it can't false-match `trun` bytes inside a sample table. ✅
4. **`index.ts`** — curated re-export restored (`type Consumer as LiveConsumer` + `TimedMediaFrame`); no more `export *` re-claiming the generic name. ✅
5. **`fragment-recording`** — live and buffered frames both carry `item.timestampMs`, so one clock instead of `Date.now()` at ingest. ✅

LGTM — nice, thorough follow-ups. 👍

max246 (approved):

PR #10, opened 2026-08-07
Per-device enum options (enumValuesFor) — workingMode as a dropdown

max246:
## Problem
`workingMode` shows up in a host (Home Assistant) as a **free-text/number field**, not a picker — even though it's a fixed set of modes. The set couldn't be a static `enumValues` because **models number the modes differently** (a 3-mode camera vs the doorbell's 4-mode set), so the property shipped as a bare `number`.

## Fix
A general **per-device enum** mechanism, reusing the resolve-time `ctx` that PR #8 (`fix/manifest-honors-available`, the base of this PR) added to `mergeProperties`:

- **`ValueMember.enumValuesFor?(ctx)`** — resolve a property's enum options from the device context when a static `enumValues` can't state them.
- **`mergeProperties(caps, ctx)`** now resolves `enumValuesFor` against the context and stamps the result onto that device's spec (a shared `memberFor()` lookup replaces `memberAvailable()`; a throwing hook falls back safely).
- **`battery.workingMode`** gains `enumValuesFor: (ctx) => workingModeMap(ctx.model)`. `workingModeMap()` is extracted from the existing per-model logic and reused by both resolvers (DRY). `kind` stays `"scalar"` (the static spec has no set to ship — the `value-kinds` guard requires one for `kind:"enum"`); the injected `enumValues` drive the host's picker regardless of `kind`.

## Result (verified)
`describe()` / `device.properties` now carries `workingMode`'s real per-model options:
- `T8114` → `{0: Optimal Battery Life, 1: Optimal Surveillance, 2: Customize Recording}`
- `T8214` → `{0: Balance Surveillance, 1: Optimal Surveillance, 2: Customize Recording, 3: Optimal Battery Life}`
- no context → no enum stamped (back-compat)

The HA integration already renders any property with `enumValues` as a select (regardless of wire `type`), so `workingMode` becomes a proper dropdown with no integration change.

Tests added; full `npm run verify` green (**2019 passed, 1 skipped**).

> **Base:** this is stacked on `fix/manifest-honors-available` (#8) — it uses the `ctx` plumbing from that PR. Merge #8 first (or merge this into it).


lenoxys:
I think this should be generalized around a single concept: a device-resolved enum domain.

We already have a closely related case in `suction`: `SUCTION_LEVELS_BY_MODEL` selects the supported levels from `ctx.model`, and the bound capability exposes them as `supportedLevels`. The difference is that this information is not currently propagated into `PropertySpec` or `Device.describe()`.

Rather than resolving `enumValuesFor` only inside `mergeProperties()`, could we introduce one shared resolver, for example:

```ts
function resolveEnumValues(
  member: ValueMember,
  ctx: CommandContext,
): Readonly<Record<number, string>> | undefined {
  return member.enumValuesFor?.(ctx) ?? member.enumValues;
}
```

That resolver should be the single source used by:

- resolved `Device.properties`;
- `Device.describe()` read descriptors;
- setter argument descriptions;
- write-domain validation;
- runtime supported-option properties such as `suction.supportedLevels`, if retained.

This would cover both existing cases:

- `workingMode`: the raw index-to-label mapping varies by model;
- `suction.level`: labels are global, but the supported subset varies by model.

At the moment, PR #10 resolves the dynamic enum only in `mergeProperties()`. Consequently, `Device.describe()` still reads the static member metadata and does not receive the per-device options, despite the PR description saying that both surfaces carry them.

There are two other contract issues to address:

1. The resolved property still declares `type: "number"` and `kind: "scalar"` while publishing `enumValues`. Callers following the documented `type`/`kind` semantics may continue treating it as a numeric field. The resolved schema should describe this consistently as an enum, and the value-kind validation should recognize a context-resolved enum declaration.

2. `workingModeMap()` falls back to the three-mode `DEFAULT` map for an absent or unknown model. Once this map is published as the device supported domain, that fallback becomes a capability claim and may offer unsupported fire-and-forget values. Unknown models should return no resolved domain unless the default is independently supported by verified family evidence.

I would keep the generalization limited to context-resolved enum domains for now. Adding generic `minFor`, `maxFor`, or similar hooks would be speculative until another concrete use case exists.

lenoxys (changes requested):
Requesting changes before merge. The dynamic enum domain needs to be resolved consistently across both `Device.properties` and `Device.describe()`, rather than only in `mergeProperties()`. The resolved metadata must also keep `type`, `kind`, and `enumValues` semantically consistent, and unknown models must not be assigned an unverified default option set. See the detailed review comment for the proposed shared resolver and the existing `suction.supportedLevels` precedent.

PR #11, opened 2026-08-07
Per-device enum options (enumValuesFor) — workingMode as a dropdown

max246:
Re-created off `beta-0.1.0` now that the context-aware manifest (#8) is merged. (Supersedes the old #10, which got auto-closed when its commits were folded during that PR's rework — this is the clean standalone version.)

## What it does
A member can resolve its enum options from the device at manifest time, for options that are real but **vary by model** — so a single static `enumValues` can't state them.

- **`ValueMember.enumValuesFor?(ctx)`** — resolved in `propertiesOf` against the device context and stamped onto that device's spec (overrides any static `enumValues`).
- **`battery.workingMode`** gains `enumValuesFor: (ctx) => workingModeMap(ctx.model)`; `workingModeMap()` is extracted from the existing resolvers (DRY). `kind` stays `"scalar"` (the value-kinds guard wants a set for `kind:"enum"`); the injected `enumValues` drive a host's picker regardless of `kind`.

## Result
`describe()` / `device.properties` now carries `workingMode`'s real per-model options:
- 3-mode cameras → `Optimal Battery Life / Optimal Surveillance / Customize Recording`
- doorbell (T8214) → those + `Balance Surveillance`, numbered its own way

The HA integration already renders any property with `enumValues` as a select, so `workingMode` becomes a proper dropdown — no integration change.

## Scope note
Orthogonal to the reads-gating discussion on #8: this only adds options to `workingMode` where it already appears in the manifest; whether it appears on a device that doesn't report 1246 is that separate decision.

`npm run verify` green (2039 passed, 1 skipped).


max246:
Carried over and applied @lenoxys' review from the previous (folded) PR #10 — all three points, pushed in `b89892b`:

1. **One shared enum-domain resolver.** Added `resolveEnumValues(member, ctx)` = `enumValuesFor(ctx) ?? enumValues`, and it's now the single source used by **both** `propertiesOf` (`Device.properties`) **and** `describeBound`/`readDescriptor` (`Device.describe()` reads). `describe()` threads the device's own context (codec/model/capabilities), so the per-model options reach that surface too — the two no longer disagree. (I kept it to the enum domain only, per your note — no speculative `minFor`/`maxFor`.)
2. **type/kind consistency.** A context-resolved enum is now described as `kind: "enum"` on the resolved spec *and* the read descriptor, instead of publishing `enumValues` under a `scalar` kind. The static member stays `scalar` (so the value-kinds guard, which checks the static schema, is unaffected).
3. **No unverified default.** `publishedWorkingModeDomain()` only claims a domain it can stand behind — an explicit per-model map (e.g. T8214), or the 3-mode `DEFAULT` for the battery-**camera** family (the documented family default; the property only appears with the battery capability). An **unknown non-camera gets no domain** rather than an unverified set. The write-side `workingModeMap()` keeps its best-effort `DEFAULT` for name↔index only.

On the `suction.supportedLevels` precedent: `resolveEnumValues` is deliberately shaped to be the same resolver that path can adopt, but I left `suction` as-is in this PR to keep it scoped to the context-resolved enum — happy to fold `supportedLevels` onto it in a follow-up if you'd like.

Verified end-to-end: camera → `enum` + 3-mode set; doorbell → 4-mode; unknown non-camera → no domain; `describe()` read carries the same options. `npm run verify` green (2039 passed).

src/model/capabilities/battery.ts
```diff
@@ -210,7 +210,11 @@ export const BATTERY_MEMBERS = {
   workingMode: {
     param: BATTERY_PARAM.WORKING_MODE,
     type: "number",
+    // kind stays "scalar" on the STATIC member (it has no set to ship — the value-kinds guard wants
```
lenoxys: Please remove this inline implementation-history comment. Repository standards require declaration-level JSDoc containing current ground truth. The existing `workingMode` JSDoc also says labels are not published, which this change makes stale; update that JSDoc to describe the verified per-device domain instead.

src/model/capabilities/battery.ts
```diff
@@ -416,20 +420,41 @@ export const WORKING_MODE_MAPS: Readonly<Record<string, Readonly<Record<number,
  * by T-code prefix, falling back to the 3-mode camera `DEFAULT`.
  */
 export function resolveWorkingMode(model: string | undefined, value: number): string | undefined {
+  return workingModeMap(model)[value];
+}
+
+/**
+ * The working-mode {index → label} map for a device model — its own {@link WORKING_MODE_MAPS} entry
+ * (matched by T-code prefix) or the 3-mode camera `DEFAULT`. This is the per-device enum a host shows
```
lenoxys: Keep this as declaration-level JSDoc, but state only the function's behavior. Presentation and processing narrative such as a host showing a dropdown and `enumValuesFor` stamping the manifest does not belong in this map resolver's ground truth.

src/model/capabilities/battery.ts
```diff
@@ -416,20 +420,41 @@ export const WORKING_MODE_MAPS: Readonly<Record<string, Readonly<Record<number,
...
+ * a domain is a capability claim, so it is only made where the labels are trustworthy: an explicit
+ * per-model map, or the 3-mode `DEFAULT` for the battery-CAMERA family (the documented family default,
+ * and the property only appears with the battery capability). An unknown non-camera gets no domain,
+ * rather than an unverified set a host could offer as real options.
+ */
+function publishedWorkingModeDomain(ctx: AvailabilityContext): Readonly<Record<number, string>> | undefined {
+  const m = (ctx.model ?? "").toUpperCase();
+  const key = Object.keys(WORKING_MODE_MAPS).find((k) => k !== "DEFAULT" && m.startsWith(k));
+  if (key) return WORKING_MODE_MAPS[key];
+  return ctx.codec === "camera" ? WORKING_MODE_MAPS.DEFAULT : undefined;
```
lenoxys: **Spec: this publishes an unverified domain to non-battery cameras.** The fallback is gated only by `ctx.codec === "camera"`, not by a verified battery-camera family. I reproduced this with PR 11 using the active-fleet T8419 record: because T8419 reports `1101` and `1246`, both `device.properties` and `device.describe()` publish these three battery-mode labels. `mega_preview` identifies T8419 as an indoor pan-and-tilt camera and contains no model-specific evidence for that domain. Please restrict the fallback to verified models/families; otherwise return `undefined`.

src/model/capabilities/members.ts
```diff
@@ -433,6 +441,17 @@ export type Surface<M extends Members> = {
 
 // ── the derived RUNTIME ─────────────────────────────────────────────────────────────────────────
 
+/**
+ * The enum option set for a member on THIS device — the single source every surface reads (the
```
lenoxys: This JSDoc overstates the contract: `resolveEnumValues()` is not used by the setter, and “single source” / “so they never disagree” describes design intent rather than observable behavior. Please state only what the function does: return context-specific enum values when available, otherwise static enum values.

src/model/capabilities/members.ts
```diff
@@ -447,14 +466,20 @@ export function propertiesOf(members: Members, ctx?: AvailabilityContext): Prope
     // do — one decision per member, so a family-gated value (a hub's `microphone`) never appears on a
     // device that can't use it. A throwing gate propagates, same as on the command path.
     if (ctx && m.available && !m.available(ctx)) return [];
+    // A per-model enum (workingMode) is resolved from the context via the ONE shared resolver, so the
```
lenoxys: Please remove this inline implementation narrative and put any necessary contract in the JSDoc above `propertiesOf()`. Document current output semantics rather than the effort to make multiple surfaces agree.

src/model/capabilities/members.ts
```diff
@@ -447,14 +466,20 @@ export function propertiesOf(members: Members, ctx?: AvailabilityContext): Prope
...
+    // resolved spec agrees with describe() / the setter. When the resolved set came from the device
+    // (enumValuesFor), the spec is described consistently as an enum rather than the member's stored
+    // scalar kind.
+    const dynamicEnum = ctx ? m.enumValuesFor?.(ctx) : undefined;
```
lenoxys: `enumValuesFor` is evaluated here and then evaluated again inside `resolveEnumValues()`. A non-repeatable callback could make `kind` and `enumValues` disagree. Please resolve it once and derive both fields from that result. The same double evaluation occurs in `readDescriptor()`.

src/model/capabilities/manifest.ts
```diff
@@ -151,16 +152,26 @@ export function describeBound(
...
  */
-function readDescriptor(name: string, m: ValueMember, descriptors: Record<string, PropertyDescriptor>): ReadDescriptor {
+function readDescriptor(
+  name: string,
+  m: ValueMember,
+  descriptors: Record<string, PropertyDescriptor>,
+  ctx?: AvailabilityContext,
+): ReadDescriptor {
   const setter = m.writeAs ?? `set${name[0].toUpperCase()}${name.slice(1)}`;
+  // Resolve the device's enum options through the ONE shared resolver, so describe() agrees with the
```
lenoxys: Please remove this inline body comment. If the behavior needs documentation, put ground-truth JSDoc above `readDescriptor()`: the descriptor reports a resolved domain with enum semantics. Avoid iteration language such as “ONE shared resolver” and comparisons with other surfaces.

src/model/__tests__/model.spec.ts
```diff
@@ -348,6 +348,26 @@ describe("mergeProperties — shared props dedupe across capabilities", () => {
     }
   });
 
+  it("stamps per-model workingMode options through resolveDevice", () => {
```
lenoxys: This covers only `device.properties`, but the PR also changes `Device.describe()` and introduces an unknown-context fallback. Please add coverage proving that `describe()` returns the same model-specific values and that an unverified model does not receive a published domain.

src/model/__tests__/model.spec.ts
```diff
@@ -348,6 +348,26 @@ describe("mergeProperties — shared props dedupe across capabilities", () => {
...
+      resolveDevice({ deviceType: 9, model, params: { 1246: "0", 1101: "50" } }).properties.find(
+        (p) => p.name === "workingMode",
+      )?.enumValues;
+
+    expect(workingMode("T8114")).toEqual({
+      0: "Optimal Battery Life",
+      1: "Optimal Surveillance",
+      2: "Customize Recording",
+    });
+    // The doorbell numbers its modes differently and adds a fourth.
```
lenoxys: Please remove this inline test comment. Express the model-specific behavior in the test name or a separate named test rather than a body comment.

lenoxys (commented):
Review against the PR specification, repository documentation rules, and the committed active-fleet evidence in `eufy_sdk`.

lenoxys:
One additional Standards finding could not be attached inline because GitHub renders `src/model/device.ts` as binary (the file contains pre-existing literal NUL separators): at `device.ts:461`, please remove the new inline `//` explanation. `AGENTS.md:148-151` requires declaration-level, ground-truth JSDoc rather than body comments and implementation narrative. If this context forwarding needs documentation, put it above `describe()` and state only that the manifest is resolved with the device’s codec, model, and capabilities.

max246:
Thanks for the review — all points addressed in `2b2d973`.

**Correctness — unverified domain on non-battery cameras (`battery.ts:449`).** `publishedWorkingModeDomain` no longer falls back on `ctx.codec === "camera"`. It returns a domain only for a model's own `WORKING_MODE_MAPS` entry or a confirmed battery-camera family (`WORKING_MODE_DEFAULT_MODELS` = eufyCam `T8114`, SoloCam `T8124`/`T8170`/`T8171`), else `undefined`. The mains Indoor Cam `T8419` — which reports `1246`/`1101` — now publishes no domain, on both `device.properties` and `describe()`.

**Double evaluation of `enumValuesFor` (`members.ts:473`).** Added `resolvedEnum(m, ctx)` which evaluates the callback at most once and returns `{ values, dynamic }`; `propertiesOf()` and `readDescriptor()` derive both `enumValues` and the enum `kind` from that single result, so a non-repeatable callback can no longer make them disagree. The now-unused `resolveEnumValues()` was removed.

**JSDoc / inline narrative** (`battery.ts:213`, `:428`; `members.ts:445`, `:469`; `manifest.ts:162`). Removed the implementation-history / presentation comments and rewrote the declaration JSDoc on `resolvedEnum`, `propertiesOf`, `readDescriptor`, `workingModeMap`, `publishedWorkingModeDomain` and the `workingMode` member to state current behavior only. The stale "no labels are published" line on `workingMode` now describes the verified per-device domain.

**Tests** (`model.spec.ts:351`, `:362`). Split into named tests (no body comments); added `describe()` coverage proving it reports the same per-model domain as the property schema, and that an unverified model (`T8419`) receives no published domain.

Separately: `device.ts` was showing as a **binary file** in the diff — `sameManifest`'s `sig` used literal NUL bytes as delimiters (from `4652ffd`). Fixed here (JSON tuple key); since the base `beta-0.1.0` carries the same NULs, I opened #12 against it so this PR's `device.ts` diff renders as text once that merges.

`npm run verify` green (2043 tests).

src/model/device.ts
```diff
@@ -458,7 +458,13 @@ export class Device {
       source: this.source,
       bound: this.bound,
       capabilities: [...this.capabilities],
-      details: describeCapabilities(this.actionMap),
+      // Resolve per-device enum domains (e.g. workingMode) against this device's own facts, so the
```
lenoxys: no inline comment !!!!

lenoxys (approved):
lgtm

PR #12, opened 2026-08-07
fix: stop device.ts being a binary file (NUL delimiters)

max246:
## Problem

`src/model/device.ts` shows as a **binary file** in diffs (`Bin … bytes`, no line-level diff). `sameManifest`'s `sig` helper used two literal NUL (`\x00`) characters as field delimiters in a template string, so git classifies the whole file as binary and its changes become unreviewable. Introduced in `4652ffd` and inherited by every branch off `beta-0.1.0` (incl. PR #11).

## Fix

Replace the NUL-delimited string with `JSON.stringify([p.name, p.writable, p.enumValues ?? null])` — same comparison key, fully printable. `device.ts` is UTF-8 text again (`file` → *JavaScript source, UTF-8 text*; 0 NUL bytes).

Merging this first makes the PR #11 `device.ts` diff render as normal text.


max246:
Folded into #11 — that PR already carries the identical device.ts NUL-delimiter fix (commit 2b2d973), so keeping everything in a single PR. Note: with the base beta-0.1.0 still carrying the old NULs, #11's device.ts diff will render as binary until #11 merges (which clears it).

PR #15, opened 2026-08-08
feat(camera): stored-only push snapshots

lenoxys:
Replaces the ambiguous `camera.snapshot()` with an explicit split: `snapshotStored()` is a passive read of the latest retained push thumbnail, `snapshotLive()` stays the explicit live capture. The old stored-then-live fallback, its P2P stored lookup and its TTL cache are removed rather than kept beside the new surface.

## What it does

- `snapshotStored()` returns retained JPEG bytes and does no work at call time: no P2P, no storage query, no live source, no FFmpeg, no network.
- Thumbnails are acquired eagerly when a push arrives, because signed media URLs expire before a caller asks for an image.
- A candidate is accepted only when the push attributes it to exactly one account-known device that has snapshot capability evidence. Station-only and conflicting attribution are discarded.
- One valid JPEG is retained per device, in memory only. A later failed candidate leaves the older image in place; an older slow completion never replaces a newer success.
- Absence is typed: `not-observed`, `pending`, `download-failed`, `invalid-image`. Session expiry stays a lifecycle error instead of being flattened into absence.
- `storedSnapshotCache: false` disables acquisition and omits the method from bound cameras. Logout, session clear and account replacement drop retained bytes.

## Transport changes

Media download is now bounded: HTTPS-only allowlisted hosts, no credentials, no non-default ports, no IP literals, and rejection of private, loopback and link-local resolution. One redirect to an allowlisted object host is permitted, account authentication headers are never forwarded to it, and each attempt is capped at 15 s and 10 MiB. Failures carry no URL, body, identifier or token. Diagnostics are sanitised and rate-limited.

`downloadMedia()` keeps returning raw bytes. `downloadImage()` adds vendor-wrapper normalisation on top, so every push media field benefits from the same bounded path.

## Verified against real hardware

The active fleet emits **v1** `eufysecurity:` thumbnails, not v2. That contradicts the previous note in the v1 decoder, which is now corrected: the decoder is production-verified and its output is a valid, viewable JPEG. A live doorbell event was carried end to end — push, attribution, bounded download, decryption, validation — and `snapshotStored()` returned the image.

Two fleet facts worth knowing: text-only notification style sends an empty `pic_url`, so nothing is acquired; and the thumbnail arrives on a later push than the text event.

## Deferred

V2 wrappers are rejected as `invalid-image`. Reconstructing them needs full image decode and re-encode, which is a caller concern and outside this contract.

## Verification

`npm run verify` is green: format, typecheck, all four guards, build, `.d.ts` seal, ESM smoke, examples, and the full suite.


martijnpoppen (approved):
lgtm

PR #16, opened 2026-08-09
feat: expose explicit device availability

lenoxys:
## Summary

- Expose the latest explicit device availability observation from the verified `eufy_life` state signal.
- Harmonize authoritative observations per device with source-aware ordering and duplicate coalescing.
- Keep silence, stale cloud facts, idle transports, failed operations, push `m`, and vacuum DP inactivity as unobserved rather than inferred unavailable.

## Validation

- `npm run verify`: 2,061 passed, 1 skipped.
- Read-only checks against both fleets confirmed separate light and vacuum MQTT scopes and no availability observation without an explicit signal.


martijnpoppen (approved):
lgtm

PR #17, opened 2026-08-10
fix(test): include alias evidence in manifest fixture

lenoxys:
## Summary

- include read-alias parameter IDs in the manifest fixture's synthetic reported-everything evidence set
- restore the release verification gate after camera status LED selected its doorbell-specific alias

## Root cause

Release run 31368356991 failed before publishing because the fixture enabled every capability, including `doorbell`, but populated `paramIds` with primary property IDs only. The camera status LED therefore selected alias parameter `1716` while the fixture incorrectly treated that parameter as unreported.

## Verification

- `npm test -- src/model/capabilities/__tests__/manifest.spec.ts`
- `npm run verify`
- 114 test files passed; 2,143 tests passed; 1 skipped


lenoxys:
merging just to fix the CI/CD publish, should not happend when we will go public!

PR #26, opened 2026-08-14
feat: expose smart-light custom color

### LINKED ISSUE #25: Expose evidenced smart-light custom color
## Problem Statement

Callers can control evidenced smart-light power, configured brightness, and gallery effects, but they cannot select a plain custom color through the typed SDK. The current app uses a separate custom-color wire path from the gallery-effect path, and treating the two as interchangeable would send the wrong frame. Fully custom DIY animations use additional, incompletely verified protocols and must not be bundled into the same feature.

The existing evidence establishes a T8L02 custom-color write, but not an authoritative current RGB observation. The SDK therefore needs an explicit, evidence-bounded command surface that is honest about write acknowledgement and does not invent readable color state.

## Solution

Add an explicit custom-color method to the `smart_light` capability for the exact device family and `0x0206` frame shape confirmed by current-app evidence and live hardware. The capability will accept a typed RGB value and emit transport-neutral intent. The secure-MQTT transport will serialize that intent using the existing RGBCW conversion and the confirmed custom-color field layout.

The method will change color only. It will not silently change configured brightness, claim that delivery proves physical convergence, or expose a current-color getter without new inbound evidence. Unsupported models, unavailable segment evidence, malformed colors, and any unverified frame requirement will fail closed before dispatch.

## User Stories

1. As an SDK caller, I want to select a custom RGB color on an evidenced smart light, so that I am not limited to cloud gallery effects.
2. As an SDK caller, I want custom color and gallery-effect selection to remain distinct methods, so that their incompatible wire protocols are not conflated.
3. As an SDK caller, I want color input to be strongly typed and range checked, so that malformed channel values never reach a fire-and-forget transport.
4. As an SDK caller, I want custom-color support gated to confirmed device families, so that another T8L0x model never receives a T8L02-specific frame.
5. As an SDK caller, I want an unsupported custom-color request to reject before dispatch, so that refusal cannot look like device success.
6. As an SDK caller, I want a color change to preserve configured brightness, so that choosing a color does not unexpectedly alter output level.
7. As an SDK caller, I want brightness to remain controllable through the existing brightness member, so that the two durable settings stay explicit.
8. As an SDK caller, I want color applied only when the SDK has authoritative segment-count evidence, so that LED positions are not guessed.
9. As an SDK caller, I want command completion to mean transport acknowledgement only, so that I do not mistake it for observed RGB state.
10. As an SDK caller, I want the SDK to omit a current-color getter until a real inbound report identifies RGB semantics, so that absent evidence is not presented as state.
11. As a maintainer, I want model semantics and wire serialization separated, so that the capability remains independent from MQTT implementation details.
12. As a maintainer, I want byte-level custom-color fixtures derived from reviewed evidence, so that tag order, widths, sentinels, and RGBCW bytes cannot drift.
13. As a maintainer, I want real identifiers excluded from tests and documentation, so that color verification introduces no account or device data.
14. As a maintainer, I want fully custom DIY animations kept separate, so that the unverified chunked and multi-array protocols cannot enter through a simpler color feature.
15. As a maintainer, I want future authoritative color reports to deepen the same capability rather than creating a second color API, so that read and write semantics remain local.

## Implementation Decisions

- Add custom color to the existing `smart_light` capability rather than creating a new capability. Plain color is one mode of the same physical light.
- Represent input as a typed RGB value with integer red, green, and blue channels in the inclusive range 0-255. Do not accept loosely parsed CSS strings or silently clamp invalid values.
- Expose custom color as an explicit method, not as a readable property. The current realtime report identifies power, configured brightness, selected/running effect identifiers, and gradient mode, but does not report authoritative RGB.
- Keep custom color distinct from `setEffect(lightId)`. Gallery effects use the cloud-catalog `0x020D` path; plain custom color uses the independently confirmed `0x0206` path and a non-catalog sentinel.
- Gate dispatch to the exact normalized model set with live-confirmed custom-color encoding. Initially this is T8L02 only. Other family members reject before reaching the command sink.
- Require reported light-length/segment evidence before dispatch and derive the foreground position list from that evidence. Do not hard-code the observed ten-position example as a family-wide fact.
- Reuse the transport-owned RGB-to-RGBCW conversion already exercised by gallery-effect serialization. The capability carries semantic RGB intent and owns the feature-command identifiers; it does not import the transport serializer.
- Introduce or extend a transport-neutral command intent named for the DP wire action, not for the `smart_light` capability. The secure-MQTT router owns custom-color frame construction.
- Serialize the confirmed `0x0206` custom-color fields exactly, including field widths, foreground color block shape, empty background color, non-catalog cloud-effect sentinel, setting mode, and all confirmed fixed values. Any field whose generation rule is not established by current-app evidence must be resolved from evidence before implementation rather than guessed.
- A plain color write changes color only. It does not issue the existing `0x0201` brightness command. Callers use `setBrightness` separately when they intend to change configured brightness.
- Promise resolution remains command acknowledgement. Do not mutate cached params, synthesize an event, or install a current-color getter after dispatch.
- Keep the method absent from generic described property controls because no authoritative readback exists. Its public documentation must explicitly state the verified model boundary and lack of current-color observation.
- Preserve the existing capability/transport decorrelation. Model tests use a hand-written command sink; transport tests never import the capability.
- If reviewed evidence shows that the local custom-color identifier, segment positions, or another required field cannot be generated safely, the method must remain unavailable and the issue should record the exact unresolved evidence gap rather than ship a partial frame.

## Testing Decisions

- Test behavior through two existing seams because the enforced architecture deliberately separates semantic intent from byte serialization.
- At the capability seam, bind `smart_light` with a hand-written `CommandSink` and assert that a valid typed RGB value on an evidenced model emits exactly one transport-neutral intent.
- At the capability seam, prove rejection without dispatch for unsupported models, missing segment evidence, non-integer channels, and channels outside 0-255.
- At the capability seam, prove custom color does not emit a brightness command, mutate readable state, or create a current-color getter.
- At the transport seam, route a synthetic custom-color intent and compare the complete DP field sequence against reviewed byte literals independent from the implementation.
- Cover pure primaries, white/mixed output, empty background color, all-segment positions, field widths, sentinel values, and malformed intent rejection.
- Reuse the existing RGBCW vector tests as prior art, but keep at least one full `0x0206` frame contract so a correct color block inside an incorrect envelope still fails.
- Verify that one request produces one dispatch and no automatic retry. Hermetic tests must not authenticate, contact a live account, or contain real serials, account IDs, keys, captures, or names.
- Run typechecking alongside focused capability and transport specifications, then the complete SDK verification gate.

## Out of Scope

- Exposing hue, saturation, color temperature, or any authoritative current-color getter without a verified inbound report.
- Claiming that command acknowledgement proves the physical color changed.
- Mapping custom color into any particular host, smart-home platform, or presentation model.
- Automatically coupling color and brightness writes.
- Generalizing T8L02 evidence to other T8L0x models.
- Gallery-effect discovery or `setEffect(lightId)`, which already has a separate contract.
- AI scene generation, whose recommendation endpoint currently returns keywords without an addressable effect identifier.
- Fully custom DIY/animated effects over `0x0202`/`0x0203` or `0x0209`; their chunking, arrays, and lifecycle require separate evidence and a separate specification.
- Persisting or replaying raw captures in the repository.

## Further Notes

- Current reverse-engineering evidence distinguishes plain custom color (`0x0206`) from gallery effects (`0x020D`) and from fully custom animations (`0x0202`/`0x0203`, `0x0209`).
- The confirmed custom-color frame carries one foreground RGBCW block, an all-segment position list, no background color, and a non-catalog cloud-effect sentinel. Foreground brightness in that frame does not change the device's configured brightness.
- Existing realtime state is insufficient for truthful RGB reconciliation. A later current-app capture that exposes color should extend the same member table with an evidence-gated read rather than infer state from the last command.
- No existing open SDK issue was found for custom color or DIY smart-light effects.

### LINKED ISSUE #989: Control smart lights with projection and reconciliation
## Parent

#981

## What to build

Expose evidenced smart-light power and brightness with authoritative observation, target-only projection, and bounded persistent-write reconciliation.

## Acceptance criteria

- [ ] Contracts cover matching and conflicting observations, serialized and coalesced writes, generations, operation failure, the eight-second deadline, and 60-second reconciliation expiry.
- [ ] Each request receives one attempt with no automatic retry or generic post-write poll, and acknowledgement is never presented as physical truth.

## Blocked by

- #983
- #986

lenoxys:
## Summary

- add typed RGB custom-color control for evidenced smart lights
- keep plain color distinct from gallery effects and DIY animations
- require exact model and reported segment-count evidence before dispatch
- preserve configured brightness and expose no unsupported color readback
- report completion as MQTT publication, not physical convergence

## Evidence

- current-app evidence confirms the dedicated `0x0206` custom-color frame and non-catalog sentinel
- live T8L02 reports 10 addressable segments
- pure red, green, blue, and red+green/yellow physically validated on live hardware
- configured brightness remained unchanged during color writes
- live red and blue produce byte-identical `0x0204` reports and `0x0A00` replies; no state field carries authoritative RGB
- direct `0x0A06` responses are byte-identical for red and blue (`status=0`, one zero-valued tag); they acknowledge the command family without echoing RGB
- the cloud record exposes no color parameter or product data-point schema
- existing power-off control reconfirmed after color validation

## Verification

- `npm run verify` (116 files, 2217 passed, 1 skipped)
- documentation guard and build pass
- standards and specification review completed

Closes #25


martijnpoppen (approved):
💡

PR #29, opened 2026-08-16
feat: decode v2 (v2_eufysecurity) push thumbnails (keyless, native TS)

max246:
## Problem

Newer camera firmware sends push thumbnails in the **v2** (`v2_eufysecurity:`) format. The SDK only decoded **v1**, so `normalizePushImage` passed v2 through undecoded and it failed JPEG validation (`StoredImageCache` → `invalid-image`) — a host would show no picture for the latest event.

## Fix

Native TypeScript v2 decoder. v2 is **head-only obfuscation, not E2E**: only a fixed JPEG prefix (quant tables + SOF dimensions + luma Huffman) is scrambled; everything from the DC-chroma DHT (`FF C4 00 1F 01`) on is plaintext standard JPEG. `decodeImageV2` reconstructs a viewable image keylessly:

- splices a freshly built standard JPEG header onto the plaintext tail;
- recovers the three lost unknowns by search — **subsampling** (lowest colour spread), **width** (lowest row shear), **height** (tallest frame the entropy scan fills — a baseline decoder *throws* on an under-filled frame, and that throw is the fill signal);
- **de-fogs** with per-channel auto-contrast (a byte-exact port of PIL `ImageOps.autocontrast`, cutoff 0.5% — whole-bin cutoff trim + integer-truncated LUT) to fix the low contrast left by the lost quant tables, then re-encodes to a clean baseline JPEG.

`normalizePushImage` now routes v2 → `decodeImageV2` before the v1 path.

## Dependency

Adds **`jpeg-js`** (v0.4.x, BSD-3-Clause, pure-JS, **zero transitive dependencies**).

A baseline JPEG decoder+encoder is unavoidable here (the geometry search renders candidate frames; de-fog is a pixel op). `jpeg-js` specifically is required because its decoder **throws on an under-filled frame** — that throw is the exact signal the height search uses. The maintained libjpeg/mozjpeg-based alternatives (`@jsquash/jpeg` WASM, `sharp` native) grey-fill the remainder instead, which would silently break the search (verified with a probe), and both add a much larger footprint (WASM asset / native prebuilds + transitive deps) and an async API. Documented in the runtime dependency list (`docs/getting-started.md`).

**CPU behavior:** `jpeg-js` is synchronous. Reconstructing one v2 thumbnail runs repeated candidate decodes and blocks the Node.js event loop until that image finishes. The synthetic 176×144 and 264×200 fixtures each took about one second on one Node 24 test host; timing varies by image and hardware, and simultaneous thumbnails queue behind this work.

**Audit footprint:** `npm audit --omit=dev` reports **0 production vulnerabilities** and `jpeg-js` has zero transitive dependencies.

## Validation

- Fixtures are **fully synthetic** — no captured device data or serials. Generated by `scripts/dev/gen_v2_fixture.py` (Pillow is a dev-time tool, not a dependency).
- `decode_v2.spec.ts` covers: geometry recovery on-ladder (176×144) and **off-ladder** (264×200), prefix detection including the trailing colon (rejects `v2_eufysecurityX…`), `autoContrast` parity against a Pillow reference (exact per-channel stats + specific pixels), and the **end-to-end** decode → de-fog → re-encode path (the emitted JPEG decodes to a full-range image with the expected deterministic means).
- `npm run verify` green (2227 tests); commits are signed.


src/transport/http/__tests__/decode_v2.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+import { decodeImageV2, isV2Image } from "../decodeImageV2.js";
+import { normalizePushImage } from "../decodeImageV1.js";
+
+// A synthetic v2 blob: the ascii `v2_eufysecurity:` wrapper followed by a standard 176×144 4:4:4
+// JPEG (its plaintext tail from `FF C4 00 1F 01` is exactly what a real v2 blob leaves in the clear).
+const FIXTURE_DIR = join(dirname(fileURLToPath(import.meta.url)), "fixtures");
+const readFixture = (name: string) => Buffer.from(readFileSync(join(FIXTURE_DIR, name), "utf-8"), "base64");
+const v2Blob = readFixture("v2_thumbnail_176x144.b64");
+// A real captured v2_eufysecurity blob (264×200, 4:4:4), from the mega_preview investigation.
+const realV2Blob = readFixture("v2_thumbnail_264x200.b64");
```
lenoxys: **Blocking:** This is real captured device data, and the fixture contains an unredacted serial. Repo policy requires synthetic fixtures only. Please remove it and replace it with a fully synthetic equivalent.

package.json
```diff
@@ -70,6 +70,7 @@
     "node": ">=24.5.0"
   },
   "dependencies": {
+    "jpeg-js": "^0.4.4",
```
lenoxys, line 73: **Blocking:** Adding a runtime dependency needs justification against the repo’s dependency policy: why no built-in or small inline implementation works, plus its transitive and audit footprint. Please add that rationale to the PR.

src/transport/http/decodeImageV2.ts
```diff
@@ -0,0 +1,262 @@
...
+const DC_CHROMA = Buffer.from([0xff, 0xc4, 0x00, 0x1f, 0x01]);
+
+/** Zig-zag scan order — DQT segments store the 8×8 quant table in this order. */
+const ZIGZAG = [
+  0, 1, 8, 16, 9, 2, 3, 10, 17, 24, 32, 25, 18, 11, 4, 5, 12, 19, 26, 33, 40, 48, 41, 34, 27, 20, 13, 6, 7, 14, 21, 28,
+  35, 42, 49, 56, 57, 50, 43, 36, 29, 22, 15, 23, 30, 37, 44, 51, 58, 59, 52, 45, 38, 31, 39, 46, 53, 60, 61, 54, 47,
+  55, 62, 63,
+];
+
+// Standard Annex-K base quantization tables (natural order, quality 50). The originals are lost with
```
lenoxys: Please remove inline body comments like this. Keep necessary rationale as declaration-level JSDoc; delete comments that only narrate the code.

src/transport/http/decodeImageV2.ts
```diff
@@ -0,0 +1,262 @@
...
+ */
+export function decodeImageV2(data: Buffer): Buffer | null {
+  if (!isV2Image(data)) return null;
+  const cut = data.indexOf(DC_CHROMA);
+  if (cut < 0) return null;
+  const tail = data.subarray(cut);
+
+  // 1) subsampling + coarse geometry: the largest ladder frame that decodes (the scan fills it),
+  //    tie-broken across subsamplings by the lowest colour spread.
+  let best: { spread: number; ss: number; w: number; h: number } | null = null;
```
lenoxys: The abbreviated geometry names make this subtle search algorithm difficult to audit. Please use explicit names for subsampling, dimensions, and MCU sizes.

src/transport/http/decodeImageV2.ts
```diff
@@ -0,0 +1,262 @@
...
+      if (acc > clip) break;
+    }
+    for (let acc = 0; hi > 0; hi--) {
+      acc += hist[hi];
+      if (acc > clip) break;
+    }
+    if (hi <= lo) continue;
+    const scale = 255 / (hi - lo);
+    const lut = new Uint8Array(256);
+    for (let v = 0; v < 256; v++) lut[v] = Math.min(255, Math.max(0, Math.round((v - lo) * scale)));
```
lenoxys: Pillow truncates the scaled value; `Math.round` changes pixel values and breaks the claimed `ImageOps.autocontrast` parity. Please use integer truncation.

src/transport/http/__tests__/decode_v2.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+    // Geometry recovered by the search (width refined, height pinned to the scan's fill).
+    expect(jpegSize(jpeg!)).toEqual({ width: 176, height: 144 });
+  });
+
+  it("decodes a real captured v2 thumbnail to its original geometry (264×200)", () => {
+    const jpeg = decodeImageV2(realV2Blob);
+    expect(jpeg).not.toBeNull();
+    expect(jpeg!.subarray(0, 2)).toEqual(Buffer.from([0xff, 0xd8]));
+    expect(jpeg!.subarray(-2)).toEqual(Buffer.from([0xff, 0xd9])); // EOI — a complete JPEG
+    expect(jpegSize(jpeg!)).toEqual({ width: 264, height: 200 });
```
lenoxys, line 48: The PR says contrast parity is tested, but these assertions only cover geometry and JPEG framing. Please add synthetic-fixture assertions for the expected post-contrast statistics.

src/transport/http/decodeImageV2.ts
```diff
@@ -0,0 +1,262 @@
...
+    while (h > mch && !decodeCandidate(tail, w, h, ss)) h -= mch;
+  } else {
+    while (h < 4096 && decodeCandidate(tail, w, h + mch, ss)) h += mch;
+  }
+  return h;
+}
+
+/** True if the blob is a v2 `v2_eufysecurity:` push thumbnail. */
+export function isV2Image(data: Buffer): boolean {
+  return data.length >= 41 && data.subarray(0, 15).toString("latin1") === "v2_eufysecurity";
```
lenoxys: Please match the complete `v2_eufysecurity:` prefix. This check also accepts unrelated values such as `v2_eufysecurityX` and routes them into the decoder.

lenoxys (changes requested):
## Standards

The fixture/privacy and runtime-dependency policy issues are blocking. The remaining comments address repository documentation and naming standards.

## Spec

The decoder is directionally aligned, but auto-contrast parity, format detection, and the claimed contrast test coverage need correction.

max246:
Thanks @lenoxys — all seven addressed in the latest push:

1. **Fixture privacy (blocking).** The 264×200 fixture was real captured data with a live HomeBase serial. Replaced with a **fully synthetic** blob (synthetic serial + a Pillow-generated 4:4:4 baseline JPEG). Added `scripts/dev/gen_v2_fixture.py` so fixtures are reproducible; Pillow is a dev-time tool, **not** a runtime dependency. Both fixtures are now synthetic and the test comments say so.

2. **`jpeg-js` dependency (blocking).** Justification added to the module `@remarks`: it's BSD-3-Clause, pure-JS, **zero transitive deps**. A baseline JPEG codec is unavoidable (decode candidate frames to detect scan-fill + read pixels for the search metrics, then re-encode the de-fogged image); an inline decoder+encoder would be 1000+ lines of bit-level code with far more audit/correctness risk than a small vetted lib.

3. **Inline body comments.** Promoted the declaration-level narration (quant tables, DHT tables, LADDER) to JSDoc; dropped pure what-comments.

4. **Abbreviated geometry names.** The search now uses `subsampling`, `width`, `height`, `mcuWidth`, `mcuHeight`, `coarseWidth/Height`, `candidateWidth/Height` throughout (was `ss`/`w0`/`h0`/`mcw`/`mch`).

5. **`Math.round` breaks autocontrast parity.** Reworked `autoContrast` into a **byte-exact** PIL `ImageOps.autocontrast` port: cutoff = `n*cutoff//100`, trimmed by zeroing whole bins off each end, range = first/last non-empty bin, and the LUT now **int-truncates** (`Math.trunc`) instead of rounding.

6. **Contrast test coverage.** Exported `autoContrast` and added two parity tests asserting exact per-channel sums/min/max and specific pixel values against a Pillow 10 reference (the cutoff-trim path is exercised via a 400-px buffer with outliers).

7. **Prefix over-match.** `isV2Image` now matches the complete `v2_eufysecurity:` prefix **including the colon**, so `v2_eufysecurityX…` is correctly rejected — added a test for it.

Full suite green (2225 passed), typecheck + all guards + prettier clean.

package.json
lenoxys, line 73: **Blocking:** `jpeg-js` has not been released since October 2022, and this path performs repeated synchronous decodes on the SDK event loop. Please use a maintained codec or provide a stronger justification for accepting an unmaintained runtime dependency. `sharp` is a possible candidate, but its under-filled-frame behavior, native package footprint, and supported-platform installation need to be verified first.

src/transport/http/decodeImageV2.ts
```diff
@@ -0,0 +1,307 @@
...
+  width: number;
+  height: number;
+  data: Uint8Array;
+}
+
+/** Splice a header for (width, height, subsampling) onto the tail and decode; null if the scan doesn't fill the frame. */
+function decodeCandidate(tail: Buffer, width: number, height: number, subsampling: number): Decoded | null {
+  try {
+    const jpeg = Buffer.concat([buildHeader(width, height, subsampling), tail]);
+    return jpegDecode(jpeg, { useTArray: true, maxMemoryUsageInMB: 128 });
```
lenoxys, line 149: This synchronous candidate search takes roughly one second for the small synthetic fixtures on the current environment. Because it runs during push-image acquisition, it can block unrelated SDK event processing. Please move the work off the main thread or redesign the search to avoid repeated full decodes.

src/transport/http/decodeImageV2.ts
```diff
@@ -0,0 +1,307 @@
...
+  } catch {
```
lenoxys, line 150: The geometry algorithm depends on the codec rejecting an under-filled frame. Please cover that behavior explicitly at the codec boundary so replacing or upgrading the codec cannot silently change the recovered dimensions.

src/transport/http/__tests__/decode_v2.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+    // Geometry recovered by the search (width refined, height pinned to the scan's fill).
+    expect(jpegSize(jpeg!)).toEqual({ width: 176, height: 144 });
+  });
+
+  it("recovers an off-ladder geometry from a synthetic blob (264×200)", () => {
+    const jpeg = decodeImageV2(v2Blob264);
+    expect(jpeg).not.toBeNull();
+    expect(jpeg!.subarray(0, 2)).toEqual(Buffer.from([0xff, 0xd8]));
+    expect(jpeg!.subarray(-2)).toEqual(Buffer.from([0xff, 0xd9])); // EOI — a complete JPEG
+    expect(jpegSize(jpeg!)).toEqual({ width: 264, height: 200 });
```
lenoxys, line 48: This verifies framing and geometry, but not the claimed end-to-end contrast parity. Please assert contrast statistics or equivalent Pillow-reference output from the complete decoder result, after re-encoding.

src/transport/http/__tests__/decode_v2.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+    data[1] = 2;
+    data[5] = 254;
+    data[2] = 1;
+    data[6] = 255;
+    return data;
+  }
+
+  it("matches PIL's per-channel output statistics exactly", () => {
+    const data = buildInput();
+    autoContrast(data, W, H);
```
lenoxys, line 121: Testing the helper is useful, but it does not prove that the decoder invokes it correctly. Keep this unit test and add the end-to-end decoder assertion separately.

src/transport/http/decodeImageV2.ts
```diff
@@ -0,0 +1,307 @@
...
+function dqtSegment(table: readonly number[], id: number, quality: number): Buffer {
+  const q = scaleQuant(table, quality);
+  const body = Buffer.alloc(65);
+  body[0] = id;
+  for (let i = 0; i < 64; i++) body[1 + i] = q[ZIGZAG[i]];
+  return Buffer.concat([Buffer.from([0xff, 0xdb, 0x00, 0x43]), body]);
+}
+
+function sofSegment(width: number, height: number, subsampling: number): Buffer {
+  // Y sampling factors; Cb/Cr are 1×1.
```
lenoxys: Please convert the remaining implementation comments to declaration-level JSDoc or remove narration that the code already expresses, per the repository comment rule.

scripts/dev/gen_v2_fixture.py
```diff
@@ -0,0 +1,53 @@
...
+import base64
+import sys
+
+from PIL import Image
+
+SYNTHETIC_SERIAL = b"v2_eufysecurity:T8000TEST00000002:0000000000:"
+
+
+def synthetic_image(w: int, h: int) -> Image.Image:
+    # Photo-like content with strong horizontal continuity so the width search (min row-shear) locks
```
lenoxys: The repository documentation rule applies here as well. Please use declaration-level documentation or a module/function docstring rather than inline implementation narration.

lenoxys (changes requested):
## Standards

Most first-round implementation issues are resolved, but this still needs changes:

- Rewrite the branch to remove the real captured fixture and serial from PR history.
- Sign the replacement commits.
- Update the PR description to remove the consumer-specific reference and stale real-fixture claims.
- Document the selected codec in the runtime dependency list and provide the required dependency, transitive, and audit justification in the PR.

## Spec

Prefix detection and Pillow-compatible truncation are fixed. End-to-end contrast parity is still not covered: the tests exercise the helper independently rather than verifying the complete decode and re-encode path.

`npm run verify` passes. Live verification was inconclusive: the account exposed 14 devices, but no thumbnail event arrived during the observation window.

max246:
Second round addressed — force-pushed a rewritten, signed branch:

1. **Real fixture/serial purged from history.** The branch was rewritten to a single commit off the base; the captured 264×200 blob (and its HomeBase serial) never appears in any commit now. Both fixtures are synthetic, generated by `scripts/dev/gen_v2_fixture.py`. Verified: `git log -p base..HEAD` has **0** hits for the old serial.

2. **Commit signed.** The replacement commit is SSH-signed (`git log --show-signature` → *Good "git" signature*).

3. **PR description updated.** Dropped the consumer-specific (`mega_preview`/Python) reference and the stale "real captured fixture" claims; added the dependency section.

4. **Codec dependency — resolved by removing it.** You flagged jpeg-js as unmaintained (correct — last release 2022). Rather than swap to a libjpeg/mozjpeg-based lib (they *grey-fill* under-filled frames instead of throwing, which would silently break the geometry search) or add a WASM/native dep, I **vendored** jpeg-js's baseline `decode`/`encode` in-tree under `src/transport/http/jpeg` (verbatim, `@ts-nocheck`, prettier-ignored, original Apache-2.0/BSD-3 licences retained; a typed `index.ts` is the only import surface). **Runtime deps stay three** — the dependency, transitive, and audit footprint is now zero external packages, and we own patching. Documented in `docs/getting-started.md`.

5. **End-to-end contrast parity.** Added a full-pipeline test: it runs `decodeImageV2`, decodes the *emitted* JPEG, and asserts the de-fog survived the decode→re-encode round-trip (full 0..255 range + deterministic per-channel means) — not just the helper in isolation. The exact-parity `autoContrast` unit tests remain too.

`npm run verify` green (2227 tests, all guards, ESM + examples typecheck).

max246:
Thanks — since you're happy to take the dependency, I've dropped the in-tree vendored copy and gone back to importing **`jpeg-js`** directly (simpler, less code to own). Force-pushed:

- removed `src/transport/http/jpeg/*` (decoder/encoder/index/README) and the `.prettierignore` entries;
- `decodeImageV2` + the spec import from `jpeg-js` again; `jpeg-js` restored to `dependencies`;
- runtime dependency list back to four in `docs/getting-started.md` (module `@remarks` notes why jpeg-js specifically — the throw-on-under-filled-frame behaviour the geometry search needs).

Everything else from the prior round stands: synthetic fixtures, PIL-exact autocontrast, the end-to-end contrast test, and a signed commit. `npm run verify` green (2227 tests).

lenoxys:
Follow-up edit pushed in signed commit `098286a`:

- kept `jpeg-js` per the maintainer discussion and documented why its under-filled-frame failure behavior is required;
- added it to the authoritative runtime dependency list and recorded `npm audit --omit=dev`: 0 production vulnerabilities;
- documented the synchronous CPU trade-off in source, user docs, and the PR body: repeated candidate decoding blocks the event loop for about one second on the synthetic Node 24 fixture runs, with hardware/image-dependent timing;
- removed remaining inline narration and retained rationale as declaration-level documentation;
- preserved the synthetic parity inputs and the end-to-end decode, de-fog, and re-encode assertions.

The branch history contains only synthetic fixtures. The superseded GitHub object remains a separate Support-side purge because GitHub does not expose cached-object deletion through Git or the API.

Local `npm run verify` is green with 2,227 passing tests, both CI builds pass, and all review threads are resolved.

lenoxys (approved):
Approved after the signed follow-up edit, full verification, and resolution of all inline review threads.

PR #30, opened 2026-08-16
fix: reconcile valueless arming transitions

### LINKED ISSUE #991: Represent security-system arming and alarms
## Parent

#981

## What to build

Expose Stay, Away, and Disarm with explicit SDK translation, truthful unsupported-mode faults, and authoritative alarm behavior.

## Acceptance criteria

- [ ] Unsupported observed modes fault current and target reads while admitted controls remain available to leave that mode.
- [ ] Delayed and active alarm phases report triggered immediately and clear only after a later authoritative arming observation.

## Blocked by

- #989

lenoxys:
## What changed
- refresh capability-declared reflected state before emitting valueless semantic transitions
- preserve transport acknowledgement semantics while authoritative convergence continues in the keyed transaction
- serialize the complete observed-command transaction before transport dispatch
- suppress duplicate push handling while command convergence is active
- reset standalone P2P sessions after convergence without interrupting active stream consumers
- reuse one SessionManager discard path for idle, immediate, removal, and full shutdown lifecycle

## Why
Arming writes could apply physically while the retained device still exposed the previous mode. Reused standalone P2P sessions could also acknowledge a subsequent mode command without applying it.

## Verification
- `npm run verify`
- 117 test files passed
- 2228 tests passed, 1 skipped
- focused concurrent-command, readback-timeout, repeated-poll, command-metadata, and session-reset contracts
- de-identified live standalone-camera check: Home -> Disarmed -> Home converged as `1 -> 63 -> 1` in one SDK client, including immediate reversal
- AGENTS.md comment/JSDoc and reuse-before-abstraction audits completed

## Review decisions
- implemented active-consumer-safe deferred session reset
- retained transport-only write acknowledgement; readback timeout is reported asynchronously
- retained fail-closed waiting for unchanged valueless pushes because duplicate and stale-cloud states are indistinguishable before convergence
- apply retained device params only once after reflected raw state converges
- allow observation metadata replacement on reused command objects

## Scope
No credentials, device identifiers, names, raw payloads, or captures are included.


max246 (commented):
Review of the arming reconciliation. The core mechanism (capability-declared `observation` → symbol-tagged command → serialized transaction: route → bounded readback → standalone-session reset → emit) is clean and the SessionManager `discard()` unification is a nice simplification; tests cover the concurrency/lifecycle contracts well. A few things to weigh, mostly around the new blocking/throwing semantics of writes and edge cases of the valueless-push refresh — inline. None are blockers, but #1 (stream teardown) and #2 (setProperty now blocks+throws) are worth an explicit decision + doc.

src/transport/p2p/command-router.ts
```diff
@@ -238,6 +238,14 @@ export class P2PCommandRouter {
     return dev ? this.stationKeyFor(dev) : sn;
   }
 
+  /** Reset only a standalone device's session; an attached device must not close its shared HomeBase. */
+  async resetStandaloneSession(sn: string): Promise<void> {
+    const device = this.deps.listDevices().find((candidate) => candidate.sn === sn);
+    if (!device) return;
+    const station = this.stationKeyFor(device);
+    if (station === sn) await this.manager.close(station);
```
max246: `manager.close(station)` force-discards the entry via `discard()`, which does **not** check `users` (unlike the idle `expire` path). So resetting a standalone camera's session after an arming write will tear down a session that has active users — e.g. a live stream running on that same standalone cam gets killed by an arming change. Is that acceptable, or should the reset defer when `users > 0` (or re-open transparently)? Worth a comment either way, since `close()`'s unconditional teardown is the sharp edge here.

src/client/eufy-mega.ts
```diff
@@ -665,7 +783,33 @@ export class EufyMega extends EventEmitter {
...
+          return;
+        }
+        const key = this.eventRefreshKey(sn, observation);
+        const epoch = this.realtimeEpoch;
+        const owner = this.commandRefreshes.get(key);
+        const currentOwner = owner?.epoch === epoch ? owner : { epoch, count: 0 };
+        currentOwner.count += 1;
+        this.commandRefreshes.set(key, currentOwner);
+        try {
+          await this.enqueueStateTransition(key, async () => {
```
max246: This makes `dispatch` (and therefore `setProperty(mode)` / the fluent `arming.set*`) **block until cloud convergence** and **throw** if `refreshEventState` times out — even though the write physically applied. A caller (e.g. an HA service call) now blocks up to `timeoutMs` (20 s) and surfaces an error on slow cloud reflection for a mode that did change. Is that the intended contract for the write? If so, a note on the setter/`dispatch` documenting the new "awaits authoritative readback, throws on non-convergence" semantics would save the next reader; if not, consider confirming out-of-band and not failing the write on readback timeout.

src/client/eufy-mega.ts
```diff
@@ -495,14 +510,114 @@ export class EufyMega extends EventEmitter {
...
+    this.stateTransitions.set(key, tracked);
+    return tracked;
+  }
+
+  private matchesObservation(
+    value: unknown,
+    refresh: SemanticEventRefresh & { expected?: boolean | number | string },
+    before: unknown,
+  ): boolean {
+    return refresh.expected === undefined ? value !== before : String(value) === String(refresh.expected);
```
max246, line 604: On the **push** refresh path there's no `expected`, so convergence is `value !== before`. If a valueless MODE_SWITCH push arrives when the value already equals `before` — a duplicate/replayed push, or the mode already reflected via another channel — this never becomes true, so `refreshEventState` loops the full `timeoutMs` and then throws → `reportError` emits an `error`. Consider treating an unchanged repeat as already-converged (no-op) rather than a 20 s wait + error, or gating the push refresh on the retained value still being stale.

src/client/eufy-mega.ts
```diff
@@ -495,14 +510,114 @@ export class EufyMega extends EventEmitter {
...
+      const rawBefore = this.registry.require(sn).params?.[refresh.param];
+      const deadline = Date.now() + refresh.timeoutMs;
+      while (Date.now() < deadline) {
+        const remaining = deadline - Date.now();
+        await this.beforeDeadline(this.registry.getDevices(), remaining);
+        if (epoch !== this.realtimeEpoch) return false;
+        const device = this.liveDevices.get(sn)?.deref();
+        const record = this.registry.require(sn);
+        if (device) {
+          device.applyParams(record.params ?? {});
```
max246, line 569: `applyParams(record.params ?? {})` runs on **every** ~500 ms poll iteration until convergence, re-applying the entire cloud param set repeatedly. If `applyParams` has observable side effects (state map churn, `deviceState`/derived-event emission), they repeat each iteration. Consider applying only when the record actually changed, or confirm `applyParams` is cheap+side-effect-free at this cadence.

src/core/contracts.ts
```diff
@@ -107,6 +119,16 @@ export type Command =
     }
   | { kind: "aiot-dp"; dp: number; value: boolean | number | string };
 
+/** Attach non-wire observation policy to a command without changing its enumerable transport intent. @internal */
+export function observeCommand(command: Command, observation: CommandObservation): Command {
+  return Object.defineProperty(command, COMMAND_OBSERVATION, { value: observation });
```
max246: `Object.defineProperty` here leaves the symbol property non-writable/non-configurable (the defaults), so observing the *same* command object twice throws `TypeError: Cannot redefine property`. It's safe today (memberWrite builds a fresh `cmd` per call), but if a command is ever reused/re-observed this becomes a surprising throw — a one-line note, or `configurable: true`, would harden it.

src/transport/p2p/command-router.ts
lenoxys: Fixed in 7c02926. Standalone reset now ignores expiring command holds but waits while a real session consumer (for example a live stream) remains attached. The reset promise resolves only after that consumer detaches and the session actually closes, so the keyed command transaction cannot dispatch the next observed write on the old session. Added focused tests for active-consumer deferral and command-hold-only immediate reset.

src/client/eufy-mega.ts
lenoxys: Fixed in 7c02926. `dispatch` now preserves the original transport-acknowledgement contract: it resolves once routing succeeds. Authoritative convergence, standalone reset, and semantic emission continue in the keyed transaction; a post-ack readback timeout is reported asynchronously and does not reject a physically applied write. A regression test pins that timeout behavior.

src/client/eufy-mega.ts
lenoxys: Not changing unchanged state to immediate convergence. For this valueless push, a duplicate and the reproduced failure are indistinguishable at arrival: in the failure, retained and cloud values both remain old until a later list refresh. Treating the first unchanged read as converged would emit the stale mode and recreate the bug. The bounded wait therefore remains fail-closed; if no authoritative change appears, the stale event is suppressed and the SDK reports the convergence failure.

src/client/eufy-mega.ts
lenoxys: Fixed in 7c02926. Poll iterations now inspect only the reflected raw parameter. The complete cloud parameter set is applied to the retained Device exactly once, after the expected value (or externally signaled change) converges. Added a regression with multiple unchanged polls asserting one `applyParams` call.

src/core/contracts.ts
lenoxys: Fixed in 7c02926. The non-enumerable observation property is now configurable, so re-observing a reused command safely replaces the policy. Added a regression that observes the same command twice and verifies the latest metadata.

src/transport/p2p/command-router.ts
lenoxys: Follow-up 23b73a4 keeps authoritative event publication ahead of a deferred active-stream reset, while the keyed transaction still waits for actual reset completion before routing the next observed command. This prevents the consumer plugin reconciliation window from expiring during a long stream without reintroducing same-session command reuse.

src/transport/p2p/session-manager.ts
martijnpoppen, line 1: **Bug: `closeAll` leaves `resetWhenUnused` callers hanging if any `session.close()` rejects**

```typescript
await Promise.all(entries.map(({ session }) => session?.close()));
for (const entry of entries) for (const resolve of entry.resetWaiters) resolve();
```

If any `session.close()` rejects, `Promise.all` short-circuits and the `resetWaiters` resolution loop never runs. Any caller currently blocked in `resetWhenUnused` will hang indefinitely.

Change to `Promise.allSettled` so the waiter loop always executes:

```typescript
await Promise.allSettled(entries.map(({ session }) => session?.close()));
for (const entry of entries) for (const resolve of entry.resetWaiters) resolve();
```

---

src/client/eufy-mega.ts
martijnpoppen, line 1: **Nit: redundant epoch guard in `commandSinkFor` dispatch — dead code**

```typescript
if (!refreshed || epoch !== this.realtimeEpoch) return;
if (epoch === this.realtimeEpoch) this.emitSemantic(observation.event, { deviceSn: sn });
```

The second `if (epoch === this.realtimeEpoch)` is always true at that point — execution only reaches that line when the previous guard did not `return`. Drop the redundant check:

```typescript
if (!refreshed || epoch !== this.realtimeEpoch) return;
this.emitSemantic(observation.event, { deviceSn: sn });
```

---

martijnpoppen (changes requested):
The overall approach is sound — serializing valueless state-transition transactions behind a keyed queue, suppressing duplicate push handling during convergence, and the active-consumer-safe deferred session reset are all well-reasoned. The test suite covers the core contracts thoroughly.

Two findings in the PR diff:

**1. Bug (blocking): `closeAll` can leave `resetWhenUnused` callers hanging** — see inline comment on `session-manager.ts`. Fix: `Promise.allSettled`.

**2. Nit: dead-code epoch guard in `commandSinkFor`** — see inline comment on `eufy-mega.ts`.

---

**Pre-existing issues found during review (outside this PR's scope — filing separately for awareness):**

- **`src/transport/p2p/fmp4.ts` — malformed `trun` box flags**: `const flags = 0x000f01` sets the sample-composition-time-offsets-present bit but the per-sample loop only writes 12 bytes/sample instead of 16, corrupting every fMP4 fragment for strict ISO 14496-12 parsers. Should be `0x000701`.

- **`src/transport/p2p/fmp4.ts` — parameter-set NALs leaking into `mdat`**: `annexbToAvcc()` has a JSDoc claiming it drops SPS/PPS/VPS NALs but the loop iterates all NALs without filtering. A second copy of parameter sets in `mdat` violates ISO 14496-15 AVCC and breaks strict MSE/EME pipelines. Fix: skip NAL types 7/8 (H.264) and 32/33/34 (H.265).

- **`src/transport/push/push-client.ts` — TLS verification disabled**: `tls.connect(PORT, HOST, { rejectUnauthorized: false })` for the FCM/MCS connection to `mtalk.google.com:5228` allows any on-path attacker to impersonate the server and read/inject push notification metadata. Remove `rejectUnauthorized: false`.

- **`src/transport/p2p/media.ts` — silent error handler in `captureSnapshotFromShared`**: `consumer.on("error", () => {})` discards consumer errors; callers then wait the full 20 s before getting a misleading timeout rejection instead of the real error.

---

src/client/eufy-mega.ts
lenoxys: Fixed in 2da4ace; the redundant post-refresh guard is removed. The adjacent transaction now also checks the captured epoch before `routeCommand()`: live testing reproduced a queued observed command reopening one P2P session after `disconnect()`, while the fix rejects that stale command and leaves zero sessions. Added a focused disconnect regression.

src/transport/p2p/session-manager.ts
lenoxys: Fixed in 2da4ace. The diagnosis was correct, but I preserved teardown failure reporting rather than using a bare `Promise.allSettled`: each discarded entry now closes through one helper that resolves or rejects only that entry’s reset waiters, `close()` uses the same path, and `closeAll()` attempts every close before rethrowing the single failure or an `AggregateError`. Added regressions for both single-session and full-shutdown failures. A live active-stream probe with an injected post-cleanup close failure now reports `closeAll` rejected and the deferred reset waiter settled.

src/client/eufy-mega.ts
lenoxys: Closing this as an explicit fail-closed design decision rather than a code change. An unchanged first read cannot distinguish a replay from the reported stale-cloud failure; accepting it would republish stale arming state. The refresh remains bounded, suppresses the event when no authoritative transition appears, and reports the convergence failure.

max246 (approved):

PR #33, opened 2026-08-18
feat(clean): MQTT topic expansion, dynamic DP catalog, and Tuya write path for eufy_home_tuya devices

martijnpoppen:
### MQTT topic subscriptions for clean-line devices

- `subscribeTopics()` in `src/transport/mqtt/topics.ts`: vacuum and mower devices now subscribe four topic families instead of only `cmd/.../res`:
  - `cmd/eufy_home/{pn}/{sn}/res` — device→app DP reports (existing)
  - `biz/eufy_home/{pn}/{sn}/res` — cloud→app business-layer responses
  - `biz/eufy_home/{pn}/{sn}/req` — cloud ACKs for app→cloud requests
  - `dt/eufy_home/{pn}/{sn}/param_info` — device-twin parameter pushes
- `parseSecureTopic()`: now accepts `biz` and `dt` roots so `SecureMqtt`'s message handler correctly extracts the device SN from messages on the new topics.
- 9 new test assertions covering the four-topic set for vacuum/mower and the two new roots.

### Dynamic DP catalog via `get_product_data_point`

- **`src/model/capabilities/dp-catalog.ts`** (new): `DpCatalog` interface (`dpIds: ReadonlySet`, `enumRanges: ReadonlyMap`) and a defensive `parseDpCatalog(raw: unknown)` parser. Handles both field-name variants (`dp_id` / `id`), plain number arrays, JSON-stringified `"[0,1,2,3]"`, and the `{"range":["0","1","2","3"]}` object shape. Always returns `EMPTY_DP_CATALOG` on any shape mismatch — never throws.
- **`src/model/capabilities/types.ts`**: `dpCatalog?: DpCatalog` added to `CommandContext` so capabilities can consume per-SKU DP ranges without importing transport.
- **`src/client/eufy-mega.ts`**: `fetchDpCatalog(model)` private method — lazy per-T-code cache, falls back to `EMPTY_DP_CATALOG` on any error. Wired into `commandContext()` for vacuum/mower codecs only.
- **`src/model/capabilities/suction.ts`**: `actions()` now reads the DP 158 enum range from the live catalog instead of a static per-model table. Falls back to all six `SuctionLevel` values when the catalog is absent or does not cover DP 158.
- **`src/model/capabilities/__tests__/dp-catalog.spec.ts`** (new): 11 tests covering every `parseDpCatalog` branch.

### Tuya write path and initial DP poll for `eufy_home_tuya` devices

- **`startCleaning` / `returnToDock` / `pauseCleaning`** now dispatch on both platforms:
  - Legacy Tuya (G-series / X8): DP 2 bool (start=true / pause=false) and DP 101 bool (return=true).
  - AIoT: DP 152 ModeCtrlRequest protobuf — unchanged.
  - `available` guard now DP-based — the verbs appear once the device has reported DP 2/101 (Tuya) or DP 152 (AIoT).
- **`TuyaCommandRouter.fetchDps(sn)`** (new): calls `thing.m.device.cache.dp.get` and defensively extracts the DP map from either `result.dps` (Tuya-native shape) or bare `result` (flat shape). Returns `null` on failure so the caller can skip delivery safely.
- **`TuyaCommandRouter`**: now accepts a public `http` config field (replacing internal `_signer`/`_http`); `ensureLoggedIn` resets `loginOnce` on failure so the next dispatch retries rather than re-throwing the stale error.
- **`subscribeMqttDevices`**: fires a non-blocking `fetchDps` + `tuyaDpRouter.deliver` for each `eufy_home_tuya` device after MQTT subscribe, hydrating capability getters before the first realtime push.
- **`EufyMegaOptions.tuyaAllowUnverified`** (new): hosts can opt into unverified Tuya DP writes without touching internal transport config. The gate remains `false` by default.
- **`src/transport/tuya/__tests__/command-router.spec.ts`** (new): 7 tests covering login retry behavior, `fetchDps` shapes, UNVERIFIED guard, and `dispatchCommand`.

## Why

The app subscribes three additional topic families per vacuum. Without them, cloud-initiated business responses and device-twin pushes were silently dropped. The `get_product_data_point` integration replaces a static feature table with a live per-SKU DP catalog. The Tuya write path unblocks `startCleaning` / `returnToDock` / `pauseCleaning` for G-series and X8 devices that speak the legacy bool DPs, and the initial DP poll ensures capability getters are populated before the first realtime push arrives.

## Evidence

- `TopicManager.java · getBizReqTopic()` in the v6 app — three additional topic families per clean-line device.
- DP 2 (`PLAY_PAUSE`, bool) and DP 101 (`GO_HOME`, bool) confirmed from protocol inspection.
- `ICleanBridgeDeviceInterface.java · PRODUCT_TYPE_*` — 26 product-family constants (reference only; feature availability resolved dynamically via `get_product_data_point`).

## Verified how

- [x] `npm run verify` is green — 2350 tests pass
- [x] Added or updated specs — 9 assertions in `mqtt.spec.ts`, 11 in `dp-catalog.spec.ts`, 7 in `command-router.spec.ts`, updated `vacuum-clean.spec.ts`

## Unverified or deferred

- `biz/` and `dt/` subscriptions not yet confirmed live. The broker may answer SUBACK with a denial; `SecureMqtt.subscribeDevice` handles partial denials gracefully.
- `get_product_data_point` response shape not yet pinned from a live capture (`parseDpCatalog` is maximally defensive).
- Dock control DP 173 (`emptyDust` / `washMops` / `dryMops`): StationRequest protobuf not yet reversed — setters remain absent until the write wire is captured.
- `tuyaAllowUnverified` defaults to `false` — `dp.publish` round-trip not yet confirmed on a real device.


src/model/capabilities/vacuum-clean.ts
```diff
@@ -586,50 +587,56 @@ export const VACUUM_CLEAN_MEMBERS = {
...
+   * AIoT: ModeCtrlRequest method 0 over DP 152.
+   * Tuya: DP 2 = true (PLAY_PAUSE bool, confirmed from DeviceHomeModule.java).
+   */
   startCleaning: method(
     ({ sink, ctx }) => {
       if (isTuyaVacuum(ctx)) {
-        return (): Promise<void> => {
-          throw new Error("startCleaning: Tuya write path not yet verified — no live publishDps capture");
-        };
+        return (): Promise<void> => sink.dispatch(aiotDp(LEGACY_VACUUM_DP.PLAY_PAUSE, true));
```
lenoxys: **Blocking — this reverts the resolved thread on #31.** In `7985728` this branch became `throw new Error("startCleaning: Tuya write path not yet verified — no live publishDps capture")` after review; it's now a live dispatch again, and the same applies to `returnToDock` (L615) and `pauseCleaning` (L632).

Three things make this hard to land as-is:

1. **The module contradicts itself.** `X8_VACUUM_DP`'s doc at L52-53 still states *"Write direction for all DPs is unverified — no live publishDps capture has been made yet."* The PR body agrees: *"`dp.publish` round-trip not yet confirmed on a real device."*
2. **Authority is a third-party decompile.** The justification is `DeviceHomeModule.java`, which `AGENTS.md:140-144` addresses directly — older third-party clients are *"provenance for a label only, never authority for behaviour"*, and *"a model will confidently reproduce them."*
3. **The setter is now present but non-functional.** It dispatches into `TuyaCommandRouter`, which refuses by default (`allowUnverified: false`). So every host that hasn't set `tuyaAllowUnverified` gets a `startCleaning` that exists on the type and throws at runtime — precisely what `AGENTS.md:126-128` rules out (*"A described control that always fails ... is worse than none"*), and why `:132-136` requires the setter to be **absent** so the caller learns at compile time.

Also note `available: isAiotVacuum(ctx) || isTuyaVacuum(ctx)` is a tautology: `device-family.ts:184` is `category !== "eufy_home_tuya"` and `:193` is `category === "eufy_home_tuya"` — exact complements, so the guard is unconditionally `true`. That was raised on #31 as well.

If a live `publishDps` capture now exists, could it go in the PR body as evidence? That would settle all three at once. Otherwise the `throw`s should come back.

src/model/capabilities/vacuum-clean.ts
```diff
@@ -381,9 +381,10 @@ export type VacuumCleanActions = Surface<typeof VACUUM_CLEAN_MEMBERS>;
  * `power` (DP 151) is part of the shared AIoT product DP schema for every T2xxx clean-line device — not
  * a model-specific extension — so its write is gated by `available: isAiotVacuum` only: no equivalent
  * power DP is confirmed on the legacy Tuya clean line (G-series/X8). The three mode-control verbs
- * (`startCleaning`, `returnToDock`, `pauseCleaning`) are gated by `available: isAiotVacuum ||
- * isTuyaVacuum` and dispatch different DP shapes per platform: legacy Tuya dispatches DP 2 / DP 101
- * (bool), AIoT dispatches DP 152 (ModeCtrlRequest protobuf). Each AIoT mode-control verb carries its
+ * (`startCleaning`, `returnToDock`, `pauseCleaning`) are gated by `available: isAiotVacuum(ctx) ||
+ * isTuyaVacuum(ctx)` and dispatch different DP shapes per platform: legacy Tuya dispatches DP 2
+ * (bool, start=true/pause=false) / DP 101 (bool, return=true), both confirmed from
+ * DeviceHomeModule.java; AIoT dispatches DP 152 (ModeCtrlRequest protobuf). Each AIoT mode-control verb carries its
```
lenoxys: `DeviceHomeModule.java` is cited as behavioural authority here and again at L598 / L615 / L632. Per `AGENTS.md:166-170` shipped `src/` shouldn't name a third-party reverse-engineering project, and `:140-144` won't accept one as authority for a wire.

Separately: L39 (`Go home (DP 101, Bool rw) — confirmed from DeviceHomeModule.java (goHomeCmd)`) is the exact line @max246 reopened on #31 and it's still unfixed. Since this PR edits the file, it'd be a good moment to clear all five occurrences.

src/model/capabilities/suction.ts
```diff
@@ -158,13 +141,14 @@ export const SUCTION: CapabilityModule = {
...
-   * non-AIoT device where no setter exists to consume it.
+   * The per-model level range — sourced from the SKU's `get_product_data_point` catalog for DP 158
+   * when available, falling back to the full 0–5 scale otherwise. Returned unconditionally:
+   * `setSuctionLevel` is only installed on AIoT devices (gated by `available: isAiotVacuum` on the
+   * `level` entry), so `supportedLevels` is harmless data on a non-AIoT device.
    */
   actions(ctx: CommandContext): CapabilityActions {
-    const supportedLevels = (ctx.model ? SUCTION_LEVELS_BY_MODEL[ctx.model] : undefined) ?? ALL_SUCTION_LEVELS;
+    const catalogRange = ctx.dpCatalog?.enumRanges.get(SUCTION_DP.SUCTION);
+    const supportedLevels = (catalogRange as readonly SuctionLevelValue[] | undefined) ?? ALL_SUCTION_LEVELS;
```
lenoxys: **Two problems in one line.**

**1. A live-verified value is replaced by a guess.** The deleted `SUCTION_LEVELS_BY_MODEL` carried `T2351: [0,1,2,3]` annotated *"`get_product_data_point` catalog + live write verified"*. `dp-catalog.ts:4-8` says the response shape *"is not yet confirmed from a live capture"*, so `EMPTY_DP_CATALOG` — and therefore `ALL_SUCTION_LEVELS` — is the **expected** path, not the exceptional one. Net effect: a T2351 host is now told it supports suction levels 4 and 5, which it does not. `AGENTS.md:82-83` is explicit that an absent optional input must degrade to `undefined`, *"never to a guess"*. Returning `undefined` when the catalog is absent would let a host distinguish "unknown" from "all six".

**2. The cast is a lie-cast.** `catalogRange as readonly SuctionLevelValue[]` asserts membership in the `0|1|2|3|4|5` union without checking it. A catalog entry of `[0,1,2,7]` yields `7` typed as a `SuctionLevelValue`. `AGENTS.md:112-115` asks for narrowing with `undefined` on mismatch rather than lie-casting — filtering against `SuctionLevel`'s values would do it.

Unrelated but in this file: `isTuyaVacuum` on L5 is still an unused import (@max246's open nit from #31) — worth dropping while you're here.

lenoxys (commented):
First review on this one. Flagging up front that this PR is **stacked on #31, which still has open review items**, and that the most consequential change here is a **revert of a thread that was resolved there**.

The MQTT topic expansion (`biz/`, `dt/`) and the `parseSecureTopic` widening look sound and well-scoped — no comments on those beyond one note inline.

My concern is concentrated in three places:

1. **The three Tuya mode-control writes that #31 turned into `throw`s are live dispatches again** (`7985728` reverted), re-justified by a third-party decompile, while `X8_VACUUM_DP`'s own doc at L52-53 still reads *"Write direction for all DPs is unverified — no live publishDps capture has been made yet."* They now dispatch into a transport that refuses by default, so the setters are present on the typed surface but fail for any host that hasn't opted in — which `AGENTS.md:126-128` calls out as worse than absent.
2. **A live-verified value was deleted in favour of an unverified dynamic source.** `SUCTION_LEVELS_BY_MODEL` carried `T2351: [0,1,2,3]` marked live-verified; the catalog that replaces it has an admittedly unpinned response shape, so the fallback path (all six levels) is the *expected* one — a regression for the one model actually confirmed.
3. **`fetchDps` reintroduces the all-or-nothing DP drop** that was fixed in #31, and duplicates a parser that already exists.

Plus a handful of dead/speculative additions (`VacuumProductType`, `DpCatalog.dpIds`) that `AGENTS.md:179-180` asks to remove rather than keep as a reference.

Worth knowing: **nothing in `npm run verify` catches any of this** — `guard:docrefs` doesn't police third-party filenames, there's no linter for the unused import or the dead enum, and `property-id-integrity` only walks `properties`. A green gate here isn't evidence.

Not requesting changes formally — that's @max246's call, and this is my first pass on the branch.

src/model/device-family.ts
```diff
@@ -124,6 +124,45 @@ export const isHomeBase = (ctx: FamilyContext): boolean => has(HOMEBASE_TYPES, c
...
+
+/**
+ * Integer product-family constants for the clean line, sourced verbatim from
+ * `ICleanBridgeDeviceInterface.java` (`PRODUCT_TYPE_*`).
+ *
+ * Retained as a reference for capability code that may need to branch on product family.
+ * Per-model feature availability is resolved dynamically via `get_product_data_point` rather
+ * than a static T-code lookup table.
+ */
+export enum VacuumProductType {
```
lenoxys: **Dead code (`AGENTS.md:179-180`).** Nothing in `src/` reads `VacuumProductType` — its only consumer is `device-family.spec.ts`, which asserts the constants equal the values in the Java file. That's a 26-constant enum tested against itself.

The doc says *"Retained as a reference for capability code that **may need** to branch on product family"*, which is the definition of the Speculative Generality the rule targets: *"Remove speculative and dead code rather than keeping old shapes beside new ones."* The PR body notes the T-code map was added and then removed — this enum looks like the leftover from that.

Also worth noting `ICleanBridgeDeviceInterface.java` is a third-party filename in shipped `src/` (`AGENTS.md:166-170`), so deleting the block resolves that too. Happy to be wrong if a follow-up consumes it, but per `AGENTS.md` that's the point at which it should land.

src/transport/tuya/command-router.ts
```diff
@@ -89,13 +96,43 @@ export class TuyaCommandRouter {
...
+   * `null` when neither yields a non-empty record so the caller can skip the delivery safely.
+   */
+  async fetchDps(sn: string): Promise<Record<string, unknown> | null> {
+    const ids = this.snMap.get(sn);
+    if (!ids) return null;
+    await this.ensureLoggedIn();
+    const res = await this.client!.getDeviceDps<Record<string, unknown>>(ids.devId);
+    if (!res.success || !res.result) return null;
+
+    // Try result.dps first (Tuya-native shape), then bare result (flat map).
```
lenoxys: **Reintroduces the all-or-nothing drop fixed on #31, and duplicates an existing parser.**

**The bug:** the second branch returns bare `res.result` with no key filtering. That map goes to `tuyaDpRouter.deliver` → `parseTuyaDpEvent`, which does `return null` on *any* non-integer key. So a single metadata field alongside the DPs (`t`, `protocol`, …) silently discards the entire hydration — the exact failure mode fixed for the AIoT-direct branch on #31.

**The duplication:** `parseTuyaDpReport` in `tuya/dp-codec.ts:62` already owns "defensively extract a DP map from an unpinned Tuya envelope, nested `.dps` first then flat" — *and* filters to integer keys. `AGENTS.md:85-97` asks to search by operation and data shape before adding a second implementation; this is the same operation on the same shape.

Suggest routing the response through `parseTuyaDpReport` (extending it if the `cache.dp.get` envelope differs) so both inbound paths share one parser and one key predicate.

src/transport/tuya/command-router.ts
```diff
@@ -23,6 +24,10 @@ export interface TuyaCommandRouterConfig {
    * confirmation and remove the gate when the write is shipped as verified.
    */
   allowUnverified?: boolean;
+  /** @internal Test-only: inject a custom signer into the underlying TuyaClient. */
```
lenoxys: These two test-only fields sit on an exported config and duplicate ownership that `TuyaClientConfig` already has — it declares `signer` and `http` for exactly this purpose (`client.ts:39/54`), and `TuyaClient` is constructed right here in `ensureLoggedIn`.

`@internal` keeps them out of the docs but not out of the public type, so `_signer` / `_http` remain callable API. A single `clientConfig?: Pick<TuyaClientConfig, "signer" | "http">` — or injecting a client factory — reuses the existing owner instead of mirroring two of its fields under new names (`AGENTS.md:85-97`).

src/model/capabilities/dp-catalog.ts
```diff
@@ -0,0 +1,100 @@
...
+ *  - `dp_id` (Tuya-native integer field) or `id` (alternate field name)
+ *  - `values` as a plain `number[]` array, a JSON-stringified `"[0,1,2,3]"`, or a
+ *    JSON-stringified `"{\"range\":[\"0\",\"1\",\"2\",\"3\"]}"` object
+ *
+ * Returns {@link EMPTY_DP_CATALOG} on any shape mismatch — never throws.
+ */
+export function parseDpCatalog(raw: unknown): DpCatalog {
+  if (!raw || typeof raw !== "object" || Array.isArray(raw)) return EMPTY_DP_CATALOG;
+  const r = raw as Record<string, unknown>;
+  const list = r.data_point_list;
```
lenoxys: The defensiveness is asymmetric in a way that can hide total failure. Below this line the parser accepts `dp_id` **or** `id`, numbers **or** numeric strings, and `values` as an array **or** a JSON string **or** a `{range:[…]}` object — but the one key everything depends on, `data_point_list`, is a single unguarded guess. If the live response names that list anything else, every branch below is unreachable and `parseDpCatalog` returns `EMPTY_DP_CATALOG` forever, silently: `fetchDpCatalog` swallows errors, and `suction` reads the fallback. The feature would be a permanent no-op with a green test suite.

Given the module doc already says *"tighten the field names below and drop the unused branches"* once a real response lands, `AGENTS.md:179-180` would rather see the speculative variants deleted now and one shape supported. At minimum, could the absent/unrecognised case be observable — a debug log in `fetchDpCatalog`, or `undefined` rather than an empty catalog — so this fails loudly instead of quietly?

There's also no test covering the `dpCatalog → supportedLevels` path; the 11 specs all exercise `parseDpCatalog` directly.

src/model/capabilities/dp-catalog.ts
```diff
@@ -0,0 +1,100 @@
...
+ * capabilities fall back to their safe defaults. Once a live response is captured and the shape
+ * confirmed, tighten the field names below and drop the unused branches.
+ *
+ * @module model/capabilities/dp-catalog
+ */
+
+/** The parsed DP capability catalog for one product SKU. */
+export interface DpCatalog {
+  /** DP ids present in this product's catalog. */
+  readonly dpIds: ReadonlySet<number>;
```
lenoxys: `dpIds` has no consumers repo-wide — `suction.ts:150` reads `enumRanges` only. Per `AGENTS.md:179-180` this is the kind of shape to add when the first caller needs it rather than ship ahead of one.

src/client/types.ts
```diff
@@ -129,6 +129,15 @@ export interface EufyMegaOptions extends MegaClientConfig {
...
   ffmpegLogLevel?: FfmpegLevel;
+  /**
+   * Opt into unverified Tuya DP writes for `eufy_home_tuya` clean-line devices (G-series / X8).
+   *
+   * By default `TuyaCommandRouter` refuses to send `dp.publish` because the request shape
+   * has been reversed but not yet confirmed from a live on-device capture — a wrong shape comes back
+   * as a generic Tuya error indistinguishable from an actual device rejection. Set `true` only once
+   * you have confirmed the full round-trip on a real device, or have accepted that ambiguity.
+   */
+  tuyaAllowUnverified?: boolean;
```
lenoxys, line 140: Worth an explicit maintainer decision rather than landing quietly, in the same category as the `TUYA_SIGN_K` flip on #31: this promotes "send writes whose wire shape is unconfirmed" to a **documented public SDK option**.

The hard rule in `AGENTS.md:132-136` is that an unverified write throws and its setter is absent, on the reasoning that fire-and-forget frames make a guess indistinguishable from success. A host-facing opt-in is a reasonable escape hatch, but combined with the change at `vacuum-clean.ts:598` it means a single boolean turns three unverified DP writes into live ones. The doc here is honest about the risk, which helps.

Two smaller points: it reads *"Set `true` only once you have confirmed the full round-trip on a real device"* — that's advice the SDK can't enforce, so a host that flips it is in guess territory the rule exists to prevent. And @max246 asked on #31 for the equivalent policy flip to be called out in the PR body; same would apply here.

src/transport/mqtt/topics.ts
```diff
@@ -82,15 +82,28 @@ export function secureTopic(device: EufyDevice, leg: "req" | "res"): string {
...
  */
 export function subscribeTopics(device: EufyDevice): readonly string[] {
-  if (device.category !== EUFY_LIFE) return [secureTopic(device, "res")];
-  const base = `${EUFY_LIFE}/${device.model}/${device.sn}`;
-  return [`cmd/${base}/app/res`, `cmd/${base}/res`, `synq/${base}/state_info`, `cmd/${base}/app/ota/res`];
+  if (device.category === EUFY_LIFE) {
+    const base = `${EUFY_LIFE}/${device.model}/${device.sn}`;
+    return [`cmd/${base}/app/res`, `cmd/${base}/res`, `synq/${base}/state_info`, `cmd/${base}/app/ota/res`];
+  }
+  if (device.deviceClass === "vacuum" || device.deviceClass === "mower") {
```
lenoxys, line 102: Two notes, neither blocking.

**Prefix is hardcoded rather than record-derived.** This builds on `EUFY_HOME` for any vacuum/mower, including `eufy_home_tuya` devices — so an X8 subscribes `cmd/eufy_home/…` while its `category` says `eufy_home_tuya`. I believe that's *correct* (PR #31's `dp-codec.ts:9` documents inbound arriving on `cmd/eufy_home/{model}/{sn}/res`), but the `eufy_life` branch above derives its prefix from the category and `AGENTS.md:171-174` asks for topics built from the device record. Since this is the one place the two diverge, could the JSDoc state that Tuya-category devices deliberately ride the `eufy_home` prefix? Otherwise it reads like an oversight and someone will "fix" it later.

**Discriminator differs between branches.** The `eufy_life` branch keys on `device.category`, this one on `device.deviceClass`. Consistent enough to work, but a reader has to notice the switch.

src/transport/tuya/command-router.ts
lenoxys: **Correcting myself on the first half of this comment — #31 has since merged and the drop concern no longer applies.**

Commit `2d146f0` on #31 (now merged to `beta-0.1.0` as `4537414`) changed `parseTuyaDpEvent` to skip unrecognised entries rather than reject the whole map:

```ts
if (!/^[1-9]\d*$/.test(key)) continue;
if (typeof value !== "boolean" && typeof value !== "number" && typeof value !== "string") continue;
```

So once this branch is rebased, `fetchDps` handing back an unfiltered `res.result` **cannot** discard the payload — metadata keys like `t` / `protocol` are simply skipped. My "reintroduces the all-or-nothing drop" claim was true against your base (`b61db88`) but is void after a rebase, and I'd rather retract it than leave it standing.

**The duplication half still stands:** `parseTuyaDpReport` (`tuya/dp-codec.ts`) already owns nested-`.dps`-then-flat extraction for an unpinned Tuya envelope, and it now shares the same `/^[1-9]\d*$/` predicate. Routing `cache.dp.get` through it keeps one parser and one key rule for both inbound paths (`AGENTS.md:85-97`) — but that's a tidiness point now, not a bug.

Apologies for the noise on the bug half.

lenoxys:
**Heads-up: #31 merged, so this branch needs a rebase and two of my comments shift.**

#31 landed on `beta-0.1.0` as `4537414`, including the follow-up fix commit `2d146f0`. I verified all twelve review items against the merged tree — they're genuinely fixed, not papered over. Two consequences here:

**1. One of my comments is retracted.** The "all-or-nothing drop" half of the [`fetchDps` thread](https://github.com/mega-yfue/eufy-sdk/pull/33#discussion_r3811816040) is void after a rebase — `2d146f0` made `parseTuyaDpEvent` skip unrecognised keys and values instead of rejecting the map. Details in the thread. The duplication point survives as a tidiness item.

**2. `_signer` / `_http` now collide with a real field.** `2d146f0` added a sanctioned `http?: TuyaHttpPost` to `TuyaCommandRouterConfig` for exactly this purpose, and threads it to the internal client:

```ts
export interface TuyaCommandRouterConfig {
  allowUnverified?: boolean;
  /** Inject a POST transport (test stub); default = native fetch. Forwarded to the internal TuyaClient. */
  http?: TuyaHttpPost;
}
```

So `_http` is now a duplicate of an existing member rather than a new one, and this will conflict textually. Dropping both underscore fields in favour of the merged `http` (plus `signer` if a test needs it) resolves that comment.

**Unchanged and still the blocking item:** the three Tuya mode-control writes. Now that #31 has shipped, re-enabling them isn't reverting a branch — it's reverting released `beta` behaviour, and the merged tree still carries `X8_VACUUM_DP`'s own *"Write direction for all DPs is unverified"* plus the newly honest `param-dictionary` entry (`writable: false`, "Write direction unverified — no live publishDps capture yet"). Three sources now say unverified; this PR would make it four disagreeing with itself.

Also worth knowing for the rebase: `2d146f0` removed `X8_VACUUM_DP.LOOK_FOR_SWEEPER` and the `locatingLegacy` member, and cleared every third-party filename from shipped `src/` — so the `DeviceHomeModule.java` citations this PR adds are re-introducing a class of reference the base branch no longer has anywhere.

src/model/capabilities/vacuum-clean.ts
```diff
@@ -584,50 +584,56 @@ export const VACUUM_CLEAN_MEMBERS = {
...
+        return (): Promise<void> => sink.dispatch(aiotDp(LEGACY_VACUUM_DP.PLAY_PAUSE, true));
       }
       let seq = 111;
       return (): Promise<void> =>
         sink.dispatch(aiotDp(VACUUM_DP.MODE_CTRL, encodeModeCtrl(ModeCtrlMethod.START_AUTO_CLEAN, ++seq)));
     },
-    "Start an auto-clean run (ModeCtrlRequest method 0, DP 152 for AIoT).",
-    (ctx) => isAiotVacuum(ctx),
+    "Start an auto-clean run (DP 152 ModeCtrlRequest for AIoT; DP 2 bool for Tuya).",
+    (ctx) => (ctx.paramIds?.has(LEGACY_VACUUM_DP.PLAY_PAUSE) || ctx.paramIds?.has(VACUUM_DP.MODE_CTRL)) ?? false,
```
max246: **Must-fix — this probably breaks control on real AIoT vacuums.**

The AIoT branch of this guard now requires `paramIds.has(VACUUM_DP.MODE_CTRL)` (DP 152). But 152 is the write-only `ModeCtrlRequest` command DP — devices don't *report* command/method DPs in their param/state set, so 152 is very unlikely to ever land in `paramIds` (which is `rec.params ∪ rec.dpParams`, i.e. reported params).

Note you deliberately kept `power` (DP 151) gated on `isAiotVacuum` *precisely because* "no equivalent power DP is confirmed on the legacy Tuya clean line" / you can't rely on the DP being reported — but then switched start/dock/pause to reported-DP gating on an even-less-reportable write-only DP. That's inconsistent, and the likely outcome is that `startCleaning`/`returnToDock`/`pauseCleaning` silently disappear on AIoT devices.

Suggest: keep the AIoT branch on `isAiotVacuum(ctx)` and only add the reported-DP check for the legacy Tuya path, e.g. `isAiotVacuum(ctx) || ctx.paramIds.has(LEGACY_VACUUM_DP.PLAY_PAUSE)`. Please confirm against a live AIoT capture whether DP 152 actually appears in reported params before relying on it here.

src/model/capabilities/vacuum-clean.ts
```diff
@@ -584,50 +584,56 @@ export const VACUUM_CLEAN_MEMBERS = {
...
+   * AIoT: ModeCtrlRequest method 0 over DP 152.
+   * Tuya: DP 2 = true (PLAY_PAUSE bool, confirmed from protocol inspection).
+   */
   startCleaning: method(
     ({ sink, ctx }) => {
-      if (isTuyaVacuum(ctx)) {
-        return (): Promise<void> => {
-          throw new Error("startCleaning: Tuya write path not yet verified — no live publishDps capture");
-        };
+      if (ctx.paramIds.has(LEGACY_VACUUM_DP.PLAY_PAUSE)) {
```
max246: **Must-fix — advertised-but-throws on Tuya.**

For a Tuya vacuum, `paramIds` has `PLAY_PAUSE`/`GO_HOME`, so the `available` guard passes and this verb is present on the surface. But the dispatch routes through `TuyaCommandRouter`, which throws `UNVERIFIED` whenever `tuyaAllowUnverified` is false — the default. So the surface now advertises verbs that always throw unless the host opted in.

Previously these were `undefined` on Tuya, which was honest. `available` is supposed to mean "callable"; this breaks that contract. Either fold `tuyaAllowUnverified` into the availability decision (so the Tuya-only branch stays hidden until opt-in), or keep the verb hidden and surface the capability some other explicit way. As-is, a host enumerating actions will call `startCleaning()` and get a hard error on the happy path.

src/client/eufy-mega.ts
```diff
@@ -1585,6 +1598,19 @@ export class EufyMega extends EventEmitter {
...
+  private async fetchDpCatalog(model: string): Promise<DpCatalog> {
+    const cached = this.dpCatalogCache.get(model);
+    if (cached) return cached;
+    try {
+      const raw = await this.mega.getProductDataPoint(model);
+      const catalog = parseDpCatalog(raw);
+      this.dpCatalogCache.set(model, catalog);
+      return catalog;
+    } catch {
+      return EMPTY_DP_CATALOG;
```
max246, line 1611: `fetchDpCatalog` only writes the cache on success. On failure it returns `EMPTY_DP_CATALOG` **without caching**, and `commandContext()` runs on every vacuum/mower command — so if `get_product_data_point` errors (and this PR explicitly notes the endpoint's liveness/shape is unverified), every subsequent command re-hits the endpoint over the network.

Cache the empty result on failure too (or a small negative marker) so a failing model is fetched at most once: `this.dpCatalogCache.set(model, EMPTY_DP_CATALOG)` in the catch.

src/transport/tuya/command-router.ts
```diff
@@ -107,6 +108,31 @@ export class TuyaCommandRouter {
...
+    const ids = this.snMap.get(sn);
+    if (!ids) return null;
+    await this.ensureLoggedIn();
+    const res = await this.client!.getDeviceDps<Record<string, unknown>>(ids.devId);
+    if (!res.success || !res.result) return null;
+
+    // Try result.dps first (Tuya-native shape), then bare result filtered to integer DP keys.
+    const nested = (res.result as Record<string, unknown>).dps;
+    if (nested && typeof nested === "object" && !Array.isArray(nested)) {
+      return nested as Record<string, unknown>;
```
max246, line 130: Doc above says this returns `null` "when neither yields a non-empty record," but this branch returns `nested` as soon as it's an object — including an empty `{}`. A device reporting `result.dps = {}` would deliver an empty map instead of `null`, so the caller can't distinguish "no DPs" from "got DPs." Either guard `Object.keys(nested).length > 0` here, or drop the "non-empty" claim from the doc.

src/model/capabilities/dp-catalog.ts
```diff
@@ -0,0 +1,98 @@
...
+  let found = 0;
+  const enumRanges = new Map<number, readonly number[]>();
+
+  for (const entry of list) {
+    if (!entry || typeof entry !== "object" || Array.isArray(entry)) continue;
+    const e = entry as Record<string, unknown>;
+
+    // DP id: try dp_id first, then id
+    const rawId = e.dp_id ?? e.id;
+    const dpId = typeof rawId === "number" ? rawId : typeof rawId === "string" ? parseInt(rawId, 10) : NaN;
```
max246: Nit (parser is billed as "maximally defensive"): `parseInt(rawId, 10)` accepts trailing garbage — `"158abc"` → `158`, `"3px"` → `3` in `toNumberArray`. Prefer `Number(x)` + `Number.isInteger(...)` so malformed strings reject cleanly rather than silently coercing to a plausible-looking id. Same applies in `toNumberArray`. Low priority.

src/transport/mqtt/topics.ts
```diff
@@ -103,17 +119,17 @@ export interface ParsedTopic {
...
+ * positionally rather than from the end — `…/<sn>/app/res` and `…/<sn>/app/ota/res` would otherwise
+ * yield the tail segment as the device id. Returns `undefined` on any shape this doesn't recognise,
+ * so an unparsed topic leaves `deviceSn` unset instead of carrying a guess.
  */
 export function parseSecureTopic(topic: string): ParsedTopic | undefined {
   const parts = topic.split("/");
   if (parts.length < 5) return undefined;
   const [root, category, model, sn] = parts;
-  if (root !== "cmd" && root !== "synq") return undefined;
+  if (root !== "cmd" && root !== "synq" && root !== "biz" && root !== "dt") return undefined;
```
max246, line 132: Question, not a blocker: `parseSecureTopic` now accepts `biz/.../req`, which per the topics doc is "cloud ACKs for app→cloud requests" — i.e. echoes of our own writes, not device state. Does the downstream `SecureMqtt` message handler branch on `root`/`tail` so these `req` ACKs don't get fed into the DP-report / device-state path? That handler isn't in this diff; worth confirming an ACK on `biz/.../req` isn't mistaken for an inbound device DP report.

max246 (commented):
Reviewed the clean-line work — solid direction and the defensive parsing/tests are good. Two things I think are must-fixes before merge (both are regressions in what the capability surface promises), plus a few smaller notes. Left them inline.

**Must-fix**
1. Mode-control verbs now gated on the device *reporting* DP 152, which is a write-only command DP — likely never in `paramIds`, so start/dock/pause may vanish on real AIoT vacuums.
2. On Tuya, the same verbs are advertised as available but throw `UNVERIFIED` by default — `available` no longer means "callable."

**Smaller**
3. `fetchDpCatalog` doesn't cache failures → a network call on every command when the (admittedly unverified) endpoint errors.
4. `fetchDps` nested branch can return an empty `{}`, contradicting its own doc.
5. `parseDpCatalog` uses `parseInt` → silently accepts trailing garbage.
6. Question about `biz/.../req` ACKs being fed into the inbound path.

src/model/capabilities/vacuum-clean.ts
martijnpoppen: Fixed in `cdd225f`. All three verbs now gate on `isAiotVacuum(ctx)` (category check) instead of `paramIds.has(MODE_CTRL)`. Since DP 152 is write-only and not present in reported params, the old guard was silently hiding the verbs on all AIoT devices — the exact opposite of the intent. Category-based guard is consistent with how `setPower` has always been gated.

---

src/model/capabilities/vacuum-clean.ts
martijnpoppen: Fixed in `cdd225f`. The Tuya write path (`PLAY_PAUSE`/`GO_HOME` dispatch) has been removed from `startCleaning`, `returnToDock`, and `pauseCleaning`. The verbs are now gated on `isAiotVacuum` only — absent on Tuya until a live `publishDps` capture verifies the direction. The spec test asserting Tuya dispatch is gone; new spec asserts the verbs are `undefined` on `"eufy_home_tuya"` category.

---

src/client/eufy-mega.ts
martijnpoppen: Fixed in `cdd225f`: `this.dpCatalogCache.set(model, EMPTY_DP_CATALOG)` is now called in the `catch` block before returning, so a failing endpoint is hit at most once per session regardless of how many commands follow.

---

src/transport/tuya/command-router.ts
martijnpoppen: Fixed in `cdd225f`: added `&& Object.keys(nested).length > 0` to the guard so an empty `dps: {}` falls through to the flat-extraction path rather than being returned as a success with an empty map.

---

src/model/capabilities/dp-catalog.ts
martijnpoppen: Fixed in `cdd225f`. Both `parseDpCatalog` and `toNumberArray` now use `Number(x)` + `Number.isInteger()` — `"158abc"` → `NaN`, rejected cleanly. `Number.isInteger` also catches non-integer floats (`"3.5"` → `3.5`, not an integer), so the guard is tighter than the original `!Number.isFinite`.

---

src/transport/mqtt/topics.ts
martijnpoppen: Confirmed safe. In `eufy-mega.ts` the MQTT `"message"` handler passes everything to `processAvailabilityMessage(m.topic, m.raw)`, which resolves to `parseStateInfoSignal` in `availability.ts`. That function gates on `parsedTopic?.root !== "synq"` — so a `biz/…/req` ACK (root = `"biz"`) returns `undefined` and is a no-op. The Tuya-DP branch gates on `parseTuyaDpReport(m.raw)` returning a non-null map (payload-shape check, not topic check), and `parseAiotDpReport`/`parseDpMessage` similarly produce null for a non-DP payload. So a `biz/…/req` ACK flows through and produces zero state mutations.

---

src/model/capabilities/vacuum-clean.ts
martijnpoppen: Fixed in `cdd225f`. The Tuya dispatch paths have been removed entirely — `startCleaning`, `returnToDock`, and `pauseCleaning` are now gated on `isAiotVacuum` and dispatch only via DP 152 `ModeCtrlRequest`. The tautology (`isAiotVacuum || isTuyaVacuum`) is gone. No live `publishDps` capture exists for the Tuya direction; the verbs stay absent on Tuya until one is available.

---

src/client/types.ts
martijnpoppen: Agreed — the `tuyaAllowUnverified` option is an escape hatch that deserves an explicit decision. With the changes in `cdd225f` the practical scope has narrowed considerably: the three control verbs (`startCleaning`/`returnToDock`/`pauseCleaning`) are now absent on Tuya entirely, so `tuyaAllowUnverified` no longer unlocks those. The only remaining surface the flag gates is anything a host dispatches directly (e.g. via the low-level API), not a named capability verb. I'll note this explicitly in the PR body.

---

src/transport/mqtt/topics.ts
martijnpoppen: The JSDoc was added in an earlier commit (the `// eufy_home_tuya devices deliberately use the eufy_home prefix` comment at the subscription site). On the discriminator: the `eufy_home` branch uses `deviceClass` because `category` for some eufy_home_tuya devices still reads `eufy_home_tuya` rather than `eufy_home` — `deviceClass` is the stable clean-line indicator. A comment clarifying this has been in the file since the clean-integration work. Happy to expand it if it still reads as an oversight.

---

src/model/capabilities/vacuum-clean.ts
martijnpoppen: Fixed in `cdd225f`. All five `DeviceHomeModule.java` citations have been removed from `src/model/capabilities/vacuum-clean.ts`. The file no longer names any third-party reverse-engineering artefact in shipped code.

---

src/model/capabilities/suction.ts
martijnpoppen: Fixed in `cdd225f`. Three changes:

1. **Lie-cast replaced**: `catalogRange as readonly SuctionLevelValue[]` → `catalogRange?.filter((v): v is SuctionLevelValue => v in SUCTION_LABEL)`. A value outside `0|1|2|3|4|5` is silently dropped rather than typed as a member.
2. **`supportedLevels` returns `undefined` when catalog absent**: The `ALL_SUCTION_LEVELS` fallback is gone — when `ctx.dpCatalog` is absent or has no DP 158 entry, `supportedLevels` is `undefined`, consistent with `AGENTS.md:82-83`.
3. **`isTuyaVacuum` unused import removed** from `suction.ts`.

`ALL_SUCTION_LEVELS` is still present in the file, but only as `decodedValues: ALL_SUCTION_LEVELS` on the `level` member — a compile-time annotation of the getter's return union. It is not a runtime fallback.

The missing `dpCatalog → supportedLevels` tests (the prior 11 specs all hit `parseDpCatalog` directly) have been added in `b93bc1d`: four cases covering catalog range present, unknown values filtered out, catalog present with no DP 158 entry, and no catalog (returns `undefined`). All bind through `buildActions` so they exercise the full `actions(ctx) → supportedLevels` path.

---

src/model/device-family.ts
martijnpoppen: Deferring to a follow-up. The concern is understood — `VacuumProductType` has no current `src/` consumer beyond its own spec, and is the clearest example of the speculative-generality rule. Removing it is agreed in principle; keeping it in this PR would enlarge a diff already touching many files. It will land as a standalone cleanup once this PR closes.

---

src/transport/tuya/command-router.ts
martijnpoppen: Thank you for the self-correction — noted and appreciated. The duplication point still stands: `parseTuyaDpReport` in `tuya/dp-codec.ts` already owns nested-`.dps`-then-flat extraction, and routing `cache.dp.get` through it would unify the key predicate. Agreed as a tidiness improvement.

The current inline extraction in `fetchDps` is kept for now because the `cache.dp.get` envelope shape is still unconfirmed from a live capture — the two inbound paths may diverge once it is pinned, and consolidating before the shape is known risks needing to re-split. Once a live response confirms the shape, routing through a shared parser is the right call.

---

src/transport/tuya/command-router.ts
martijnpoppen: Fixed in `cdd225f`. `TuyaCommandRouterConfig` now has only `allowUnverified` and `http` — `_signer` and `_http` have been removed. The `http` field is passed directly to `TuyaClient` construction in `ensureLoggedIn` (`new TuyaClient({ signer: new HmacSigner(), http: this.http })`), which is the existing owner.

---

src/model/capabilities/dp-catalog.ts
martijnpoppen: The parser already handles both `data_point_list` and `dp_list` (line 41 of `dp-catalog.ts` — `r.data_point_list ?? r.dp_list`). The "silent empty catalog" concern is valid: a missing key degrades quietly, which could mask a shape change. The current trade-off is resilience over loudness until the live response shape is confirmed; the module doc flags this explicitly — *"tighten the field names and drop unused branches"* once a real capture lands. At that point silent fallback becomes a confirmed no-op rather than a hidden failure.

The missing `dpCatalog → supportedLevels` tests have been added in `b93bc1d`: four cases covering catalog range present (with filtering to valid `SuctionLevelValue` entries), unknown values filtered out, catalog present but no DP 158 entry, and no catalog (`supportedLevels` is `undefined`). All bind through `buildActions` via the `bind()` helper, so they exercise the full `actions(ctx) → supportedLevels` path rather than calling `parseDpCatalog` directly.

---

src/model/capabilities/dp-catalog.ts
martijnpoppen: Fixed in `cdd225f`. `dpIds` has been removed from `DpCatalog` — the interface now has only `enumRanges`, which is the only field any consumer reads.

---

max246 (approved):

PR #4, opened 2026-08-05
test: merge policy (should be allowed)

lenoxys:
Temporary. Verifying that a beta-X.Y.Z branch is accepted.

PR #34, opened 2026-08-19
feat(clean): Tuya clean-line read surface, DP-aware availability, and the X8 Pro sensor DPs

martijnpoppen:
> **Body rewritten** after @lenoxys and @max246 both flagged that it described a fraction of the diff. This now lists everything the change set actually does. The write paths that were in earlier revisions are gone — see *Unverified or deferred*.

## What changed

Five things, not one:

**1. Six new read members on `vacuum_clean`** (Tuya clean line, all read-only):

| member | DP | source |
|---|---|---|
| `lifetimeCleanTime` | 119 | schema (`ClearTotalTime`) |
| `lifetimeCleanArea` | 120 | schema (`ClearTotalArea`), `unit: "m²"` / `kind: "squaremeters"` |
| `waterTank` | 127 | schema (`water_tank_status`) |
| `mopPad` | 129 | schema (`mop_status`) |
| `doNotDisturb` | 107 | live read (`forbid_mode`) — **read-only** |
| `rssi` | 134 | schema, `unit: "dBm"` / `kind: "dbm"` |

**2. `X8_*` → `TUYA_*` rename.** `X8_VACUUM_DP` → `TUYA_VACUUM_DP`, and the value-set consts and their types alongside. These DPs are the `eufy_home_tuya` category's schema, not one model's, and the old name implied otherwise.

**3. `paramIds` reaches the resolve-time availability gate.** `AvailabilityContext` gains `paramIds`, `registry.ts` populates it from the record's params, and `members.ts`'s doc is corrected — a gate may now read any field a `CloudRecord` carries, but still never a transport-only one like `channel`. This is what lets a member say "only when this device actually reported that DP".

**4. `locate` owns both locate DPs.** DP 103 becomes a `readAlias` on `locate`'s own read rather than a second `findRobot` verb on `vacuum_clean`. One physical feature, one name.

**5. `squaremeters` added** to `KNOWN_VALUE_KINDS` / `UNIT_FOR_KIND`, so `lifetimeCleanArea` can declare its unit in the schema rather than in prose.

## Why

The reads were already being collected by `decodeState` and thrown away — `TUYA_VACUUM_DP` carried 119/120/127/129/134 before this branch. Surfacing them is a cheap win with no new protocol work, which is what @lenoxys verified at the base commit.

The rename and the `locate` consolidation are both one-owner cleanups: the DP set belongs to a category, and a feature should not be reachable under two names split by platform.

## Evidence

- **119, 120, 127, 129, 134** — `thing.m.device.ref.info.list` v5.4 `schemaInfo.schema` for product `wahqax6ifjgs1c4n` (X8 Pro, T2266). Schema-sourced, so their dictionary entries are `observed: false` and their notes say "Schema-confirmed", not "Live-confirmed".
- **107** — read live from an idle report on the Tuya line. Read direction only.
- Nothing here claims a write is confirmed. The module's own `TUYA_VACUUM_DP` doc still states Tuya write direction is unverified, and that is still true.

On shipping typed getters for `observed: false` DPs, per @lenoxys' thread: getters are evidence-gated at bind time, so on a device that never reports DP 119 the getter is simply not installed. A product schema is also the device's own declaration rather than a guessed id. That is the acceptable end of `AGENTS.md:137-139` rather than a violation — recorded here so `observed: false` members are not taken as precedent for genuinely guessed ones.

## Verified how

- [x] `npm run verify` is green locally (Node 24.5.0) — 121 files, 2385 tests

## Unverified or deferred

**Every Tuya write path that earlier revisions of this branch added has been removed** (2795849), which is what @max246's remaining must-fix asked for:

- `startCleaning` / `returnToDock` / `pauseCleaning` no longer dispatch DP 2 / DP 101. AIoT-only, gated on `isAiotVacuum`.
- `locate` no longer dispatches DP 103. AIoT-only.
- `doNotDisturb` has no `write:`. Read-only.

Those dispatches routed through `TuyaCommandRouter`, which refuses unverified writes unless the host sets `tuyaAllowUnverified` (default `false`) — so they were `available` and threw when called. `tuyaAllowUnverified` never reaches `AvailabilityContext`, so folding it into the gate would mean plumbing a transport setting into `model/`; dropping the dispatch keeps the decision where the evidence is.

Also restored to the reviewed #33 state in the same commit:

- `power.available` back to `isAiotVacuum` — DP 151 is shared-schema, not a reported DP, and is `observed: false` in our dictionary.
- `battery` / `cleanType` readAliases regain `available: isTuyaVacuum`.

**Four specs asserted the removed behaviour, and two encoded a premise no device can satisfy** — the reason CI stayed green over a real break. The "Tuya path" write tests passed `category: undefined`, which makes `isAiotVacuum` **true**, so they never described a Tuya device. `model.spec`'s reresolve test used a `T2351` declared `category: "eufy_home_tuya"` — an X10 Pro Omni on the Tuya line, which does not exist. All four now use coherent devices and assert the current contract.

Still open:

- **No Tuya clean-line write is offered at all.** Confirming one needs a `publishDps` capture on a T2266.
- **`rssi` stays on `vacuum_clean`.** Per @lenoxys' thread: DP 134 is clean-line-specific and a different wire id from the security line's 1141, and there is no shared connectivity capability for vacuums yet. Happy to move it when one exists.
- The six new reads have **not** been observed on a live device; they are schema-sourced and runtime-gated.


lenoxys (commented):
First review here too. The core idea — surfacing DPs that `decodeState` already collects but that never reached callers — is a genuinely cheap win, and I verified the claim: `X8_VACUUM_DP` already carries 119/120/127/129/134 at the base commit, so no new protocol work is involved for the reads.

**The blocking issue is that the description doesn't match the diff.** The body says *"Four schema-confirmed X8 Pro Tuya DPs"* and *"All four are read-only (`writable: false`); **no write path added**"*. The diff adds **seven** members, two of which are write paths:

| In the body | Actually added |
|---|---|
| `lifetimeCleanTime`, `lifetimeCleanArea`, `waterTank`, `mopPad` | those four, plus **`doNotDisturb`** (a `write:` on DP 107), **`findRobot`** (dispatches DP 103), and **`rssi`** |

The "Verified how" and "Unverified or deferred" sections both omit all three. Someone reviewing from the body approves a read-only surface change and merges two write wires — which matters given #31's history on exactly this rule. Could the body be brought in line, or the writes split into their own PR? The reads look mergeable much sooner than the writes.

Two other threads inline: **`findRobot` duplicates a verb `locate` already owns** on the same DP, and the evidence cited for `doNotDisturb` is from a device on which that member cannot exist.

Same caveat as on #33: `npm run verify` is green and catches none of this. Not requesting changes formally — @max246's call.

src/model/capabilities/vacuum-clean.ts
```diff
@@ -587,6 +594,94 @@ export const VACUUM_CLEAN_MEMBERS = {
...
+    provenance: "mega",
+    description: "WiFi RSSI in dBm (DP 134, Value ro). X8 Pro Tuya clean line. Schema-confirmed.",
+    available: isTuyaVacuum,
+  },
+  /**
+   * Locate the robot by triggering its buzzer (DP 103 = true). Schema-confirmed from
+   * `thing.m.device.ref.info.list` v5.4. The device responds with a short audible tone so the user
+   * can find it under furniture. Tuya clean line only — no equivalent DP confirmed on AIoT.
+   */
+  findRobot: method(
```
lenoxys: **This duplicates a verb that already exists on another capability, on the same DP.**

DP 103 is owned by `locate.ts`: `LEGACY_LOCATE_DP = 103` (L9), the `locatingLegacy` read member (L44), and its `decodeState` pickup (L91). `AGENTS.md:34-36` states each capability owns its own wire ids *"because every feature-command and parameter id is used by exactly one capability"*.

The user-visible result after this PR is one physical feature reachable under two names, split by platform:

- AIoT device → `dev.locate()?.locate()` (dispatches DP 160)
- Tuya device → `dev.vacuumClean()?.findRobot()` (dispatches DP 103)

`locate()` is gated `isAiotVacuum` (`locate.ts:69`) and `findRobot` is gated `isTuyaVacuum`, so a host supporting both lines has to know both spellings and branch on which one exists. That's the shape `AGENTS.md:85-97` asks to avoid — and it's the same argument that turned `batteryLegacy` into a `readAlias` on #31.

`locate` is the natural owner: it already holds DP 103, and its `locate(on = true)` method is a better fit than a bare `findRobot()` since the underlying DP is a bool that can also cancel. Suggest extending `locate`'s guard to the Tuya line and dropping `findRobot` here.

Worth noting the one-owner guard can't catch this: `property-id-integrity.spec.ts:93` walks `properties`, and a `method` member isn't one.

Separately — the write itself: the evidence given is *"Schema-confirmed from `thing.m.device.ref.info.list`"*, which is a schema listing, not a write capture, while `X8_VACUUM_DP`'s doc at L52-53 still says *"Write direction for all DPs is unverified."* Per `AGENTS.md:132-136` this should be `unverified` with no dispatch until a `publishDps` capture exists.

src/model/capabilities/vacuum-clean.ts
```diff
@@ -587,6 +594,94 @@ export const VACUUM_CLEAN_MEMBERS = {
...
+    kind: "boolean",
+    provenance: "mega",
+    description: "Mop pad attached (DP 129, Bool ro). X8 Pro Tuya clean line. Schema-confirmed.",
+  },
+  /**
+   * Do-not-disturb mode (DP 107, Bool rw). Live-confirmed `false` from a T2351 idle report.
+   * When `true` the robot suppresses voice announcements; the app allows toggling this from its
+   * settings screen. Tuya clean line only — no equivalent DP confirmed on AIoT.
+   */
+  doNotDisturb: {
```
lenoxys, line 650: **The cited evidence is from a device where this member cannot exist.**

The doc says *"Live-confirmed `false` from a **T2351** idle report"*, but:

- `registry.ts:143` → `T2351: { codec: "vacuum", name: "Clean X10 Pro Omni" }`
- `registry.ts:131` → `T2266: { codec: "vacuum", name: "RoboVac X8 Pro" }`
- `device-family.ts:181` → the `"eufy_home_tuya"` category is confirmed from a **T2266**
- this file's own L48 → *"Tuya DP ids for the X8 Pro (**T2266**)"*

This member is gated `available: isTuyaVacuum`, so on a T2351 — an AIoT device — it would never be installed. Either the T-code in the evidence is wrong, or the observation came from the AIoT line and doesn't ground a Tuya-gated DP. (The PR body has the same conflation: *"the X8 Pro (T2351)"*.) `AGENTS.md:140-144` asks every wire claim to be grounded in evidence you can point at, so it'd be good to pin which device this actually came from.

**Also — the write direction isn't verified by this evidence.** An *idle report* reading `false` confirms the **read**. `AGENTS.md:136` is explicit: *"Verification is per direction; don't flatten it."* Shipping `write:` on the strength of a read observation is the pattern #31 blocked twice, and L52-53 of this file still declares Tuya writes unverified. Combined with #33's `tuyaAllowUnverified` defaulting to `false`, `setDoNotDisturb` is a setter that exists on the type and throws at the transport for any default-configured host — `AGENTS.md:126-128`.

The read half looks fine to ship now; it's just the `write:` that needs its own capture.

src/model/capabilities/vacuum-clean.ts
```diff
@@ -587,6 +594,94 @@ export const VACUUM_CLEAN_MEMBERS = {
...
+    provenance: "mega",
+    description:
+      "Lifetime total cleaning time in seconds from DP 119 (ClearTotalTime). X8 Pro Tuya clean line. Schema-confirmed.",
+  },
+  /**
+   * Lifetime total cleaned area in m² (DP 120, Value). Counts across all sessions.
+   * Confirmed from `thing.m.device.ref.info.list` v5.4 schemaInfo.schema (X8 Pro,
+   * product `wahqax6ifjgs1c4n`). Read-only accumulator — no write expected.
+   */
+  lifetimeCleanArea: {
```
lenoxys, line 614: `kind`/`unit` is inconsistent with its sibling. `lifetimeCleanTime` directly above declares `unit: "s", kind: "seconds"`, but this one carries m² in prose only — `kind: "scalar"` and no `unit`. `rssi` at L665 has the same gap (dBm in the description, `kind: "scalar"`).

`AGENTS.md:125` asks every property to declare what its value **means**, not just how it's stored, and a host offering these as readings has to parse the description to learn the unit. If there's no `kind` for area, `unit: "m²"` alongside would at least carry it in the schema.

src/model/capabilities/vacuum-clean.ts
```diff
@@ -587,6 +594,94 @@ export const VACUUM_CLEAN_MEMBERS = {
...
+    provenance: "mega",
+    description: "Do-not-disturb mode (DP 107, Bool rw). X8 Pro Tuya clean line. Live-confirmed.",
+    available: isTuyaVacuum,
+    write: (v: unknown) => aiotDp(X8_VACUUM_DP.FORBID_MODE, asBool(v)),
+  },
+  /**
+   * WiFi RSSI in dBm (DP 134, Value ro). Schema-confirmed from `thing.m.device.ref.info.list` v5.4.
+   * Negative integer; closer to zero is stronger. Useful for diagnostics. Tuya clean line only.
+   */
+  rssi: {
```
lenoxys, line 662: `rssi` isn't mentioned anywhere in the PR description — the body's table lists four members and this is a fifth. Not a problem in itself (it's a read-only diagnostic, and DP 134 is registered in the dictionary), just worth adding so the change set is reviewable from the body.

One design question: signal strength is a device-wide reading rather than a cleaning concern, and `property-id-integrity.spec.ts:61-66` already treats `rssi` (id 1141) as a `SHARED_READS` id surfaced by several capabilities on the security line. Is `vacuum_clean` the right owner, or should this sit wherever the equivalent security-line reading lives so hosts read "signal" from one place per device?

src/model/param-dictionary.ts
```diff
@@ -3554,4 +3554,49 @@ export const CLEAN_PARAMS: Record<number, ParamDef> = {
     observed: true,
     note: "X8 Pro DP 113 (CleanType, Enum: Sweep|SweepMop|Mop). Live-confirmed Sweep.",
   },
+  119: {
```
lenoxys, line 3557: These five entries are honest about their standing (`observed: false`, "Schema-confirmed"), which I appreciate — but it does mean the PR adds five **typed getters** for DPs no device has been seen to report, and `AGENTS.md:137-139` says *"Never ship a guessed parameter as a typed getter. Give a member a parameter id only when a real device reports it."*

The mitigating factor, which I think makes this defensible: getters are evidence-gated at bind time, so on a device that never reports DP 119 the getter simply isn't installed — the typed surface advertises a capability the runtime withholds, rather than returning a fabricated value. A product schema is also a much stronger source than a guessed id, and it's the device's own schema.

So this reads as the acceptable end of that rule rather than a violation — but since the rule's wording is absolute, it's worth stating the reasoning in the PR body so the next reader doesn't take `observed: false` members as precedent for genuinely guessed ones. Would also be good to confirm the `type`/`kind` pairs survive the Tuya inbound path: bools arrive as real booleans through `parseTuyaDpEvent` but as `"1"`/`"0"` strings through `parseAiotDpReport`, and `waterTank`/`mopPad` are declared `type: "bool"`.

src/model/capabilities/vacuum-clean.ts
```diff
@@ -369,9 +369,12 @@ export function decodeVacuumActivity(raw: ParamValue | undefined, codec: RawDpCo
  * Bound RoboVac reads and controls — the object returned by `dev.vacuumClean()`.
  *
  * All reads, `setPower`, and the three mode-control verbs are DERIVED from `VACUUM_CLEAN_MEMBERS`.
- * Each getter is present only when the device reports the backing DP. `setPower` and the mode-control
- * verbs are optional — they are absent on any device whose `category` is not a confirmed AIoT string
- * (see `isAiotVacuum` in `device-family.ts`).
+ * Each getter is present only when the device reports the backing DP. `setPower` is absent on
+ * `eufy_home_tuya` devices (no confirmed power DP). The three mode-control verbs are present on
```
lenoxys: This now asserts *"The three mode-control verbs are present on both AIoT and Tuya devices"* as a plain fact, but on the parent branch (#33) those Tuya dispatches go to a transport that refuses unless the host sets `tuyaAllowUnverified`. So "present" is true of the type and not of the behaviour — I've raised the underlying revert on #33; flagging here only because this doc is what a caller reads.

The sentence below is affected by the same stack issue: X8 read members are described as populated *"over MQTT or the initial Tuya DP poll"*, and that poll (`TuyaCommandRouter.fetchDps`, added in #33) currently drops the whole payload if the response carries any non-integer key — so the poll half of that promise may not hold yet.

src/model/capabilities/vacuum-clean.ts
```diff
@@ -389,6 +392,10 @@ export type VacuumCleanActions = Surface<typeof VACUUM_CLEAN_MEMBERS>;
  * objects both starting at 112 do not cause the device to complain, so the seq is not enforced as
  * globally monotonic).
  *
+ * Tuya-only additions (gated by `available: isTuyaVacuum`): `doNotDisturb` (DP 107, Bool rw,
```
lenoxys: This block is the clearest statement of what the PR actually does — `doNotDisturb`, `rssi` and a `findRobot` action — and it contradicts the PR body's *"All four are read-only ... no write path added"*. Worth reconciling the two; the code comment is the accurate one.

lenoxys:
**Heads-up: #31 merged, and this branch won't compile once rebased.**

#31 landed on `beta-0.1.0` as `4537414` (including the follow-up fix commit `2d146f0`). I verified all twelve review items against the merged tree — genuinely fixed. One of those fixes removes something this PR depends on.

**`findRobot` references a constant that no longer exists.** `2d146f0` deleted `LOOK_FOR_SWEEPER: 103` from `X8_VACUUM_DP`, because DP 103 was dead in that module while `locate.ts` already owned the id. On merged `beta-0.1.0`:

```
$ grep -n "LOOK_FOR_SWEEPER" src/model/capabilities/vacuum-clean.ts
(no match)
```

So `sink.dispatch(aiotDp(X8_VACUUM_DP.LOOK_FOR_SWEEPER, true))` is a type error after a rebase. This turns the [`findRobot` thread](https://github.com/mega-yfue/eufy-sdk/pull/34#discussion_r3811822707) from a design suggestion into the required resolution — and conveniently, #31 already built the mechanism:

```ts
// locate.ts, merged beta-0.1.0
locating: {
  param: LOCATE_DP,
  readAliases: [{ paramType: LEGACY_LOCATE_DP, available: isTuyaVacuum }],
  description: "Find-robot trigger (DP 160 AIoT / DP 103 Tuya). …",
}
```

DP 103 is already decoded under `locate`'s own key on Tuya devices. What's missing is only the *write* side: widening `locate`'s method guard beyond `isAiotVacuum` so `locate()` is the one find-robot verb on both lines. That's strictly less code than `findRobot`, keeps one owner for DP 103, and drops the two-names-one-feature split.

(The same rebase also removes the `locatingLegacy` member this branch still carries.)

**Still open from the review, independent of the rebase:** the PR body says four read-only DPs and "no write path added" while the diff adds `doNotDisturb`'s `write:` and `findRobot`; and the T2351-vs-T2266 evidence conflict on `doNotDisturb`. Note the merged tree now also has `param-dictionary` DP 2 corrected to `writable: false` with *"Write direction unverified — no live publishDps capture yet"*, so the base branch is now explicit that Tuya writes aren't confirmed — worth reconciling before adding two more.

The four read members themselves look fine and would merge cleanly on their own.

src/model/capabilities/vacuum-clean.ts
martijnpoppen: Removed in an earlier commit — `findRobot`, DP 160, and DP 103 entries are all gone from this file. `locate` remains the sole owner of both locate DPs.

---

src/model/capabilities/vacuum-clean.ts
martijnpoppen: Fixed in c99e1cf. `write:` removed — `doNotDisturb` is now read-only. Description updated to `"Bool ro"`. The JSDoc says `"Read confirmed from a live idle report; write direction unverified — no publishDps capture yet"`.

The T2351 mention was only in the original PR body, not in the code. The property description correctly says `"X8 Pro Tuya clean line"` (T2266) throughout. Will reconcile the T-code in the PR body.

---

src/model/capabilities/vacuum-clean.ts
martijnpoppen: Fixed in c99e1cf. `lifetimeCleanArea` now carries `unit: "m²"` and `kind: "squaremeters"` — `"squaremeters"` was added to `KNOWN_VALUE_KINDS` and `UNIT_FOR_KIND` to satisfy the kind/unit contract enforced by `value-kinds.spec.ts`. `rssi` already had `unit: "dBm"` and `kind: "dbm"` before this commit — consistent with its sibling `lifetimeCleanTime`.

---

src/model/capabilities/vacuum-clean.ts
martijnpoppen: Keeping `rssi` in `vacuum_clean` for now. Tuya vacuum DP 134 is specific to the clean-line schema and a different wire id from the security-line DP 1141. There is no shared "network" or "connectivity" capability for vacuums yet. Happy to move it if one is established in a follow-up.

---

src/model/param-dictionary.ts
martijnpoppen: Agreed on the reasoning — will add a note to the PR body.

The mitigating factor you named is the key: evidence-gated getters are never installed for unreported DPs, so `observed: false` reads as "schema-sourced, runtime-gated" rather than "fabricated value shipped to a caller". A product schema is also stronger than a guessed id.

On the bool path: `waterTank` and `mopPad` are gated by `TUYA_VACUUM_DP` params and can only arrive through `parseTuyaDpEvent`, which delivers native JS booleans (`typeof value !== "boolean"` guard at dp-codec.ts line 31). The `"1"/"0"` string path applies only to the AIoT pipeline, which never sees these DPs.

---

src/model/capabilities/vacuum-clean.ts
martijnpoppen: Doc fixed in c99e1cf — the verbs are now described as `isAiotVacuum`-gated (AIoT only, Tuya write unverified), no Tuya dispatch mention anywhere.

On `parseTuyaDpEvent`: it *skips* non-integer keys via `continue` (dp-codec.ts line 30: `/^[1-9]\d*$/.test(key)`) rather than aborting the whole loop. It only returns `null` when no integer-keyed DPs survived iteration. Non-integer keys in the cache response are safely ignored and the integer-keyed DPs are still delivered to the caller.

---

src/model/capabilities/vacuum-clean.ts
martijnpoppen: Removed along with `findRobot` itself in an earlier commit. The doc no longer references it.

---

src/model/capabilities/vacuum-clean.ts
```diff
@@ -577,42 +583,140 @@ export const VACUUM_CLEAN_MEMBERS = {
...
+        return (): Promise<void> => sink.dispatch(aiotDp(LEGACY_VACUUM_DP.PLAY_PAUSE, true));
+      }
       let seq = 111;
       return (): Promise<void> =>
         sink.dispatch(aiotDp(VACUUM_DP.MODE_CTRL, encodeModeCtrl(ModeCtrlMethod.START_AUTO_CLEAN, ++seq)));
     },
-    "Start an auto-clean run (ModeCtrlRequest method 0 over DP 152).",
-    isAiotVacuum,
+    "Start an auto-clean run (DP 2 = true for Tuya; DP 152 ModeCtrlRequest for AIoT).",
+    (ctx) => (ctx.paramIds?.has(LEGACY_VACUUM_DP.PLAY_PAUSE) || ctx.paramIds?.has(VACUUM_DP.MODE_CTRL)) ?? false,
```
max246: **Must-fix — this reverts the round-2 fix for my #1 and re-breaks AIoT control.**

This gates the AIoT path on `ctx.paramIds.has(VACUUM_DP.MODE_CTRL)` (DP 152) again. In your own reply resolving my #1 on #33 you wrote: *"Since DP 152 is write-only and not present in reported params, the old guard would have made the verbs disappear."* That's exactly this guard. On a real AIoT vacuum, 152 is never in `paramIds` and DP 2 (`PLAY_PAUSE`) is Tuya-only — so `startCleaning`/`returnToDock`/`pauseCleaning` all resolve to `undefined`, i.e. AIoT loses every control verb.

The reason CI stays green: the new spec seeds `aiotDps = new Set([POWER, MODE_CTRL])` and asserts the verbs are present — but that set can't occur on a real device (152 is write-only). The test encodes the false premise, so it can't catch the regression.

Keep the AIoT branch on `isAiotVacuum(ctx)` (the #33 round-2 state) and only DP-gate the *legacy Tuya* path: `isAiotVacuum(ctx) || ctx.paramIds.has(LEGACY_VACUUM_DP.PLAY_PAUSE)`. Same for `returnToDock`/`pauseCleaning`. And fix the spec to seed a realistic AIoT param set (no 152).

src/model/capabilities/vacuum-clean.ts
```diff
@@ -577,42 +583,140 @@ export const VACUUM_CLEAN_MEMBERS = {
...
+   * the app allows toggling this from its settings screen. Tuya clean line only — no equivalent
+   * DP confirmed on AIoT. Schema-confirmed from `thing.m.device.ref.info.list` v5.4 (forbid_mode).
+   */
+  doNotDisturb: {
+    param: TUYA_VACUUM_DP.FORBID_MODE,
+    type: "bool",
+    kind: "boolean",
+    provenance: "mega",
+    description: "Do-not-disturb mode (DP 107, Bool rw). X8 Pro Tuya clean line. Live-confirmed.",
+    available: (ctx: AvailabilityContext) => ctx.paramIds?.has(TUYA_VACUUM_DP.FORBID_MODE) ?? false,
```
max246, line 656: **Must-fix — advertised-but-throws (my #2, now on `doNotDisturb` and the control verbs).**

`doNotDisturb` is `available` when the device reports DP 107, and its `write` emits `aiotDp(107, …)`. For an `eufy_home_tuya` device that dispatch routes to `TuyaCommandRouter.dispatchCommand` (`eufy-mega.ts:928`), which throws `UNVERIFIED` unless `tuyaAllowUnverified` is set — default `false`. So the surface advertises `doNotDisturb` (and, via the DP 2/101 gates, `startCleaning`/`returnToDock`/`pauseCleaning`) as callable, but calling throws on the happy path. This is the same `available` ≠ callable break that #33 round-2 removed by dropping the Tuya dispatch entirely.

Either fold `tuyaAllowUnverified` into the `available` decision for every Tuya-dispatched write (so they stay hidden until opt-in), or drop the Tuya write path again as #33 did. As-is a host enumerating actions gets a hard error.

src/model/capabilities/locate.ts
```diff
@@ -49,11 +48,14 @@ export const LOCATE_MEMBERS = {
...
-          sink.dispatch(aiotDp(LOCATE_DP, on)),
+      ({ sink, ctx }) => {
+        if (ctx.paramIds.has(LEGACY_LOCATE_DP)) {
+          return (on = true): Promise<void> => sink.dispatch(aiotDp(LEGACY_LOCATE_DP, on));
+        }
+        return (on = true): Promise<void> => sink.dispatch(aiotDp(LOCATE_DP, on));
+      },
       "Trigger the find-robot beep; pass false to cancel one in progress.",
-      (ctx) => isAiotVacuum(ctx),
+      (ctx) => (ctx.paramIds?.has(LEGACY_LOCATE_DP) || ctx.paramIds?.has(LOCATE_DP)) ?? false,
```
max246: Same regression pattern as the mode-control verbs: `locate` now gates on `has(LOCATE_DP=160) || has(LEGACY_LOCATE_DP=103)`. But your own `findRobot` member doc calls DP 160 "a momentary write trigger — the device sends it to begin or cancel a beep but holds no durable state," i.e. it isn't durably reported → 160 rarely/never in `paramIds`, so `locate()` disappears on AIoT. And the DP 103 (Tuya) branch dispatches through `TuyaCommandRouter` → `UNVERIFIED` throw. Restore `isAiotVacuum(ctx)` for the AIoT arm and DP-gate only the legacy Tuya arm.

src/model/param-dictionary.ts
```diff
@@ -3554,4 +3554,49 @@ export const CLEAN_PARAMS: Record<number, ParamDef> = {
...
+    paramType: 119,
+    name: "ClearTotalTime",
```
max246, line 3559: `observed: true` here (and on 120/127/129/134) contradicts this PR's own evidence section — the body says *"`observed: false` … live device reporting has not yet been confirmed"* — and each `note` says "Schema-confirmed," not "Live-confirmed" like the truly-observed X8 DPs above (e.g. DP 113 "Live-confirmed Sweep"). Marking schema-only DPs as `observed: true` pollutes the signal that distinguishes confirmed-live from schema-guessed. Set these to `observed: false` to match the notes and the body. (Also: DP 134 `rssi` isn't listed in the PR body's table at all — add it.)

max246 (commented):
Reviewed PR #34 (stacked on #33). The four read-only sensor DPs (119/120/127/129) + rssi are fine — schema-sourced, read-only, DP-gated. But this PR quietly **re-introduces the two must-fixes that #33 round-2 just removed** (commit `cdd225f`), and does it in a way CI can't catch. Those are blockers. Details inline.

**Must-fix (regressions vs #33)**
1. Mode-control verbs + `locate` are back to gating on write-only command DPs (152/160), which never appear in a real device's reported params — you confirmed this yourself when fixing my #1 on #33. On a real AIoT vacuum these verbs vanish. The new test asserts the opposite by seeding DP 152 into `paramIds`, so green CI masks it.
2. The Tuya branch (DP 2/101) + the new `doNotDisturb` write are `available: true` but dispatch through `TuyaCommandRouter`, which throws `UNVERIFIED` by default — the exact advertised-but-throws break from my #2.

**Should-fix**
3. `observed: true` on the new param-dictionary entries contradicts the PR body ("observed: false") and the notes ("Schema-confirmed", not live).
4. PR body describes only "four read-only DPs" but the diff also does X8→Tuya renames, a locate refactor, new write actions, and a systemic category-guard→DP-guard migration. Please make the description match — or split the write-path change out so it's reviewed as such.

src/model/capabilities/vacuum-clean.ts
martijnpoppen: Fixed in 2795849 — took your second option and dropped the Tuya write path, as #33 round 2 did.

`tuyaAllowUnverified` is a client option (`client/types.ts:140`) consumed when the router is built (`eufy-mega.ts:350`); it never reaches `AvailabilityContext`, so folding it into the gate would mean plumbing a transport setting into `model/`. Dropping the dispatch keeps the decision where the evidence is.

So `startCleaning` / `returnToDock` / `pauseCleaning` are back to `isAiotVacuum` with no Tuya branch, `locate` likewise, and `doNotDisturb` stays read-only. No Tuya clean-line write is offered until one is confirmed on a device.

You were right that the specs were masking it, and it was worse than the DP-152 seeding you spotted: the two "Tuya path" write tests passed `category: undefined`, which makes `isAiotVacuum` **true** — so they never described a Tuya device at all. They now use `category: "eufy_home_tuya"` on a T2266 and assert the verbs are absent, with a companion test that an AIoT device dispatches ModeCtrl and never a bool DP. Same treatment for the locate tests.

While in there I also put back two guards this branch had moved off the reviewed #33 state:

- `power.available` → `isAiotVacuum`. DP 151 belongs to the shared AIoT product schema rather than a device's reported set, and our own dictionary has it `observed: false`, so gating the setter on the reported DP is your #1 failure one member over.
- `battery` and `cleanType` readAliases regain `available: isTuyaVacuum` — the gate had been dropped while `battery`'s JSDoc still claimed it.

Your #4 (`observed: true` on the schema-only entries) was already fixed earlier — 119/120/127/129/134 all read `observed: false` now.

---

src/model/param-dictionary.ts
martijnpoppen: Already fixed before this round — DPs 119, 120, 127, 129 and 134 all read `observed: false` on the current head, matching their `"Schema-confirmed"` notes and the PR body. Nothing further needed here.

`rssi` (DP 134) is still missing from the PR body's table, though; that part of your comment stands.

---

src/model/capabilities/locate.ts
```diff
@@ -33,13 +33,15 @@ export const LOCATE_MEMBERS = {
     kind: "boolean",
     provenance: "mega",
     writtenElsewhere: true,
-    readAliases: [{ paramType: LEGACY_LOCATE_DP, available: isTuyaVacuum }],
+    readAliases: [{ paramType: LEGACY_LOCATE_DP }],
```
max246: The `isTuyaVacuum` gate was dropped from this read-alias (and the import removed above). `locate` binds on codec `vacuum` broadly (`detection.codecs: ['vacuum']`), so DP 103 will now be surfaced as the `locate` property on **any** vacuum that reports param 103 — not just Tuya ones. Practical risk is low (AIoT robots use DP 160 and shouldn't report 103), but the defensive gate is gone. If DP 103 is genuinely Tuya-only, keep `available: isTuyaVacuum` here. Intentional, or collateral from the DP-103 consolidation?

src/model/capabilities/vacuum-clean.ts
```diff
@@ -367,23 +367,31 @@ export function decodeVacuumActivity(raw: ParamValue | undefined, codec: RawDpCo
...
 export type VacuumCleanActions = Surface<typeof VACUUM_CLEAN_MEMBERS>;
 
 /**
  * Every `vacuum_clean` read plus the writes and mode-control verbs.
  *
- * `power` (DP 151) is part of the shared AIoT product DP schema for every T2xxx clean-line device — not
- * a model-specific extension — so its write is gated by `available: isAiotVacuum` only: no equivalent
- * power DP is confirmed on the legacy Tuya clean line (G-series/X8). The three mode-control verbs
- * (`startCleaning`, `returnToDock`, `pauseCleaning`) are DP-gated and dispatch different DP shapes per
+ * `power` (DP 151) write is gated by `paramIds.has(151)` — present on AIoT T2xxx clean-line devices
```
max246: This header block is stale relative to the final gating and will mislead maintainers:

- It says `power` (DP 151) write is gated by `paramIds.has(151)`, but the member uses `available: (ctx) => isAiotVacuum(ctx)` (:415). (The inline `power` doc at :404 is the correct one.)
- It says the mode-control verbs are `paramIds`-gated and that "legacy Tuya dispatches DP 2 / DP 101", but `startCleaning` / `returnToDock` / `pauseCleaning` (:672–700) are each `method(…, isAiotVacuum)` dispatching **only** DP 152 — no `paramIds` gate, no Tuya DP 2/101 dispatch.

Looks like leftover prose from the revision whose Tuya writes were removed in 2795849. Please update it to match the `isAiotVacuum`-only reality — otherwise it reads as though `paramIds`-routed Tuya control writes exist.

martijnpoppen: You're right, and it was exactly that — leftover prose from the revision whose Tuya writes came out in 2795849. Fixed in 14fe61f.

The block now says every write here is AIoT-only on `isAiotVacuum` (`power` DP 151, and the three verbs on DP 152 `ModeCtrlRequest`), that no legacy Tuya write is dispatched at all, and that `doNotDisturb` / `rssi` are DP-gated **reads**. I also added the reason 151 and 152 are gated by platform rather than by reported DP — both belong to the shared AIoT product schema, so a reported-DP gate hides them on real hardware — since that is the thing this file has now got wrong twice.

Also dropped a stale "X8 Pro read members" phrase left from the `X8_*` → `TUYA_*` rename.

---

src/model/capabilities/locate.ts
martijnpoppen: Collateral, not intentional — the gate went missing during the DP-103 consolidation earlier on this branch. Restored in 14fe61f: `readAliases: [{ paramType: LEGACY_LOCATE_DP, available: isTuyaVacuum }]`.

Your reasoning is the deciding one: `locate` binds on codec `"vacuum"` broadly, so without the gate DP 103 surfaces on any vacuum reporting param 103. Practical risk is low — AIoT robots use DP 160 — but this is the same defensive gate `battery` and `cleanType` got back in 2795849, and leaving one of the three ungated is how it drifts again.

---

max246 (approved):

PR #39, opened 2026-08-20
feat(clean): surface language DP 162 as a read for AIoT vacuums

martijnpoppen:
> **Retitled and rescoped.** This was "language DP 162 read+write" and also flipped `volume` (DP 161) to writable. Both writes are held after @max246's evidence review — see *Unverified or deferred*. What remains is one new read.

## What changed

- `VACUUM_DP.LANGUAGE` (162) added to the AIoT clean-line DP constants.
- `language` exposed as an evidence-gated read on `VacuumCleanActions`, gated `isAiotVacuum`.
- `volume` (DP 161) stays a read, as it was before this branch.

## Why

DP 162 is registered in `CLEAN_PARAMS` from `get_product_data_point` and is reported by AIoT clean-line devices, but nothing surfaced it. Landing the read costs nothing and is reversible.

## Evidence

DP 162 is `language` in `CLEAN_PARAMS`, generated from `get_product_data_point` — the same source that grounds every other clean-line DP id in this module. Read direction only; see below for what that source does *not* establish.

## Verified how

- [x] `npm run verify` is green locally (Node 24.5.0) — 121 files, 2392 tests

## Unverified or deferred

**Both writes are held (82c511e).** `setVolume` (DP 161) and `setLanguage` (DP 162) rested on `get_product_data_point` reporting `writable: true` and nothing else — no live `publishDps` capture of either DP, and no confirmation of the value format DP 162 expects.

That bar is weaker than the writes beside them (the DP 152 mode-control frames were live-verified on a T2351), and it fails safe in the wrong direction: `aiot-dp` routes to `TuyaCommandRouter` — which refuses unverified writes by default — only for `eufy_home_tuya` devices. Everything else goes to the MQTT router, which has no such guard. So these two would have reached real robots on schema evidence alone.

Not done here, deliberately: **ratifying schema-`writable: true` as a sufficient bar for AIoT writes.** That is a project-wide precedent for every future AIoT write and wants its own discussion rather than being set as a side effect of this PR.

**DP 162's declared shape is doubtful, beyond the write question.** The member says `String rw`, *"locale code set by the app, e.g. `"en"`, `"zh"`, `"de"`"*, with `kind: "text"`. Three things point away from that:

- The app has a dedicated voice-**package** service — `IVoiceService.getVoicePackage()` → `GET resource/voicePackage`, driven from `BaseMegaDevice.requestVoicePackage()`. Voice on the clean line is a downloadable-package flow, not a locale setting.
- The vendor's `language.proto` matches that flow: `LanguageRequest.Desc { set_id, url, md5, version, size }` + `Selection`, and `LanguageResponse { default_id, current_id, version, set_id, State }` with `State` = `IDLE/UPDATING/SUCCESS/FAILURE` — states that only make sense for a download.
- **The catalog cannot support the "String" claim.** In `CLEAN_PARAMS`, DP 153 (`WorkStatus`), DP 154 (`CleanParam`) and DP 173 (`StationResponse`) are all `type: "string"` too, and all three are Raw protobuf DPs. The generator collapses `data_type: "Raw"` and `"String"` onto one `"string"`, so it does not distinguish a locale from a base64 message.

Not provable offline: the Java bridge sends DPs as an opaque `Map<String, Object>` (`AiotDpManager.sendDpRequest`) and never names DP 162 or its shape, and there is no live capture here.

The read is safe either way — it lands whatever string arrives — so this is not a blocker. But if the wire is protobuf, `kind: "text"` labels a base64 blob as human-readable text, and the "locale code" wording documents a format nobody has observed. Both can be softened before merge on request.


src/model/capabilities/vacuum-clean.ts
```diff
@@ -452,6 +457,22 @@ export const VACUUM_CLEAN_MEMBERS = {
...
+   * Write direction confirmed from `get_product_data_point` (`writable: true`); locale format is
+   * an open string (no live report observed for a closed set of values yet).
+   */
+  language: {
+    param: VACUUM_DP.LANGUAGE,
+    type: "string",
+    kind: "text",
+    provenance: "mega",
+    description: "Device UI language locale code (DP 162, String rw). AIoT clean line.",
+    write: (v) => aiotDp(VACUUM_DP.LANGUAGE, v as string),
```
max246: **Write-evidence bar** — worth a maintainer decision before merge. The doc is candid that this is schema-only: *"Confirmed writable via `get_product_data_point` (`writable: true`); no live publishDps capture yet."* And `aiotDp()` (access.ts) builds the command directly — it does **not** go through `TuyaCommandRouter`'s `allowUnverified` guard (that throws-by-default gate only covers the Tuya `dp.publish` path). So `setLanguage(...)` is `available` on any AIoT vacuum and **will actually dispatch to the device**, uncaptured.

This is the same bar PR #34 was forced to resolve by *removing* unverified writes ("dropping the dispatch keeps the decision where the evidence is"). The existing DP 152 mode-control writes were **live-verified on T2351**; this one is not. Either hold the write until a live publishDps capture confirms DP 162 + the locale format, or explicitly ratify schema-`writable:true` as a sufficient bar for AIoT writes and record it (the way #34 recorded the `observed:false` *read* decision in AGENTS.md). Same applies to `setVolume` below.

src/model/capabilities/vacuum-clean.ts
```diff
@@ -426,16 +428,19 @@ export const VACUUM_CLEAN_MEMBERS = {
...
    */
   volume: {
     param: VACUUM_DP.VOLUME,
     type: "number",
     unit: "%",
     kind: "percent",
     provenance: "mega",
-    description: "Speaker volume 0-100 (DP 161, Value).",
+    description: "Speaker volume 0-100 (DP 161, Value rw). AIoT clean line.",
+    write: (v) => aiotDp(VACUUM_DP.VOLUME, v as number),
```
max246: Two notes on the new `volume` write:

1. **Same schema-only evidence** as `setLanguage` (see the comment there) — no live publishDps capture, and `aiotDp` bypasses the `allowUnverified` guard, so `setVolume(...)` will dispatch to the device on schema evidence alone.
2. **Out of the PR's stated scope** — the title/description is "language DP 162", but this flips `volume` (DP 161) from read-only to read+write. Fine to include, but it should be reviewed *as a write*, not merged as a side-effect of the language change.

Nit: `v as number` with no 0–100 clamp — a caller can send an out-of-range volume.

src/model/capabilities/vacuum-clean.ts
martijnpoppen: Agreed, and held. 82c511e removes `write:`/`writeAs:` from **both** `language` (DP 162) and `volume` (DP 161). The reads stay; the descriptions now say `ro`.

Your point about `aiotDp` is the one that decided it. I checked and it holds: `eufy-mega.ts:928` routes `aiot-dp` to `TuyaCommandRouter` only when `dev.category === "eufy_home_tuya"`, everything else to the MQTT router — which has no `allowUnverified` equivalent. So an unverified **Tuya** write fails safe by throwing, and an unverified **AIoT** write reaches the robot. These two would have dispatched on schema evidence alone, which is a weaker bar than the DP 152 frames beside them that were live-verified on a T2351.

I deliberately did **not** take the other half of your option — ratifying schema-`writable: true` in AGENTS.md. That is a project-wide precedent for every future AIoT write and shouldn't be set as a side effect of a language PR. Worth its own issue if you want it.

**One more thing on DP 162, separate from the evidence bar.** Chasing this I no longer think the member describes the right wire at all:

- The app has a dedicated voice-**package** service — `IVoiceService.getVoicePackage()` → `GET resource/voicePackage`, driven from `BaseMegaDevice.requestVoicePackage()`. Voice on the clean line is a downloadable-package flow, not a locale-string setting.
- The vendor's `language.proto` matches that exactly: `LanguageRequest.Desc { set_id, url, md5, version, size }` plus a `Selection`, and `LanguageResponse { default_id, current_id, version, set_id, State }` where `State` is `IDLE/UPDATING/SUCCESS/FAILURE`. Those states only make sense for a download.
- **The catalog can't support the current claim.** `VACUUM_DP.LANGUAGE`'s doc says *"String rw. Locale code set by the app, e.g. `"en"`, `"zh"`, `"de"`"*, cited to `get_product_data_point`. But in `CLEAN_PARAMS`, DP 153 (`WorkStatus`), DP 154 (`CleanParam`) and DP 173 (`StationResponse`) are **also** `type: "string"` — all three are Raw protobuf DPs. The generator collapses `data_type: "Raw"` and `"String"` onto the same `"string"`, so it does not distinguish a locale from a base64 message.

What I can't do offline is prove it: the Java bridge sends DPs as an opaque `Map<String, Object>` (`AiotDpManager.sendDpRequest`) and never names 162 or its shape, and there's no live DP 162 capture here.

The read is safe either way — it lands whatever string arrives — so this doesn't block the PR. But `kind: "text"` and the "locale code" wording are describing a format nobody has observed, and if it is base64 protobuf then `"text"` hands a caller a blob labelled as human-readable. Happy to soften both to something the evidence supports before merge if you'd like.

---

src/model/capabilities/vacuum-clean.ts
martijnpoppen: All three points taken.

1. **Evidence** — `setVolume` is removed in 82c511e along with `setLanguage`, for the reason in the other thread: `aiotDp` bypasses the `allowUnverified` guard, so this would have dispatched to a real robot on schema evidence alone.
2. **Scope** — you're right that flipping DP 161 from read-only to read+write inside a PR titled "language DP 162" is a write smuggled in as a side effect. With the write gone the scope objection goes with it; the PR is now purely the DP 162 read plus the existing DP 161 read.
3. **The clamp nit** — moot now that the setter is gone, but noting it so it isn't lost: when DP 161 gets a capture and the write comes back, it wants a declared 0–100 domain rather than a bare `v as number`. `memberWrite` already enforces a member's declared domain and rejects out-of-range with one message, so that is where the bound belongs rather than in the `write:` lambda.

Specs follow: the four setter tests are replaced by one asserting no setter is installed on any category, and one asserting both values still read.

---

src/model/capabilities/vacuum-clean.ts
```diff
@@ -452,6 +455,20 @@ export const VACUUM_CLEAN_MEMBERS = {
     readAliases: [{ paramType: LEGACY_VACUUM_DP.BATTERY_LEVEL, available: isTuyaVacuum }],
     description: "Battery level 0-100 (DP 163 AIoT / DP 104 Tuya). NOTE: clean namespace — not param 1101.",
   },
+  /**
+   * Device UI language — the locale the robot uses for its voice prompts (DP 162, String rw).
```
max246, line 459: Nit (doc only): now that the write is held, this JSDoc is stale — it still says **`DP 162, String rw`** and *"Write direction confirmed from `get_product_data_point` (`writable: true`)"*, but the member is read-only (the `description` correctly says `ro`). Worth trimming to match — otherwise a reader thinks a setter exists. The `VOLUME`/`LANGUAGE` DP consts up top carry the same `String rw` wording.

max246 (approved):

PR #46, opened 2026-08-21
feat(auth): distinct sessionExpired event + widen kicked-token classifier

max246:
## Why
When another client logs into the same eufy account, the single active session is transferred and the running client's token is **kicked** (`401 … token … kicked out`). `postSigned` already detects a genuine auth failure — it `clearSession()`s and throws `SessionExpiredError` — but that signal only rides the **generic `error` event**, indistinguishable from a transient error. A host that wants to react to auth loss (re-login, surface a "reauth" state) has to string-match the error bus. (This bit the bridge: it kept reporting healthy while every poll silently 401'd — companion fix in ha-eufy-sdk-bridge#7.)

## Changes
1. **`sessionExpired` event** — emitted alongside `error` whenever a `SessionExpiredError` is reported (`reportError`). A normal event, so emitting with no listener is a harmless no-op (unlike `error`, which throws when unhandled). Matched by name so it survives an error crossing a module boundary. `error` still fires too (back-compat).
2. **Widen the `authFailure` classifier** (`mega-client.ts`) to also match `kicked` / `does not exist`, so a kicked-out 401 is classified as an auth failure (→ `clearSession` + `SessionExpiredError`) even if eufy stops sending code `26084` or rewords the message — instead of slipping through as a *generic* (retryable) error that leaves a dead session in place.

`SessionExpiredError` is **already exported** from the package barrel (`transport/http/index.ts` → `transport/index.ts` → `index.ts`), so no export change was needed.

## Verified
- `npm run build` clean; full suite green (**2322 tests**).
- New specs: `sessionExpired` fires exactly once for a `SessionExpiredError` and not for a plain `Error`; both still reach the `error` bus; emitting with no listener does not throw.


src/transport/http/mega-client.ts
```diff
@@ -441,7 +493,8 @@ export class MegaHttpClient {
     const authFailure =
       authed &&
       last?.status === 401 &&
-      (last.code === 26084 || /user_id is empty|invalid.*token|token.*(expired|error)|unauthor/i.test(last?.msg ?? ""));
+      (last.code === EufyCloudErrorCode.SESSION_KICKED ||
+        /user_id is empty|invalid.*token|token.*(expired|error)|kicked|does not exist|unauthor/i.test(last?.msg ?? ""));
```
lenoxys: `kicked` is fine — unambiguous, and it's the case you set out to catch. `does not exist` isn't.

This tests the detail field, which falls back to a serialised (and where decryptable, decrypted) slice of the response body, not just the vendor's `msg`. So the substring can match almost anywhere in a 401 payload — and "member/device/user does not exist" is generic gateway wording.

The consequence is exactly the hazard the comment directly above this warns about: we clear a healthy persisted session and force the host back through 2FA. It's reachable on the login path too, so a 401 about an account can wipe a live session.

Anchoring it — e.g. `(token|session).*does not exist` — keeps the intent and closes the hole.

Worth noting the asymmetry: the kicked *code* carries a provenance note saying exactly how it's known. These two message patterns don't, and the ungrounded one is the one that can do damage.

src/client/types.ts
```diff
@@ -253,6 +253,13 @@ export type EufyMegaEventMap = {
    * report broker publication (`acked: true`) without an `instanceIp`; that is not device convergence.
    */
   commandAck: [info: { sn: string; kind: string; acked: boolean; instanceIp?: string; getAcked?: boolean }];
+  /**
+   * The cloud session was kicked or invalidated — another client logged into the same account, or the
+   * token expired. The SDK has already cleared the persisted session; a host should re-drive `login()`
+   * (which usually needs 2FA). Distinct from `error` so a host can react to auth loss without
+   * pattern-matching the generic error bus; the same failure ALSO fires `error` for back-compat.
```
lenoxys: We're pre-1.0, with a documented rule against keeping old shapes beside new ones — there's nothing to be back-compatible with yet. Worth picking one emission rather than shipping both and pinning the duplication with a test.

src/client/eufy-mega.ts
```diff
@@ -380,6 +380,10 @@ export class EufyMega extends EventEmitter {
    */
   private reportError(e: unknown): void {
     const err = e instanceof Error ? e : new Error(String(e));
+    // Auth loss (a kicked/expired token) gets its own event so a host can react without string-matching
```
lenoxys: House rule wants this rationale in the JSDoc above the method, not narrated inside the body. Same applies to the expanded captcha comment in the transport.

src/client/eufy-mega.ts
```diff
@@ -380,6 +380,10 @@ export class EufyMega extends EventEmitter {
...
+    // the generic error bus. A normal event, so emitting with no listener is a harmless no-op (unlike
+    // `error`, which throws when unhandled). Matched by name to survive an error crossing a module edge.
+    if (err.name === "SessionExpiredError") this.emit("sessionExpired", err);
```
lenoxys: Two idioms for one type: this matches on `name`, but the same class is already matched with `instanceof` further up this file. The "survives an error crossing a module edge" rationale doesn't really apply inside a single ESM package where `client/` may import `transport/` outright.

Also worth stating out loud: this only fires on the reported-error paths. A session error thrown straight out of a direct call emits nothing — matches what the description promises, just narrower than the stated goal.

src/client/__tests__/connect.spec.ts
```diff
@@ -432,3 +432,29 @@ describe("EufyMega auto-realtime", () => {
...
+    (c.eufy as any).reportError(new Error("transient boom"));
+
+    // Auth loss fires the dedicated event exactly once; a plain error does not.
+    expect(expired).toHaveLength(1);
+    expect(expired[0]?.name).toBe("SessionExpiredError");
+    // Both still reach the generic error bus (back-compat).
+    expect(errors).toHaveLength(2);
+  });
+
+  it("sessionExpired is a no-op (no throw) when nothing listens", () => {
```
lenoxys: This asserts a risk that never existed — the emit two lines up is already guarded by `listenerCount("error")`, so the "unlike `error`, which throws when unhandled" rationale doesn't apply here. Harmless test, misleading premise.

lenoxys (commented):
Event is the right shape — this is the piece a host actually needs. One blocking thing, one missing half, rest are small.

**Blocking:** the widened 401 matcher can clear a healthy session (inline).

**Missing half:** `sessionExpired` exists only in the type map's JSDoc. The troubleshooting and getting-started guides still say "listen for `error`", and the events page doesn't mention it — so a host following the docs lands exactly where this PR is trying to move them away from.

**CI was red on both checks** — `prettier --check`, and the publication guard rejecting the provenance note once `export *` carried the error-code const into the generated reference. Fixed in e0a582d: the const and its type are now `@internal`, so the provenance stays in source rather than being watered down to satisfy a docs grep. Both green now, `verify` included (2322 tests).

**Follow-up:** this lands the *signal*, which is the right first step, but it isn't the re-authentication #59 asks for — a host still can't tell an empty device list from a dead session. Worth linking the two so the remainder doesn't get lost.

max246:
Thanks @lenoxys — all addressed in e4abcda, and thanks for the CI fix in e0a582d.

- **🔴 401 matcher (blocking):** you're right, `does not exist` against the serialised detail is unanchored and could wipe a healthy session (incl. on login). Anchored to `(token|session).*does not exist`; `kicked` stays as-is.
- **Emit one, not both:** a `SessionExpiredError` now emits **only** `sessionExpired`, never `error`. Test updated to assert sessionExpired-only for the session error and error-only for a plain error; dropped the misleading "no-op when unlistened" test.
- **`instanceof`, not `name`:** switched to `instanceof SessionExpiredError` (one idiom; the module-edge rationale doesn't hold inside the package).
- **Rationale in JSDoc:** moved the reportError rationale into its JSDoc, and trimmed the expanded captcha comment in the transport.
- **Missing half (docs):** `sessionExpired` is now documented in getting-started, troubleshooting, and the events page (they previously pointed only at `error`).

`verify` green (2321 tests), `prettier --check` clean.

On the follow-up: agreed this lands the **signal**, not the re-auth #59 asks for — a host still can't distinguish an empty device list from a dead session. This PR is the prerequisite for that; the re-auth/"is the session actually alive" half should continue under #59.

lenoxys (approved):
lgtm

Issue #36, opened 2026-08-19
snapshotLive() intermittently fails decode with non-existing PPS 0 referenced


lenoxys:
## Observation

`snapshotLive()` intermittently rejects with a decode failure instead of returning a still, on cameras
whose live burst otherwise works:

```
ffmpeg JPEG decode failed (code 69): [h264 @ 0x…] non-existing PPS 0 referenced
```

The message means the burst handed to ffmpeg started before the parameter sets, so the decoder had no
SPS/PPS for the first slices.

## Reproduction

Observation-only, against a real fleet through a host that calls `snapshotLive()` with no options:

- Wired indoor camera A: failed with the message above on one attempt, succeeded (41 KB JPEG) on a later
  attempt.
- Wired indoor camera B: failed on the first attempt, then returned 76 KB and 20 KB stills on later
  attempts.

So the same device alternates between success and this failure, which points at burst framing rather
than device capability. Each attempt was a separate cold acquisition; no attempt ran concurrently with
another consumer.

## Expectation

`snapshotLive()` either primes the decoder with the current SPS/PPS and starts collection at a keyframe,
or it rejects with a typed reason a host can distinguish from a permanent failure. Today the host sees an
opaque decode error and cannot tell "retry is worthwhile" from "this camera cannot produce a still",
which matters because a host that rate-limits acquisition spends its budget on an attempt that was
doomed by framing.

## Environment

- SDK: current `main` (`4537414`)
- Node.js 24.19.0, Homebridge 2.4.0, Debian 13 (kernel 6.12)
- ffmpeg: the `ffmpeg-for-homebridge` 2.2.2 build, reachable on `PATH` (see the bare-spawn issue)

lenoxys:
More retest data on the same host, now with a system ffmpeg present (`7.1.5-0+deb13u1`), so nothing here is caused by binary resolution.

One wired indoor camera refused four consecutive cold acquisitions across two runs, then returned a valid 62 KB still on the fifth, with no configuration or network change in between. The other two wired cameras succeeded on their first acquisition in the same runs (41 KB and 192 KB).

That is the shape the report describes: the failure is per-acquisition framing, not per-device capability. It matters for a host that rate-limits acquisition, because four wasted attempts translate into minutes with no image while the SDK reports only an opaque decode error rather than a typed, retryable reason.

PR #49, opened 2026-08-21
feat(ffmpeg): let a host name the ffmpeg binary instead of requiring one on PATH

### LINKED ISSUE #35: Media paths spawn a bare ffmpeg, so hosts without a system ffmpeg cannot use snapshotLive/record
## Observation

`transport/ffmpeg.ts` is the single spawn point for every media path that shells out, and it spawns the
binary by name:

```js
const ff = spawn("ffmpeg", ["-hide_banner", "-loglevel", ffmpegLogLevel(opts.level), ...args], { … })
```

There is no option to supply the binary. `EufyMegaOptions` exposes only `ffmpegLogLevel`.

## Impact

A Homebridge host installed through `hb-service` has no `ffmpeg` on `PATH`; `/opt/homebridge/bin`
contains only `corepack`, `hb-service`, `node`, `npm`, `npx`. Homebridge plugins are expected to use the
bundled `ffmpeg-for-homebridge` binary instead, and they own its path.

On such a host every SDK path that shells out fails immediately. Observed while qualifying a HomeKit
snapshot request against a real fleet, with Homebridge 2.4.0 and Node 24.19.0:

```
ffmpeg not runnable (is it installed?): spawn ffmpeg ENOENT
```

`snapshotLive()` therefore never returns an image, and a host cannot repair it without mutating the
process `PATH` globally, which is not acceptable for a plugin that already knows the exact binary it
ships. Prepending the bundled directory to `PATH` proved the rest of the path works, so the only gap is
binary resolution.

## Requested contract

Accept a host-provided executable, for example `new EufyMega({ ffmpegPath })`, and use it in the shared
spawn helper. Falling back to `"ffmpeg"` when the option is absent keeps current behavior. A per-call
override is not required; one client-level option is enough for the host to pass the binary it already
validates.

## Environment

- Cameras: wired indoor models, live burst reached the decode stage once `PATH` contained a binary

### LINKED ISSUE #1003: Deliver Refresh snapshots and persistent last successful images
## Parent

#981

## What to build

Serve Refresh requests responsively while preserving the last successful validated source image across restart and backup.

## Acceptance criteria

- [ ] Refresh immediately serves the last successful image, may try stored-only acquisition when empty, and schedules at most one live refresh every two minutes without permanent polling.
- [ ] Validated bounded JPEGs are atomically stored under opaque serial-derived names with owner-only access and survive restart and full Homebridge backup.

## Blocked by

- #1002

lenoxys:
Closes #35.

## What

`transport/ffmpeg.ts` was already the single spawn point, but it spawned the binary by NAME. `new EufyMega({ ffmpegPath })` now names the executable, threaded to the spawn through the same route `ffmpegLogLevel` already takes (client options → router deps → the media opts bag). Absent, it defaults to the bare name, so the `PATH` lookup and today's behaviour are unchanged.

## Why

Every media path that shells out — JPEG `snapshotLive`, one-shot `record`, the WebRTC container mux — was unavailable on a host with no `ffmpeg` on `PATH`. That is an ordinary state for a managed install, and the only workarounds were installing a second unmanaged ffmpeg or mutating `process.env.PATH` process-wide to reach a file the caller already holds the absolute path to. Neither is a contract a host can offer its users.

## Decisions worth a reviewer's attention

**The availability probe had to move, and that changes the public surface.** `hasFfmpeg()` / `hasFfprobe()` lived in `webrtc/peer.ts` with their own hardcoded `spawnSync("ffmpeg", …)`, contradicting `ffmpeg.ts`'s claim to be the one place that shells out. The WebRTC container gate reads that answer, so a host-supplied build reported "missing" and the mux **silently degraded to a raw elementary stream** — a caller asking for `.mp4` got something else. They are now `ffmpegAvailable(executable)` / `ffprobeAvailable()` in `ffmpeg.ts`, resolving through the same helper the spawn uses, so probe and spawn cannot disagree. Keeping the old names beside the new ones would be the "old shape beside new" that `AGENTS.md` forbids below 1.0, so they are removed rather than aliased. **Flagging it explicitly** — the path-awareness is required by the issue, the rename is a judgement call for the module's `ffmpeg*` naming, and it is the one breaking bit here.

**`WebRTCPeerOptions.ffmpegPath` exists even though the issue says "a per-call override is not required".** The SDK never constructs a peer — the caller does — so no client-level option could reach it, and without this the container mux stays broken on exactly the host in the report.

**`ffprobeAvailable()` deliberately takes no executable.** No SDK path spawns ffprobe and no option configures it, so a parameter there would be unreachable surface.

**A blank value counts as absent; a non-blank one is never rewritten.** An unset host config arrives as `""`, and spawning that fails as an ENOENT on the empty string — the misleading message this option exists to remove. But a leading/trailing space is legal in a POSIX path, so nothing non-blank is trimmed.

**Nothing is probed at construction.** A wrong path surfaces as the media call's own "not runnable" rejection, the same signal a missing `PATH` entry gives. `ffmpegAvailable(path)` is there for a caller that wants to ask first.

## Tests

- `transport/__tests__/ffmpeg-binary.spec.ts` — resolution precedence and the probe, with `node:child_process` mocked so no ffmpeg is needed (the point of the option is hosts that have none).
- `transport/webrtc/__tests__/ffmpeg-container-binary.spec.ts` — the mux gate probes the caller's executable, muxes with the same one, and still falls back to raw when it is not runnable.
- `transport/p2p/__tests__/ffmpeg-path-plumbing.spec.ts` — both P2P media legs receive it, and `new EufyMega({ ffmpegPath })` actually hands it to the router.

`npm run verify` green: 123 files, 2337 tests.

## Not in scope

Issue #36 (intermittent `non-existing PPS 0` decode failures) is a separate branch. The retest comment on #35 already established the bare spawn is not its cause.


lenoxys:
## Live-fleet verification

Run against a real fleet of eight cameras on a host with a system ffmpeg at `/usr/bin/ffmpeg` (7.1.5) **and** a bundled `ffmpeg-for-homebridge` binary, so both resolution paths are exercisable. Read-only: no device writes.

**The probe distinguishes all three cases, and agrees with the spawn:**

| `ffmpegAvailable(...)` | `PATH` intact | `PATH` scrubbed |
| --- | --- | --- |
| bare name (no argument) | `true` | **`false`** |
| bundled absolute path | `true` | **`true`** |
| bogus path | `false` | — |

**And the option is what makes the media path work under the issue's exact host condition** — `PATH` emptied in-process, so the bare name cannot resolve:

| variant | result |
| --- | --- |
| no `ffmpegPath`, `PATH` scrubbed | fails with `ffmpeg not runnable (is it installed?): spawn ffmpeg ENOENT` |
| `ffmpegPath` = bundled binary, `PATH` scrubbed | **a real 91657-byte 2560x1440 JPEG in 4.6 s** |

The failure line is the one quoted in the report, reproduced deliberately; the row below it is the same call on the same camera with nothing changed but the option. That is the whole contract, end to end.

Note the first row is also the honest answer to "should the SDK probe the path at construction": it doesn't, and the failure a wrong path produces is exactly the failure a missing `PATH` entry produces — which is why `ffmpegAvailable(path)` exists for a caller that wants to ask first.

lenoxys:
**Bottom of a stack:** #50 → #53 → #54 build on this branch, in that order. The tip of #54 is a single SDK build carrying all four, for integration testing against a host.

lenoxys:

Issue #38, opened 2026-08-19
A camera stops delivering live frames for every later session until the client restarts


lenoxys:
## Observation

On a wired camera (`T8410`, P2P, model recorded without serial), `camera().live()` resolves normally but then emits no `video` frame at all. The handle stays open, no `error` or `stop` event fires, and the consumer's frame handler is never called, so a host cannot distinguish this from a slow source except by its own deadline.

## Characterisation

Measured with a HAP controller driving a Homebridge host against this SDK, one session at a time:

1. The same camera streamed successfully four times in one client lifetime (first frame 12.5-12.8 s each time, 30 fps sustained).
2. It then produced zero frames on three consecutive sessions about one to fifteen minutes apart. The host bounded each session on its own 10 s deadline.
3. A second camera in the same client streamed normally between two of those failures, so the client, account session, and push/MQTT path were healthy. The state is per device.
4. Restarting the client cleared it immediately: the same camera delivered a first frame after 12.6 s and 31 fps on the next attempt.

Every session was fully closed by the host before the next one started, and no session was left abandoned.

## Why this is an SDK gap

A host cannot recover from this. It can only fail the request after a timeout, and it has no typed signal to act on: there is no `error`, no budget notice, no stop, and no observation that says this source is dead. Either the shared live source should surface a typed failure when the device stops delivering after a successful start, or the acquisition should fail fast rather than resolve a handle that never produces a frame.

## Environment

SDK `0.1.0-beta.17` equivalent working copy, Node 24.19.0, wired camera over local P2P, no cloud fallback involved. No device serial, address, or account identifier is included.

lenoxys:
## Correction: the SDK does surface this, the host cancels it

I filed this claiming no typed failure exists. That is wrong, and I want the record straight before anyone works on it.

`SharedLiveSource.warm()` arms a 20 s warm-up deadline with 2 s nudge retries, and `onWarmTimeout()` warns `live stream failed to start (no frames within warm-up window)` and calls `fail(err)` on every consumer, which reaches the host as an `error` event. That is exactly the signal I said was missing.

The reason my host never saw it is an ordering problem on the host side, not a gap here:

| t | Actor | Effect |
|---|---|---|
| 0 s | host starts a session | `attach()` → `warm()`, warm deadline armed for 20 s, nudges every 2 s |
| 10 s | host's own video-start deadline | host stops its consumer |
| 10 s | `onDetach` | last consumer left, 8 s linger armed |
[...]

PR #50, opened 2026-08-21
fix(media): prime the decoder with the stream's parameter sets, and type why a still failed

### LINKED ISSUE #36: snapshotLive() intermittently fails decode with non-existing PPS 0 referenced

### LINKED ISSUE #1003: Deliver Refresh snapshots and persistent last successful images

### LINKED ISSUE #1004: Enforce snapshot presentation and cache lifecycle
## Parent

#981

## What to build

Apply truthful snapshot presentation precedence and all specified image-cache validation, provenance, warm-stream capture, and deletion rules.

## Acceptance criteria

- [ ] Presentation order is disabled placeholder, last successful real image, typed-offline placeholder when no image exists, then unavailable placeholder; silence and timestamps never imply offline.
- [ ] Contracts cover corrupt and oversized files, restart provenance reset, one fallback capture from an already-warm successful stream, privacy withholding, and every explicit deletion trigger.

## Blocked by

- #998
- #1003
- External gate: A published SDK build containing the accepted typed device-availability contract is required for offline presentation.

### LINKED ISSUE #1042: A refused snapshot cannot be attributed, so an intermittently failing camera looks broken
## Parent

#981

## What happens

A HomeKit snapshot request that the plugin cannot answer produces no record anywhere. The controller gets
an HTTP error for the `/resource` request and the Homebridge log, the plugin JSONL, and the diagnostics
archive all stay silent, so there is nothing to attribute the failure to.

Measured on a wired fleet through `scripts/live-hap-snapshot-check.mjs`, two rounds per camera 40 s apart:

| camera | round 1 | round 2 | later run |
| --- | --- | --- | --- |
| A | refused | served 143 KiB JPEG | served |
| B | refused | served 43 KiB JPEG | served |
| C | refused | refused | **served 60 KiB JPEG** |

Camera C fails **intermittently**: two runs, with 20 s and 40 s between rounds, produced nothing at all with
no live session involved, and a later run with a longer settle served a real image. A and B behave exactly as
the `Refresh` policy intends throughout. Which acquisition failed on C — the stored one, the live one, or
both — and how often, cannot be told from the outside, and the plugin does not say. An intermittent failure is
the case that most needs attribution, because there is nothing to reproduce on demand.

Possibly the same underlying intermittency as
[mega-yfue/eufy-sdk#36](https://github.com/mega-yfue/eufy-sdk/issues/36), which cannot be confirmed from this
side without the records this ticket asks for.

## Why the records are missing

Two separate holes:

1. **A failed acquisition is not a condition at all.** `LiveCameraDelegate.handleSnapshotRequest` passes the
   error straight to the callback and reports nothing, so `hap-nodejs` answers the resource request and the
   reason is lost.
2. **The two snapshot and streaming capability conditions that DO exist are dropped.** The camera bundle
   emits `camera-snapshot-capability-unavailable` and `camera-streaming-capability-unavailable`, but neither
   code appears in `HOMEKIT_CONDITIONS` in `src/diagnostics.ts`, and `reportHomeKit` returns early for a
   code it does not recognise. So a camera that loses snapshot or live-media evidence entirely also produces
   no human log line. The `member` allowlist has the same shape of gap: `snapshotStored` and `snapshotLive`
   are not in `MEMBERS`.

This is the snapshot-side twin of #1032, which fixed exactly this for live sessions.

## What to build

- Report one bounded reason when a snapshot request cannot be answered, distinguishing at least a missing
  stored acquisition, a failed live acquisition, and no retained image, and clear it when a later request
  succeeds. The existing live-session latch is the shape to copy.
- Allowlist the two capability-unavailable codes and the two snapshot members so an evidence withdrawal is
  visible in the log it is supposed to appear in.
- Do not change acquisition policy here; #1004 owns presentation, and its placeholder steps have since
  landed, so a camera like C now shows an explicit "unavailable" image and one bounded
  `camera-snapshot-unavailable` condition that clears when a real image arrives. What is still missing is
  *which* acquisition failed, and a count of how often — an intermittent camera looks identical to a broken
  one in the log.

## Acceptance criteria

- [x] A refused snapshot request records exactly one bounded, allowlisted reason with no device material,
      distinguishing the failing acquisition rather than only reporting that none answered.
- [x] The reason clears when a later request for the same camera succeeds (landed with the placeholder in
      `c3ec032`, proven on the wire: `active: true, reason: no-acquisition` then `active: false, reason:
      recovered`).
- [x] `camera-snapshot-capability-unavailable` and `camera-streaming-capability-unavailable` reach the log.
- [x] A live run can state which acquisition failed on a camera that serves a snapshot only intermittently.

## Blocked by

Nothing. The two code paths that swallow a reason are named above, both are in this repository, and the
condition allowlist is a table in `src/diagnostics.ts`.

## Relationships

- **Does not block #1004.** #1004's remaining scope is the typed-offline placeholder and the cache lifecycle
  rules, none of which need attribution. Its placeholder steps have already landed (`c3ec032`, `c94cbab`), so
  an unattributable failure is now presented truthfully even though it is not explained.
- **Sibling of #1032**, which fixed exactly this shape of silence for live sessions. That ticket is the
  precedent for the reason vocabulary and the latch-and-clear behaviour this one should copy.
- **Related to [mega-yfue/eufy-sdk#36](https://github.com/mega-yfue/eufy-sdk/issues/36)**, an intermittent
  `snapshotLive()` decode failure that is a plausible cause of the measurement above. Confirming or excluding
  it from this side is not possible until the records this ticket asks for exist, which is why the SDK issue
  is related rather than blocking.

lenoxys:
Closes #36.

## Root cause

Framing, not capability. A camera commonly announces SPS/PPS **once**, with the first keyframe of a stream. `snapshotLive` skips that first (often partial) IDR by default, and a consumer joining a warm source never sees it at all — so the burst handed to the decoder routinely begins *after* the only announcement and has no parameter sets for its first slices. Whether a given attempt lands before or after one is luck, which is exactly the reported alternation between a still and a failure on the same camera.

Only the shared source sees every frame from stream start, so only it can answer what the current parameter sets are; a consumer cannot recover them from what it was given.

## What changed

- `SharedLiveSource.parameterSets` retains the sets last announced, cleared on teardown so a rebuilt stream never primes with a dead stream's sets.
- `prefixParamSets` / `updatedParamSets` in `annexb.ts`, next to the NAL scan that already owned `extractParamSets`. `updatedParamSets` folds **per kind**, because a decoder retains the last set it was given of each — a unit re-stating only its SPS says nothing about a PPS still in force.
- `snapshotLive` primes its burst; `record` does too, because its clip starts at the second keyframe and had the identical defect. That also settles the codec label: the sniff looks for a config NAL, so a run of bare slices was falling back to H.264 and would have mislabelled an H.265 clip.
- `LiveSnapshotUnavailableError` with `reason` (`no-keyframe`, `source-failed`, `undecodable-burst`, `decoder-unavailable`) and a derived `retryable`.

Priming alone could not satisfy the issue, because a burst can fail for reasons no priming fixes and the caller could not tell them apart. `retryable` is derived from `reason` rather than passed in, so the two cannot disagree and callers don't each re-derive the mapping.

## Live-fleet verification

Eight cameras, same script, read-only, baseline vs branch:

| | `non-existing PPS 0 referenced` | failures typed |
| --- | --- | --- |
| `beta-0.1.0` | **4 of 16 attempts** (two camera models) | none — all bare `Error` |
| this branch | **0 of 40 attempts** | all of them |

Two further findings came out of that run and are fixed here:

**`-pix_fmt yuvj420p` is pinned on the JPEG output.** One camera failed with `Non full-range YUV is non-standard, set strict_std_compliance to at most unofficial` — the mjpeg encoder refusing a limited-range input, which is what camera streams signal (`yuv420p(tv, bt709)`). Whether it sees one depends on which pixel format negotiation settles on, so the same camera decoded on one attempt and failed on the next. Against a captured real burst the flag produces **byte-identical** output where negotiation already chose that format, so it constrains only the failing case. The camera that never produced a still now does.

**A burst can yield no image while ffmpeg exits 0** — asked for one frame it finds no complete frame and reports success. That was reported as `ffmpeg JPEG decode failed (code 0)`, blaming ffmpeg for succeeding. Now described as a property of the burst, under the same reason.

### Honest reading of the remaining failures

Success count is the same on both branches (this fleet is flaky — see #38, and two cameras hit the warm-up timeout on both). What changed is *which* failures happen and what a caller is told:

- The reported PPS-0 class is gone.
- The two warm-timeout cases now arrive as `source-failed` **at the moment the source gives up**, instead of `no-keyframe` after burning the full 20 s window on a source that had already failed.
- One model (`T8114`) still fails, now as `undecodable-burst` / "no complete frame in the burst", with ffmpeg reporting `bytestream -28` truncation. It failed on the baseline too (3 of 4 attempts), where the missing parameter sets masked it. **That is a distinct pre-existing defect this PR does not fix** — priming removed the error that was hiding it. I'll file it separately rather than widen this change.

One more observation for a follow-up, not touched here: on the camera whose burst switches resolution mid-stream (`Reinit context to 1280x720` then `1920x1088`), the returned `width`/`height` come from the frame header and disagreed with the JPEG's own SOF (reported 2304x1296, image 1280x720). Every other camera matched.

## Tests

Specs pin the **bytes handed to the decoder**, not just the resolved image — asserting only the image would pass with the priming dropped, because a burst that happens to carry its own sets decodes either way. Covered: stream-order emission for H.264 and H.265, the per-kind fold, the SPS-only burst, a codec mismatch never substituting, the argv pixel format, and each typed reason with its `retryable`.

`npm run verify` green: 124 files, 2349 tests.

## Not in scope

#35 (host-provided ffmpeg binary) is PR #49. The retest comment on #36 already established the bare spawn is not this issue's cause, and this branch is cut from `beta-0.1.0` so the two are independent.


lenoxys:
**Stacked:** base is now `feat/host-provided-ffmpeg-binary` (#49), so this diff shows only the priming change. The two overlap in `media.ts` — `annexbToJpeg` takes #49's spawn-options bag and this PR's primed burst — resolved in the rebase.

Stack order: #49 → **#50** → #53 → #54. The tip of #54 is a single SDK build carrying all four.

lenoxys:

PR #53, opened 2026-08-22
fix(p2p): recover a camera whose live start failed, instead of latching it dead until restart

### LINKED ISSUE #36: snapshotLive() intermittently fails decode with non-existing PPS 0 referenced

### LINKED ISSUE #38: A camera stops delivering live frames for every later session until the client restarts

### LINKED ISSUE #1032: A live session that never produces video leaves no diagnostic record
## Observation

Automating the log step of #1029 showed that a complete negotiated live session writes nothing to either log. During a 20 second session with a mid-session reconfiguration and a concurrent second session, the Homebridge service log received zero lines and the plugin JSONL received only SDK debug records, no plugin condition.

That is fine while streaming works. It is a problem when it does not: when the SDK live source delivered no frame at all, the plugin correctly bounded the session on its video-start deadline, force-stopped the HomeKit session, and returned the accessory to an available streaming status, and still recorded nothing. From the logs the failure is invisible, so a user reporting "live view spins and gives up" gives a maintainer no evidence at all.

## What is missing

`FfmpegLiveMedia` fails a session closed through `onVideoFailure` for source acquisition timeout, missing keyframe inside the start deadline, RTCP silence, and adaptation process failure, but none of those paths reaches `diagnostics.ts`. The media domain has no diagnostic seam at all; the only camera-streaming conditions are emitted at adapter attachment.

## Suggested shape

An allowlisted media condition with a bounded reason vocabulary matching the existing failure paths (`source-acquisition-timeout`, `no-keyframe`, `rtcp-timeout`, `adaptation-failed`), plus one record for a normal start and stop so #1029's "one start/stop lifecycle" check has something to read. Reasons only, with the aliased accessory identity already used elsewhere and no media, key, or address material.

## Evidence

Observed with `scripts/live-hap-stream-check.mjs --instance-log --jsonl` against a dedicated unpaired instance on 2026-08-20. Session lifetime policy is #998's subject, so this may belong there rather than as its own change.

lenoxys:
Closes #38.

Addresses the surviving requirement from the issue's own correction — **the non-recovery**, not the missing signal (which the SDK did already emit).

## Why it latched

Three things combined, and all three had to go:

1. A failed start fails its consumers but does **not** detach them, leaving the source `stopped` with a non-zero consumer count.
2. The router only replaced a stopped source when that count had reached zero — so a caller holding its handle got the **same** source back on every later acquisition.
3. That source's stream factory closes over the session resolved when it was built, so every attach rebuilt a stream over the same cached session. When it is the session (or the per-device state on it) that has stopped serving the camera, rebuilding in place reproduces the dead start forever. A client restart worked because it was the only thing that built a new session.

## What changed

- `SharedLiveSource` reports a stream torn down having **never delivered a frame**, via a new `onStartFailed`.
- The router drops the dead source on that report and **recycles the device's P2P session**.
- A **stopped** source is now always replaced, whatever is still attached to it.
- `dispose()` releases the session user when consumers were still attached (`onActive` fired on 0→1; this is the 1→0).

### The condition is "no frame ever arrived", not "the deadline fired"

My first attempt reported only from `onWarmTimeout`, which **missed the case the issue actually records**: *"each session was fully detached before the next attach"* — the caller's departure tears the source down and `clearWarmWatch()` cancels the deadline before it fires. Enumerating the ways a start can fail is how the one that happens gets left out; an upstream error or stop arriving first, and the battery budget stopping the pull, were missing too. One rule in `teardown` now covers every path. `dispose()` is excluded — the owner asked for that one, and disposing is what an owner does *in response to* this callback.

### A latent re-entrancy this exposed

`teardown` called `stream.stop()` while still holding the reference, and `stop()` emits `"stop"` synchronously, which the source listens for — so **teardown re-entered itself** via `onUpstreamEnd`, whose `!this.stream` guard could not help. It reported twice and closed the session twice. The reference is now dropped before stopping. The double teardown predates this PR; it was invisible until something acted on it.

## Only standalone sessions are recycled

An attached camera shares its HomeBase session with every other camera on it, and closing that to recover one would drop the rest — so an attached camera gets the stream rebuild and nothing more. Derived through `stationKeyOf` so this and `resetStandaloneSession` cannot disagree about what standalone means.

Unlike `resetStandaloneSession` this does **not** wait for the station to fall idle: the failed source's own session user is still counted, so a deferred reset would never fire. It therefore hard-closes, which can interrupt other in-flight work on that station — for a standalone device that work is on the same camera whose stream just died, so I judged that acceptable. **Worth a second opinion.**

## Live verification

One client for the whole run — a restart anywhere would destroy the evidence — with the host deadline set *longer* than the SDK's warm window so the SDK's own failure is what is observed. Serials redacted.

```
attempt 4  warming → "failed to start (no frames within warm-up window) (20000ms, consumers=1)" → fail
           → "[session <device>] connecting now (battery, on demand)" → "connected"
attempt 5  warming → first frame in 2.5 s → ok
```

`connecting now (on demand)` is logged **only** when the manager holds no cached session, so the recycle is directly observed rather than inferred from the outcome. `consumers=1` shows the consumer was still attached at the failure — the state that used to block replacement.

| | sequence over consecutive sessions | reconnect between attempts |
| --- | --- | --- |
| `beta-0.1.0` | `ok, fail, fail, fail` — latched, exactly as reported | none |
| this branch | `ok, fail, ok, fail` and `ok, ok, ok, fail, ok, ok` | yes, after each failure |

A full-fleet run of eight cameras, three sessions each, was 24/24 — including both models that failed repeatedly on the baseline — and no camera was disrupted by another's recycle.

## Caveats

- **The attached-camera path is not covered live.** Every camera in this fleet is its own station, so the "never close a shared HomeBase" behaviour is proven by unit test only.
- **This fleet is flaky**, which is why the evidence is the *reconnect trace* rather than a success-rate delta: which attempts fail varies run to run, but a failure is now always followed by a rebuilt session.
- A caller holding a replaced source now gets a throw from `attach()` (it has been disposed) and must re-acquire through `camera.live()`. Documented on `sharedLiveSourceFor`.

`npm run verify` green: 122 files, 2340 tests.


lenoxys:
**Stacked:** base is now `fix/snapshot-live-decoder-priming` (#50), so this diff shows only the recovery change. Both touch `shared-live-source.ts` — #50 adds the parameter-set cache that `teardown` clears, this PR rewrites `teardown` to drop the stream reference first; verified merged (`lastParamSets = undefined` survives inside the new body).

Stack order: #49 → #50 → **#53** → #54. The tip of #54 is a single SDK build carrying all four.

lenoxys:

PR #54, opened 2026-08-22
fix(client): let a long-lived client observe enablement, and state what a device cannot report

### LINKED ISSUE #47: A long-lived client never observes camera.enabled changing
## What happens

A long-lived client's read-through freshness refresh never updates `camera.enabled`, so a host that keeps
one client for the life of its process can never observe a camera being turned off or on.

Measured on a wired camera (`T8400`, standalone, reports enablement under the `2001` read alias), with a
Homebridge plugin holding one `EufyMega` for the whole run:

| step | observation |
| --- | --- |
| `camera.setEnabled(false)` from a second, freshly constructed client | acknowledged in 1.2 s |
| read from a **newly constructed** client ~30 s later | `enabled === false` |
| read from the **long-lived** client, repeatedly, over 4 minutes | `enabled === true` |
| read from the long-lived client's host after a **restart** | `enabled === false` immediately |
| `camera.setEnabled(true)`, then a fresh client ~30 s later | `enabled === true` |

The reads were spread over minutes and there were many of them, each of which should have found the cached
entry stale (`staleAfterMs` is the default 15 s) and scheduled the coalesced background refresh. The value
never moved. A restart of the same host reading the same account resolves the correct value at once, so the
cloud has it and discovery reads it; only the refresh path on an established client does not.

## Why it matters

An observation that cannot change during a session cannot be acted on during a session. A HomeKit host
uses `camera.enabled` to refuse live view for a camera that is off and to end a session when it is switched
off mid-view. The refusal works, because it reads the value as discovery resolved it. The mid-session half
cannot work at all today: the host re-reads the observation (there is no event to subscribe to — see
below), and the read never changes, so a viewer keeps watching a stream from a camera that is off.

## Suspected mechanism

Not diagnosed from inside the SDK, only bounded from outside. Two candidates, both consistent with the
measurements:

- `Device.getProperty` schedules the refresh only when `Date.now() - value.ts > staleAfterMs`. If `ts` is
  being renewed by unrelated realtime params applied to the same device, the enablement entry may never be
  considered stale even though its value is months old.
- `DeviceRegistry.record()` overlays a live `get_device_param_list` on the device-list params and swallows
  any failure of that call. A refresh whose live overlay keeps failing silently returns the startup params
  forever, which would look exactly like this.

The second is easy to confirm with a log line; today a failed overlay is indistinguishable from a
successful one that returned the same values.

## Related gap in the same area

There is also no event for this transition, so re-reading is the only mechanism available to a consumer:
`CAMERA` declares no `events` and no `emits`; `Device.applyParams` computes the list of changed property
names and every caller discards it; `deviceState` carries `{ sn, stationSn, lastSeenMs }` and no values.
The shape that would solve it already exists in the SDK — `arming.mode` declares an `observation` policy
with a bounded readback and a `armingModeChanged` emit. An equivalent for camera enablement would let a
consumer act on the transition instead of polling for it.

Separately, privacy mode is unreadable by construction (`writeOnly`, "nothing reports it back"), and on the
families whose power rides the `6250` privacy envelope that wire is not aliased into `enabled`, so a camera
in privacy mode is not observable at all. That is a design consequence rather than a bug, but it is the
reason enablement is the only observation a consumer can gate on.

## Not this issue

The write itself is fine: `setEnabled` took effect both directions on this device and was visible to a
fresh client within ~30 s each time.

### LINKED ISSUE #48: Privacy mode is unobservable, including where it is the camera's power state
## What is missing

`camera.privacy` is a setter with no getter, so a consumer can put a camera into privacy mode but can never
observe that it is in one. `CAMERA_MEMBERS.privacy` is declared `writeOnly` with no `param` ("Nothing
reports it back, so it is a setter with no getter"), and `members.ts` drops write-only members from the
property schema by design.

For most families that is a gap on a secondary feature. For three families it is a gap on the primary
on/off state, because their power rides the privacy envelope: `usesSeparatePowerEnvelope` routes
`setEnabled` for indoor-cam mini, the S350 indoor pan/tilt family, and outdoor pan/tilt to the `6250`
privacy burst, inverted. The `enabled` member documents the consequence itself — "The S350/outdoor-PT
privacy form (6250) is a separate wire and is not aliased here until its polarity is captured" — so on those
devices `setEnabled(false)` succeeds and `enabled` does not move.

## Why it matters to a consumer

A HomeKit host has to decide whether to offer live view for a camera. It can only ask `enabled`, so:

- a camera in privacy mode is indistinguishable from a camera that is recording, on every family;
- on the three families above, a camera that has been turned off is also indistinguishable from one that is
  on, because the observation those devices would move is the one that is not readable.

The host cannot work around this without guessing, and guessing what a camera is doing is exactly what it
must not do: refusing live view on a wrong guess withdraws a working camera, and allowing it on a wrong
guess shows a viewer a stream that will never carry frames.

## What would resolve it

Either of:

1. A readable privacy observation, if the `6250` read polarity can be captured — which the `enabled`
   member's own note implies is the blocker ("not aliased here until its polarity is captured").
2. An explicit, typed statement that privacy is unobservable for a given device, so a consumer can
   distinguish "observed as not in privacy mode" from "cannot be observed" instead of inferring it from the
   absence of a getter.

Related: [#47](https://github.com/mega-yfue/eufy-sdk/issues/47) covers the separate problem that even the
readable enablement observation does not update on a long-lived client.

### LINKED ISSUE #1043: Qualify mid-session live termination for a camera switched off while streaming
## Parent

#981

## What to verify

Mid-session termination of a live view for a camera that is switched off while it streams is implemented and
covered hermetically, but has never been observed on a real device. This ticket is the live qualification, and
it cannot start until its external gate clears.

The behaviour under test, decided in #1031 and shipped in `fd07d25` under #998: while at least one live session
is active, the plugin re-reads the camera's admitted enablement observation on a bounded tick, and when that
observation reads disabled it tells HomeKit the session ended, stops the adaptation, and stops the SDK
consumer, instead of letting a viewer watch frames from a camera that is off.

## Acceptance criteria

- [ ] A camera streaming a negotiated session, switched off through a typed SDK write, has its session ended by
      the plugin, with the delay from the acknowledged write to the ended session measured and recorded.
- [ ] No adaptation process survives that termination and the stream management service returns to available.
- [ ] The plugin records `camera-live-session-refused` with reason `disabled-mid-session`, and records **no**
      `camera-live-session-failed`, which is what distinguishes the gate acting from the source dying.
- [ ] A new session is refused with HAP `ERROR` while the camera stays off, and admitted again once it is on.
- [ ] The camera is restored to the state it started in, and the run performs no other device write.

`scripts/live-hap-disabled-camera-check.mjs` already drives every one of these in one pairing and restores the
camera in a `finally` block, so this is a single command once the gate clears:

```bash
node scripts/live-hap-disabled-camera-check.mjs \
  --device-id AA:BB:CC:DD:EE:FF --address 127.0.0.1 --port 51955 --pin 000-00-000 \
  --serial T8XXXXXXXXXXXXXX --eufy-storage /tmp/hb-check/homebridge-eufy \
  --homebridge-pid <homebridge pid> --instance-log /tmp/hb-check/instance.log
```

## Why it cannot be verified yet

The observation the plugin re-reads does not move inside a running plugin. Measured on a wired camera, plugin
at `fd07d25`, with the camera switched off through the typed SDK write:

| reader | observation |
| --- | --- |
| a newly constructed SDK client, ~30 s after the write | `enabled === false` |
| the running plugin, repeatedly, over 4 minutes | `enabled === true` |
| the same plugin's own HomeKit "Camera Enabled" switch, read twice, minutes apart | `1` (on) |
| the same plugin after a restart | `enabled === false` immediately |

So the cloud has the new value and discovery reads it; only the read-through refresh on an established client
does not. The plugin's supervision tick therefore has nothing to act on, and the live script reports a timeout
on this step by design rather than by defect. The refusal half of the same decision is unaffected, because it
reads the observation as discovery resolved it, and is already proven on the wire on two cameras (see #998).

## Blocked by

- External gate: [mega-yfue/eufy-sdk#47](https://github.com/mega-yfue/eufy-sdk/issues/47) — a published SDK
  build in which a long-lived client's read-through refresh updates `camera.enabled`. Without it there is no
  transition for the plugin to observe, so no amount of plugin work can make this qualification pass.

## Relationships

- **Implements nothing new.** The code landed in `fd07d25` (#998) with a hermetic contract that drives the tick
  under fake timers and is mutation-checked, so this ticket adds live evidence rather than behaviour. If the
  gate clears and the qualification still fails, this ticket becomes the bug report for that path.
- **Decision:** #1031 chose this behaviour over letting a disabled camera keep streaming, and #998 absorbed it.
- **Sibling gap:** [mega-yfue/eufy-sdk#48](https://github.com/mega-yfue/eufy-sdk/issues/48) — privacy mode is
  unobservable, and on the families whose power rides the privacy envelope a switched-off camera may not be
  observable either. Those cameras will not pass this qualification even after #47, and the honest outcome then
  is a documented per-family limitation rather than a plugin change.
- **Related:** #1042 (the same class of silence on the snapshot path), #1004 (disabled presentation, whose
  snapshot half is already live-proven).

lenoxys:
Closes #47. Closes #48.

**Stacked** on #53 → #50 → #49. Base is `fix/recycle-session-on-warm-timeout`; review only the last commit, or wait for the stack to land.

Two issues in one area: what a caller can know about whether a camera is on. #48 explicitly points at #47 as related, and neither is complete alone — #48 without #47 ships a typed "observable" for a value that never moves.

## #47 — the observation could not change

**Diagnosed live, not from the code**, because the issue named two candidate mechanisms and only one is visible from outside. `get_device_param_list` is **owner-gated**, and on a shared/member account it is refused for every device, forever:

```
/app/devicemanage/get_device_param_list failed (200/20004): Only the owner can change settings.
```

`record()` swallowed that in a bare `catch {}` — not even routed to `onError` — and fell back to the cached device-list params, which it only fetched when it had **none at all**. So the read-through refresh ran happily every ~15 s and re-applied the same values with a fresh timestamp. Measured on a real fleet, the entry's age climbed 8.5 → 13.5 → 18.5 s then reset to 4.9 s, over and over, while the value never moved. Nothing distinguished that from a device whose state genuinely never changed.

Note this is **not** the mechanism the issue guessed (`applyParams` renewing `ts` from unrelated params — that loop only touches params present in the incoming map). I've corrected the issue.

`getDeviceParamList`'s own JSDoc already names the fix: the device list carries the same `{param_type, param_value, update_time}` and is **not** owner-gated. So when the overlay is refused, the list is re-fetched. The refusal is reported once per device, and not retried — it's a property of the account, so asking again spends a request to learn the same thing.

## #48 — privacy is unobservable, and nothing said so

Implemented as the issue's **option 2**. Option 1 needs the `6250` read polarity; a live read did show that parameter reported alongside `1035` on one of those families, but **only one polarity has been observed**, and aliasing it into `enabled` on a single sample is the guess the ground-truth rule forbids. Evidence is recorded on the issue for whoever captures the other half.

`unobservableMembers(dev.camera())` names the members a device accepts but never reports back, derived from the same table the getters and setters come from, per device. `cam.privacy === undefined` reads identically for a device reporting the value as unset and one that never reports it — and guessing between them is exactly what a caller must not do.

Carried on a **symbol**, mirroring `core/contracts`' command-observation metadata: it describes the surface rather than being a member of it, so it must not appear among a capability's own keys. (I tried it as a plain property first; two existing specs correctly caught that it had joined every capability's member list.) `unexposed` members are deliberately excluded — the device *does* report those; what's missing is a confirmed meaning for the value.

`enabled`'s JSDoc now records the consequence on the families whose power rides the privacy envelope: the write goes there while the read still observes 1035/2001, so `setEnabled(false)` succeeds without moving the value.

## Live proof of #47, end to end

Write issued from a **second, freshly constructed client**; the client under test never restarted and only ever reading — the issue's own methodology.

| | result |
| --- | --- |
| `beta-0.1.0` | write acked in 2.2 s, then `false` for **52 s — no change observed** |
| this branch | write acked in 1.0 s, **`false → true` observed at 21.0 s** by the long-lived client |

That is the measurement the issue reports as impossible.

**On the device write:** confirmed with the maintainer beforehand, per CONTRIBUTING's read-only default. The target was chosen because it was **already disabled**, so the sequence was off → on → off — no camera was ever left disabled — and the final state was verified back to `enabled:false` from a fresh client. The restore needed a second assertion because the cloud takes ~20-30 s to propagate, which the script handles in a `finally` with a rescue path.

## Not covered

- No **change event**. The issue's "Related gap in the same area" asks for one so a consumer can act on the transition instead of re-reading. The machinery exists (`events: [{source:"poll", …}]` + the poll diff), but `pollOnce` calls `decodeCapabilityEvent` without a capabilities argument, so a contested param id emits nothing — worth its own change rather than bolting on here. Re-reading now works, which is what made the observation unusable.
- The owner-gated path is what this fleet exercises. An **owner** account still takes the overlay, unchanged; that branch is covered by unit test only.

`npm run verify` green: 131 files, 2399 tests (whole stack).


lenoxys:
## Review pass — two wrong answers were shipping

**`unobservableMembers` named members with no setter.** It gated on `available` alone, so every `unverified` write-only member was reported as something the device "accepts" — six on the lock table, all with `setX === undefined` and an intent path that throws. That is the unverified-write guess restated by another route. Now collected inside the bind loop, after the same `write && !unverified && installs` decision that installs the setter (which also removes a second `available` evaluation that could drift from the first).

Before → after, on the real tables:

```
LOCK   unobservable: [oneTouchLock, scramblePasscode, wifiStatus,     →  []
                      logEnabled, privacyMode, oneTouchRearLock]
       (every setter undefined)
```

**#48's point (b) is now addressed, not just documented.** Both review axes flagged that `enabled` on the families whose power rides the privacy envelope was left as prose — and that is the case the issue cares most about: *"a camera that has been turned off is also indistinguishable from one that is on, because the observation those devices would move is the one that is not readable."* A readable value that silently disagrees with its own setter is worse than an unreadable one, because a caller has no reason to distrust it.

A member can now declare `readReflectsWrite`, and `unreflectedMembers(dev.camera())` names it per device:

```
INDOOR_COST_DOWN_CAMERA (mini)   unreflected: ["enabled"]
ordinary indoor cam              unreflected: []
```

## Also fixed

- **The overlay latch fired on any error.** One timeout or expired session would permanently give up the freshest param source for a device on an account entitled to it. Now latches only on the owner-gated refusal (by code `20004` or wording); a transient failure falls back for that call only.
- **Request amplification.** The fallback re-fetched an account-wide list (`get_house_list` + one `get_devs_list` per house) per device per refresh, and every fetch clears the capability caches — at a 15 s freshness window those caches would have stopped working entirely. Concurrent refreshes now coalesce onto one in-flight fetch.
- The double list fetch on first refusal is gone; one condition was spelled twice.
- The `6250` evidence names what was measured (T8170 and T8171 each returned `6250="0"` alongside `1035="0"` while streaming) rather than "a live read of a fleet member", which was unfalsifiable next to neighbours that cite theirs. Still no polarity claim — that is one reading, once.
- **Dropped a vacuous spec.** It asserted the metadata was absent from `Object.keys(getOwnPropertyDescriptors(...))`, which never lists symbol keys, so it passed however the metadata was stored. It now asserts the statement survives neither a spread nor `JSON.stringify`.
- Matched `core/contracts`' descriptor (`configurable: true`) and moved the symbols above their use.

## Deferral now tracked

The #47 event gap is filed as #55 — the poll path calls `decodeCapabilityEvent` without a capabilities argument, so a contested param id resolves to nothing and a declared poll event silently never fires. That is a poll-diff bug in a different subsystem, not camera-specific.

## Re-verified live

`false → true` observed at **21.0 s** by the long-lived client after the changes; camera returned to `enabled:false` and verified from a fresh client.

`npm run verify` green: 131 files, 2410 tests.

lenoxys:

PR #58, opened 2026-08-22
fix(media): make live media trustworthy — a host's ffmpeg, primed decoding, recovered sessions, per-camera frames, observable enablement

### LINKED ISSUE #35: Media paths spawn a bare ffmpeg, so hosts without a system ffmpeg cannot use snapshotLive/record

### LINKED ISSUE #36: snapshotLive() intermittently fails decode with non-existing PPS 0 referenced

### LINKED ISSUE #38: A camera stops delivering live frames for every later session until the client restarts

### LINKED ISSUE #47: A long-lived client never observes camera.enabled changing

### LINKED ISSUE #48: Privacy mode is unobservable, including where it is the camera's power state

### LINKED ISSUE #51: A camera's live burst arrives truncated, so no complete frame can be decoded from it
## Observation

On one camera model, a live burst collected for a still contains no decodable complete frame: ffmpeg reports bitstream truncation and then exits **0** having encoded nothing.

```
[h264 @ 0x...] error while decoding MB 105 36, bytestream -28
```

`bytestream -28` means the decoder ran off the end of a slice — the access unit it was given is short of the bytes its own headers promise.

## Why this is filed separately

It was previously masked. On `beta-0.1.0` the same camera failed with `non-existing PPS 0 referenced`, because the burst also lacked parameter sets, and that error came first. With the parameter sets primed (#36), the decoder gets past the headers and the truncation is what remains.

So this is not a regression from that change — measured on the same fleet with the same script, the camera failed on the baseline too (3 of 4 attempts). The reason it reports is simply now accurate.

## Characterisation

- Two devices of the same model, both affected; six other cameras in the same fleet, same client, same run, are not.
- The failing attempts are consistently **fast** (2.7–3.7 s) while the same camera's successful attempts take ~9–10 s. So the failure is not a timeout — a burst is collected, it just cannot be decoded.
- The camera does succeed sometimes, so this is per-acquisition, like the framing issue was.
- `record`/`openReadable` on the same camera were not measured; it is not yet known whether the truncation is specific to the snapshot collection window or affects every egress.

## Where to look

Truncation before the decoder points at frame reassembly rather than the media pipeline: the burst is whatever `LiveVideoFrame.data` carried, so a short access unit means a fragment was dropped or a payload was cut before it reached the collector. Worth checking whether the affected model fragments differently (larger frames, more fragments per unit), and whether a dropped fragment is currently silent.

## Impact on a caller

`snapshotLive` rejects with `LiveSnapshotUnavailableError`, reason `undecodable-burst`, `retryable: true`. That is honest — a retry does sometimes succeed — but a caller that rate-limits acquisition spends repeated attempts on a camera that fails most of them, which is the cost the typed reason was meant to reduce.

## Environment

- SDK: `beta-0.1.0` and the #36 branch, both measured
- Node.js 24.19.0, Debian 13
- ffmpeg 7.1.5

### LINKED ISSUE #52: snapshotLive reports the frame header's resolution, which can disagree with the image it returns
## Observation

`snapshotLive()` returns `{ jpeg, width, height }` where `width`/`height` come from the P2P frame header, not from the image that was produced. On a camera whose stream changes resolution mid-burst the two disagree:

| | value |
| --- | --- |
| returned `width`/`height` | 2304x1296 |
| the returned JPEG's own SOF0 | 1280x720 |

ffmpeg's decode of that same burst shows the switch:

```
Reinit context to 1280x720, pix_fmt: yuv420p
Reinit context to 1920x1088, pix_fmt: yuv420p
Reinit context to 1280x720, pix_fmt: yuv420p
```

So the burst carried three configurations. The header value is whichever frame started the capture; the image is whatever ffmpeg emitted for the one frame it was asked for.

## Scope

Measured across eight cameras. Seven agreed exactly (1920x1080, 2304x1296, 640x480 — each matching its JPEG's SOF0). Only the camera with the mid-burst resolution change disagreed, so this is not a widespread mislabelling — it is specific to a stream that reconfigures.

## Why it matters

The dimensions are the only description of the image the SDK returns, so a caller sizing a buffer, choosing a scaler, or caching by resolution is told something the bytes contradict. There is no way for a caller to detect the disagreement short of parsing the JPEG itself, which is work the SDK already had the answer to.

## Options

1. Report the dimensions of the image actually produced (read them back from the encoded JPEG, or from the decoder rather than the frame header). Authoritative by construction.
2. Keep the header value but state in the type that it describes the stream at capture start and may differ from the returned image — honest, but pushes the parse onto every caller.

The first looks right: the return value describes an image, so the image should be its source of truth.

## Related

The multi-resolution burst itself may deserve its own question — a stream that reconfigures twice inside one short collection window is also a reason a burst decodes to fewer frames than expected.

## Environment

- SDK: `beta-0.1.0` and the #36 branch, same result on both
- Node.js 24.19.0, Debian 13, ffmpeg 7.1.5

### LINKED ISSUE #55: No event for a cloud-param change a capability declares, because the poll path drops contested ids
Split out of #47, whose "Related gap in the same area" asks for a change event so a consumer can act on a transition instead of re-reading for it. PR #54 fixes the re-read half; this is the other half, and it is not camera-specific.

## The machinery already exists

A capability can declare `events: [{ source: "poll", match: <paramId>, emit: "<name>" }]`, and the poll loop diffs cloud params and turns each change into a semantic event:

```
pollOnce → registry.pollChanges() → for (change of diff.params)
             decodeCapabilityEvent({ source: "poll", ...change }) → emitSemantic(...)
```

So declaring a poll event for camera enablement (1035 / 2001) ought to be all it takes.

## Why it would not fire

`pollOnce` calls `decodeCapabilityEvent` with **no capabilities argument**:

```ts
for (const change of diff.params)
  for (const out of decodeCapabilityEvent({ source: "poll", ...change }))
    this.emitSemantic(out.event, out.payload, { refresh: out.refresh });
```

Without it, `resolveHits` cannot disambiguate a param id claimed by more than one capability, and a contested id resolves to nothing. The push path passes `this.capsForEvent(ev.deviceSn)`; the poll path does not. So a poll event on a contested id is declared and silently never emitted.

Compare the two call sites — the push one has the argument, the poll one does not, and nothing marks that as deliberate.

## Why it matters beyond one capability

Every observation that only ever arrives as a cloud param is in this position: re-reading is the only mechanism, and re-reading cannot tell a consumer *when*. `arming.mode` works around it with a push event plus a bounded readback, but that needs a push id to hang off. A state that has no push event of its own has nothing.

## What resolving it looks like

1. Pass the device's resolved capabilities into the poll path's `decodeCapabilityEvent`, so a contested id resolves the same way it does on the push path.
2. Then a capability can declare its poll events and they will actually fire — camera enablement being the case #47 wanted.

Worth checking while there: whether an uncontested id emits today (it should), so the fix is scoped to the contested case rather than the whole path.

## Not this issue

The freshness problem — that the value being diffed was itself frozen on an owner-gated account — is #47, fixed in PR #54. This is only about the transition never being announced.

### LINKED ISSUE #56: Live media leaks across channels: every stream on a station receives every warm camera's frames
## Observation

When two cameras on one HomeBase have a warm live source at the same time, every `LiveStream` on that
station emits **both cameras' media**. Each consumer receives an interleaving of two unrelated elementary
streams, on one `video` event sequence, with no way to tell the frames apart.

Reproduced with one client, on a HomeBase (`sn*2199`) carrying three cameras. Two of them differ in coded
geometry, which makes the leak directly visible:

```
--- one camera warm at a time ---
sn*0ED4 T8114  video=h264 1920x1080 x150                      audio=aac-lc
sn*22DA T8210  video=h264 640x480 x79, h264 1920x1080 x3      audio=aac-lc

--- both cameras warm together on one station ---
sn*0ED4 T8114  video=h264 640x480 x20, h264 1920x1080 x15     audio=aac-lc
sn*22DA T8210  video=h264 640x480 x21, h264 1920x1080 x15     audio=aac-lc
```

`sn*0ED4` streams 1080p and nothing else when it is alone. Warmed alongside the doorbell it emits 20
`640x480` frames against 15 of its own — more than half of what it delivered belonged to another camera.
The doorbell sees the mirror image. The proportions match because the fan-out is unconditional, not racy.

The three trailing `1920x1080` frames in the doorbell's solo run are the same defect with a 4 s gap
instead of a concurrent one: the previous camera's source was still inside its 8 s linger.

## Cause

`src/transport/p2p/live-stream.ts`, `onFrame`, states the assumption and why the filter was removed:

```ts
// NOTE: no channel filter here — the station tags inbound media frames channel 0 regardless of
// which camera channel was started, so filtering by `opts.channel` would drop everything. One
// LiveStream maps to one started camera; emit all video/audio frames it receives.
```

"One LiveStream maps to one started camera" holds only while one camera per station is streaming. Every
`LiveStream` built over a station's `P2PSession` subscribes to the same frame feed, so with N warm cameras
each of the N streams receives all N cameras' frames.

The cache key is not the problem: `sharedLiveSourceFor` keys by `${parentSn}:${channel}` and the three
cameras resolve to `sn*2199:0`, `sn*2199:1` and `sn*2199:2` — correctly distinct. Each gets its own
`SharedLiveSource` and its own `LiveStream`; the streams then all read one undifferentiated feed.

Standalone cameras are unaffected: they own their session, so the fan-out has one member.

## Consequences observed downstream

- **Snapshots of different cameras come back byte-identical.** Three cameras on this HomeBase retained the
  same image in normal use, and three tiles in a HomeKit consumer showed the same picture. A snapshot burst
  is assembled from "the frames that arrived", which is now several cameras' frames.
- **`LiveSnapshotUnavailableError: undecodable-burst` ("no complete frame in the burst")** fires constantly
  on these cameras. A burst containing fragments of two encoders' streams has no single decodable picture.
- **A decoder fed one of these streams reports exactly what two merged streams look like:** `Duplicate POC
  in a sequence: n`, `Could not find ref with POC n`, `Error constructing the frame RPS`, `Skipping invalid
  undecodable NALU: 1`, continuously. Two independent encoders each number pictures from zero.
- **A live consumer that reacts to a geometry change re-initialises continuously**, because the geometry
  alternates every few frames between the two cameras.

Every one of these disappears when only one camera on the station is warm; the same camera alone streams
150 frames in 10 s with no decoder complaint.

## What to build

Media frames must reach only the stream that started that camera. The station tagging inbound frames
channel 0 means the channel field cannot be the discriminator, so the demultiplexer needs something that
is: a per-start stream identifier if the protocol carries one, or — failing that — an explicit constraint
that a station serves one live channel at a time, enforced in `sharedLiveSourceFor` rather than assumed in
a comment.

Refusing or serialising a second concurrent channel would be a correct and honest fix. Silently
interleaving is not, because no consumer can detect it.

## Acceptance criteria

- [ ] With two cameras of different coded geometry warm on one station, each handle emits only its own
      camera's frames, or the second acquisition is explicitly refused.
- [ ] A live snapshot never returns another camera's image.
- [ ] The linger window does not deliver a departed camera's frames to the next camera warmed.
- [ ] A regression test covers two concurrent channels on one station.

## Related

- The same cameras also split single access units across several `video` events; separate and independent,
  measured with one warm camera: #57.

### LINKED ISSUE #57: video event delivers byte-stream chunks, not access units, and flags continuations as keyframes
## Observation

`LiveVideoFrame` is documented as "one decoded video access unit, as Annex-B", but the `video` event delivers
fixed-size byte-stream chunks, and `keyframe` is repeated on the continuations.

Logged from a real 1080p H.264 camera, printing the Annex-B NAL unit types found in each delivered
`frame.data` (empty means the chunk contains no start code at all, so it is a continuation of the previous
one):

```
  1234ms  keyframe=yes  bytes= 64000  nals=SPS,PPS,SEI,IDR
  1244ms  keyframe=yes  bytes= 24814  nals=
  1254ms  keyframe=no   bytes= 12625  nals=non-IDR
  1452ms  keyframe=no   bytes= 64000  nals=non-IDR
  1463ms  keyframe=no   bytes= 27606  nals=
  2124ms  keyframe=no   bytes= 64000  nals=non-IDR
  2161ms  keyframe=no   bytes= 64000  nals=
  2174ms  keyframe=no   bytes= 42434  nals=
```

Every access unit larger than 64000 bytes is split, and the second chunk of a keyframe is itself flagged
`keyframe: true` while carrying no parameter sets and no NAL header of its own.

## Why it matters

A consumer that pipes the bytes onward is unaffected, but the contract invites two things that then break:

- Deciding anything per access unit — "start on the first keyframe", "switch codec at a keyframe", "count
  frames" — silently operates on chunks. A geometry or codec change applied at a chunk boundary cuts an
  access unit in half.
- `keyframe` cannot be trusted as "this buffer is independently decodable", which is the only reason a
  consumer would read it.

On H.265 sources the downstream decoder reports `Duplicate POC in a sequence`, `Skipping invalid undecodable
NALU`, `Could not find ref with POC n` and `Error constructing the frame RPS` throughout a session, and the
same cameras fail `snapshotLive` with `no complete frame in the burst` most of the time, which suggests the
splitting is not only a contract question.

## Acceptance criteria

- [ ] Either `video` delivers whole access units as documented, or the contract states that it delivers a
      byte stream and `keyframe` is defined for a chunk.
- [ ] `keyframe` is true only where a consumer may begin decoding.
- [ ] H.265 frame assembly produces a stream that decodes without reference errors.

### LINKED ISSUE #59: A kicked-out session resolves login() as ok and getDevices() as empty, and nothing re-authenticates
## Observation

A cached session the cloud has since invalidated makes `login()` resolve **`Ok`** and the first
`getDevices()` resolve with an **empty list**. Nothing re-authenticates, so every later call fails the same
way for the life of the client, and a host is looking at an account that appears to have no devices.

Reproduced on a client whose persisted session had been displaced by another login on the same account
(serials and account hash redacted):

```
[mega] restored persisted session for <redacted>
[mega] err envelope: {"code":26084,"msg":"token does not exist because it was kicked out"}
[eufy] /app/house/get_house_list failed (401): token does not exist because it was kicked out
[mega] key/exchange (anon) -> 200 …
[mega] err envelope: {"code":401,"msg":"token error"}
[eufy] /app/house/get_devs_list failed (401): token error
```

The caller's view of that run:

```ts
const r = await eufy.login();     // { status: "ok" } — from the restored session
const devices = await eufy.getDevices();  // []  (resolves; does not reject)
```

## What is already right

- The transport classifies it correctly: 26084 is a recognised auth failure, and it throws
  `SessionExpiredError` and clears the persisted session (`mega-client.ts`, covered by
  `session-expiry.spec.ts`).
- `DeviceRegistry.getDevices` marks the refresh partial, so the poll diff correctly suppresses `removed` —
  a kicked token does **not** announce every device as removed. That hazard is already handled.
- The failure does reach `eufy.on("error")`.

## What is missing

**Nobody re-authenticates.** The client holds the credentials that produced the dead session, and the one
error that says "this token is finished, log in again" is the one error it could act on by itself. Instead
`getDevices()` treats it like a transient outage: report and return what it has, which on a fresh client is
nothing.

Two consequences:

1. **`login()` reporting `Ok` is not a statement that the session works** — it is a statement that one was
   restored from the store. The first authenticated call is where a caller finds out, and it finds out as an
   `error` event beside a successful-looking empty result.
2. **A host that treats an empty device list as truth** will tear down every accessory it had. Nothing in
   the resolved value distinguishes "this account has no devices" from "this client is no longer
   authenticated".

## Where to look

- `MegaHttpClient.postSigned` — the `authFailure` branch that throws `SessionExpiredError`.
- `DeviceRegistry.getDevices` — the `catch` that sets `partial` and calls `onError` for BOTH the house and
  the device-list queries; a `SessionExpiredError` is not the same kind of failure as a house-query outage.
- `EufyMega.login` — resolves from the store without an authenticated round trip.

## What resolving it could look like

1. On `SessionExpiredError` from an authenticated call, re-run the login state machine ONCE with the
   credentials already held, then retry the call. A single retry, so a genuinely rejected account cannot
   loop.
2. Failing that (no credentials, or the re-login needs a captcha or 2FA the host must answer), let the
   rejection PROPAGATE out of `getDevices()` rather than resolving with an empty list — a caller can handle
   a rejection; it cannot detect a lie.
3. Either way, distinguish it from a partial outage: a rejected session is not a subset of the account.

Worth deciding whether `login()` should verify a restored session with one cheap authenticated call, so
"ok" means what a caller reads it as.

## Environment

- SDK: current `beta-0.1.0` line (observed while live-testing the branch for #58)
- Node.js 24.19.0, Debian 13

### LINKED ISSUE #60: A poll event is swallowed when the realtime path updates the same record: the poll baseline shares its objects
## Observation

A cloud-param change is announced only when the realtime channel is **off**. With `autoRealtime` on — the
default — the same change on the same device produces no event, because the poll baseline was rewritten in
place before the poll could diff against it.

Measured on one device, twice, with the same write (a camera's enablement, `2001` `"false"` → `"true"`),
observed by a long-lived client that only ever read:

| observer | result |
| --- | --- |
| `autoRealtime: false`, `pollOnce()` driven directly | `{ paramType: 2001, from: "false", to: "true" }` — announced |
| `autoRealtime` default (P2P sessions warm), `pollMs: 15000`, 90 s | **nothing**, while the same client's re-read moved to `true` |

The read moving while the event never fires is the tell: the value reached the registry, and the poll then
found nothing to report.

## Cause

`DeviceRegistry.pollChanges` snapshots the device OBJECTS, not their values:

```ts
this.pollSnapshot = { devices: new Map(devices.map((d) => [d.sn, d])), complete: !this.lastRefreshPartial };
```

`this.devices` holds those same objects, so every path that updates a device's params in place also updates
the baseline the next pass diffs against:

```ts
const now = dev.params ?? {};
const was = prev.params ?? {};   // ← the same object as `now` once anything has mutated it
```

The inbound realtime path does exactly that — `applyRealtimeState` folds the params a station volunteers
into the device's record — so on a device whose station reports over P2P, the station's own report of the new
value erases the poll's evidence that it changed. The louder the realtime channel, the fewer poll events.

The field's own JSDoc states the invariant this breaks:

> Held separately from `devices` because that cache is refreshed by anything that needs a device … Diffing
> the shared cache in place would let any of those silently absorb the delta, and the next poll would then
> see an unchanged account and emit nothing.

The MAP was separated; the objects inside it were not.

## Scope

Every `source: "poll"` event mapping is affected — `batteryLevel`, `contactState`'s poll arm, and any
capability that adds one. `reported` is affected the same way: it compares `dev.lastSeenMs` against
`prev.lastSeenMs` on what may be the same object, so a `deviceState` can be lost for the same reason.

`added`/`removed` are unaffected — they use the snapshot for keys only.

## What resolving it looks like

Snapshot the VALUES the diff reads, not the records:

```ts
this.pollSnapshot = {
  devices: new Map(devices.map((d) => [d.sn, { lastSeenMs: d.lastSeenMs, params: { ...d.params } }])),
  complete: !this.lastRefreshPartial,
};
```

A shallow copy of the param map is enough: the values are strings.

Worth a regression test that mutates a device's params between passes — the current specs pass because they
drive the poll with no realtime path mutating anything underneath, which is precisely the configuration in
which the bug does not appear.

## Environment

- SDK: current `beta-0.1.0` line (found while live-verifying the branch for #58)
- Node.js 24.19.0, Debian 13

### LINKED ISSUE #61: Every fMP4 fragment is undemuxable: trun declares composition-time offsets it never writes
## What happens

`Fmp4Muxer` writes every `trun` box with `tr_flags = 0x000f01`, which sets
`sample-composition-time-offsets-present` (`0x000800`) in addition to data-offset, sample-duration,
sample-size and sample-flags. The sample body, however, writes only three 4-byte fields per sample.

Every `trun` is therefore `4 * sample_count` bytes shorter than its own flags require, so a conforming
parser runs past the end of the box while reading the sample table. FFmpeg rejects the whole stream:

```
[mov,mp4,m4a,3gp,3g2,mj2] overread end of atom 'trun' by 268 bytes
[mov,mp4,m4a,3gp,3g2,mj2] overread end of atom 'trun' by 284 bytes
[mov,mp4,m4a,3gp,3g2,mj2] error reading header
Error opening input: Invalid data found when processing input
```

This affects **every** consumer of the muxer, so `recordFragments()` and `record()` cannot produce
media any standard tool or HomeKit Secure Video controller will accept.

## Evidence

`src/transport/p2p/fmp4.ts` — the JSDoc states the intended flag set, and the value disagrees with it:

```ts
private trun(samples: Sample[]): Buffer {
  // flags: data-offset(0x1) + sample-duration(0x100) + sample-size(0x200) + sample-flags(0x400)
  const flags = 0x000f01;
  const parts: Buffer[] = [u32(flags), u32(samples.length), u32(0) /* data_offset patched later */];
  for (const s of samples) {
    parts.push(u32(s.duration), u32(s.data.length), u32(s.keyframe ? 0x02000000 : 0x01010000));
  }
  return box("trun", Buffer.concat(parts));
}
```

The comment describes `0x000701`. The code writes `0x000f01`.

Measured on a real wired camera's `recordFragments({ fragmentSeconds: 4 })` output, every `trun` in
every `moof` is short by exactly one 4-byte field per sample:

| samples | declared size | size required by flags | delta |
| --- | --- | --- | --- |
| 67 | 824 | 1092 | -268 |
| 71 | 872 | 1156 | -284 |
| 60 | 740 | 980 | -240 |
| 63 | 776 | 1028 | -252 |
| 90 | 1100 | 1460 | -360 |
| 103 | 1256 | 1668 | -412 |

Each declared size is exactly `20 + 12 * sample_count`, which is the header plus three fields per
sample — confirming the body is correct and the flags are wrong.

## Proof of the fix

Patching only the flag value from `0x000f01` to `0x000701` in a captured fragment stream, changing no
other byte, makes the same stream demux cleanly:

```
Duration: 00:00:14.46, start: 0.000000, bitrate: 823 kb/s
Stream #0:0: Video: h264 (High) (avc1), yuv420p, 1920x1080, 795 kb/s, 15 fps, 90k tbn
Stream #0:1: Audio: aac (LC) (mp4a), 16000 Hz, mono, fltp, 24 kb/s
frame=217 time=00:00:14.40
```

## Why the specs did not catch it

`src/transport/p2p/__tests__/fmp4.spec.ts` reads sample fields at absolute offsets inside `trun`, which
stays green because the body layout is right. Nothing validates a box against its own declared flags,
and nothing feeds the muxer's output to a real demuxer.

## Suggested fix

1. Set `tr_flags` to `0x000701`, matching the fields actually written.
2. Add a spec that derives the required `trun` size from the flags it declares and asserts the declared
   box size equals it, so any future field/flag divergence fails immediately.

## Secondary observation, filed separately if wanted

Once the stream demuxes, the audio track's decode times are non-monotonic across fragment boundaries:
transcoding the patched capture reports `Non-monotonic DTS; previous: 6360, current: 6296` and similar
at several boundaries. Video is monotonic. This is tolerated by FFmpeg but is a real timeline defect.

lenoxys:
Closes #35. Closes #36. Closes #38. Closes #47. Closes #48. Closes #51. Closes #52. Closes #55. Closes #56. Closes #57. Closes #59. Closes #60.

The former stack #49 → #50 → #53 → #54 → #58, collapsed into one PR — 21 commits, four superseded PRs closed (their bodies stay readable as the long-form write-up of each part).

**Stacked on #46**, so this is based on `fix/surface-session-expired`, not `beta-0.1.0`. Two consequences worth knowing before you read the diff:

- **#46 is currently 13 commits behind `beta-0.1.0`** (it predates the clean-line merge), so this branch is too. `npm run verify` here therefore runs without that work — one spec file's worth, 24 tests. Nothing is lost; it comes back the moment #46 is brought up to date, and this PR only becomes whole once #46 lands.
- **The token classifier is a genuine union, not a conflict resolution.** #46 widens it and names the gateway codes; my #59 commit widens it for the `401 "token not exist"` wording and adds token replacement with adoption and backoff. Merged so his named codes are the single source of truth *inside* my classifier, his distinct `sessionExpired` event survives untouched, and his `authFailure` branch is replaced by the recovering one — a dead session is now both named and recovered. That resolution is the only place his design and mine had to agree; worth his eyes.

## The spirit of it

A host could not trust anything the SDK said about live media. It couldn't run the decoder on a managed install, couldn't decode the burst it was handed, couldn't recover a camera that failed once, was quietly given a *different camera's* frames, and couldn't tell whether a camera was even switched on. Twelve issues, all the same shape: the SDK answered, and the answer was wrong or unfalsifiable.

Every part was verified against a real 8-camera fleet, one of them a HomeBase carrying three, with the host's own bridge stopped. Three of these fixes were **wrong in their first form and only hardware said so** — that's called out per part, because reviewing the code without knowing it would look like reviewing a different plan.

## Five parts, in commit order

**1 — a host names the ffmpeg binary (#35, was #49).** Managed installs have no `ffmpeg` on `PATH`, and the only workarounds were a second unmanaged install or mutating `process.env.PATH` process-wide. Now a client option names it, defaulting to today's `PATH` lookup. The availability probe moved out of the WebRTC layer, where its own hardcoded spawn meant a host-supplied build reported "missing" and the container mux **silently degraded to a raw stream** — a caller asking for `.mp4` got something else. Probe and spawn now resolve identically, so they cannot disagree. That rename is the one breaking change here.

**2 — prime the decoder (#36, was #50).** Framing, not capability: a camera announces its parameter sets *once*, with the first keyframe, and the snapshot path skips exactly that frame. Whether an attempt landed before or after the only announcement was luck — the reported alternation on one camera. Only the shared source sees the stream from the start, so it retains the sets and every burst is primed with them. Failures are now typed with a reason and a derived retryability, because a burst can also fail for reasons no priming fixes. `non-existing PPS 0`: **4 of 16 attempts on base, 0 of 40 here.** Two live extras: the JPEG encoder was refusing the limited-range input cameras actually signal, and a burst yielding no image while ffmpeg exits 0 was being reported as ffmpeg's fault. Honest reading — the success *count* is unchanged; what changed is which failures happen and whether a caller can act on them.

**3 — recover a camera whose live start failed (#38, was #53).** A failed start left its consumers attached, the router only replaced a source at zero consumers, and the source rebuilt every stream over the session it was born with. So one failure latched the camera dead until the process restarted, restarting being the only thing that built a new session. Now a stream that tears down having **never delivered a frame** kills its source and recycles the device's session. First version keyed on the warm-up deadline and missed the case the issue actually records; enumerating the ways a start can fail is how the one that happens gets left out, so one rule covers all of them. Shared HomeBase sessions are never closed to save one camera. Base: `ok, fail, fail, fail`. Here: a rebuilt session after every failure, and 24/24 on a full-fleet run.

**4 — observe whether a camera is on (#47 #48, was #54).** The read that backs enablement is owner-gated, and on a member account it is refused for every device, forever — swallowed silently, then papered over by re-stamping cached values with a fresh timestamp. So the value looked live and never moved: nothing distinguished it from a camera whose state genuinely never changed. The non-gated device list carries the same values, so that's what's read on refusal, reported once per account rather than retried. And where a device accepts a member it never reports back, the SDK now *says so* instead of returning `undefined` for both "unset" and "unknowable". Base: a write acked, then 52 s with no change observed. Here: observed at 21 s by a client that never restarted.

**5 — one camera's frames, whole (#51 #52 #55 #56 #57 #59 #60, was #58).** A station's inbound feed *is* tagged per camera — the code asserted otherwise, so with two cameras warm each stream received both. That single fact explains byte-identical snapshots across cameras, undecodable bursts, duplicate picture-order counts, and none of it was detectable by a consumer. Large access units also arrive split across frames with the header repeated verbatim, so continuations were emitted as units of their own: no start code, flagged as keyframes, carrying no parameter sets. They're reassembled by identity now, and a unit whose tail never arrives is dropped and reported rather than handed to a decoder that runs off the end of it. Also here: snapshot dimensions now describe the image rather than whichever frame started the capture; a poll change is decoded with the device's capabilities so enablement transitions actually emit; the poll baseline stopped sharing the records it diffs; and a cloud-rejected token stopped resolving as *an account with no devices*, which a host acts on by deleting everything it had. Foreign frames per stream: **28 and 9 on base, 0 and 0 here.** Start-code-less events and setless keyframes: **0 across the fleet.** Two cameras on one HomeBase now stream at once.

## Worth a second opinion

- Recycling a standalone session hard-closes it rather than waiting for idle — the failed source still counts as a user, so a deferred reset would never fire. For a standalone device the interrupted work is on the same dead camera, which I judged acceptable.
- The attached-camera recovery path and the owner-account read path are unit-tested only; this fleet has no attached camera on a shared station and the account is a member.
- H.265 assembly is expected to follow from channel isolation, not claimed — no H.265 camera here to measure.
- An E2E camera's split frame still yields nothing: the body is sealed under one tag, so a first chunk can't be opened.
- One-shot recording still opens its own pull instead of joining the shared source, costing a second stream on a camera already streaming.
- A caller holding a source that got replaced must re-acquire it; the handle now throws instead of silently reviving a dead stream.

`npm run verify` green on this tip, rebased onto #46: 131 files, 2455 tests. Docs build green. (On `beta-0.1.0` it was 132 / 2479 — the difference is the clean-line spec that #46's base predates.)


lenoxys:
Linked issues, since GitHub only registered five of the twelve closing keywords from the body (the base here is `beta-0.1.0`, not the default branch, so nothing auto-closes on merge regardless): #35, #36, #38, #47, #48, #51, #52, #55, #56, #57, #59, #60.

lenoxys:
Rebased onto `fix/surface-session-expired` (#46) — 21 commits replayed, 20 applied clean. Two conflicts, both where #46 and #59 touch the same classifier, resolved as a union: #46's named `EufyCloudErrorCode` values are now the single source of truth inside `tokenRejected`, its distinct `sessionExpired` event is untouched, and its `authFailure` branch is replaced by the recovering path. One further conflict came only from the stale base (an option added next to one this base predates).

`npm run verify` green: 131 files, 2455 tests. The pre-rebase tip is kept locally as `backup/live-media-integrity-prerebase` (3460a6f) if the old base is ever wanted back.

@max246 — the classifier resolution is the one place our two designs had to agree, so it's worth your eyes. Also note #46 is 13 commits behind `beta-0.1.0`, which this branch now inherits; refreshing #46 onto current beta fixes it for both.

src/model/capabilities/__tests__/unobservable-members.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+  type: "bool" as const,
+  kind: "boolean" as const,
+  writeOnly: true as const,
+  provenance: "verified" as const,
+  description: "a setting the device never reports back",
+  write: () => ({ kind: "set-param" as const, param: 1, value: 1, channel: 0 }),
+};
+
+const readableMember = {
+  param: 1035,
```
max246, line 46: is this part of the p2p enum?

src/transport/p2p/live-stream.ts
```diff
@@ -103,6 +121,8 @@ export class LiveStream extends EventEmitter {
   ) {
     super();
     this.logger = opts.logger ?? noopLogger;
+    // Only a HomeBase serves several cameras over one session, so only there is there anything to tell apart.
+    if (opts.homeBaseAttached && opts.channel !== undefined) this.mediaChannel = opts.channel;
```
max246, line 125: This per-camera demux gate never engages on the **one-shot `record()`** path. `record()` (command-router.ts:521) destructures `{ session, channel, accountId }` — **without `homeBaseAttached`** — and calls `recordClip(session, { channel, accountId, ...opts })`, so `opts.homeBaseAttached` is undefined here, `mediaChannel` stays undefined, and `acceptsMedia` returns `true` for every frame. On a HomeBase with two cameras streaming at once, `record()`'s LiveStream then takes both cameras' frames and the assembler merges/drops across them → corrupted clip — the exact failure this PR fixes for the shared path. Same omission passes `homeBaseAttached=false` into `session.startLiveMedia` (line 154), risking the wrong encryption level for an attached camera. The live/shared paths thread it (`command-router.ts:568`, `:633`); `record()` should too.

src/client/device-registry.ts
```diff
@@ -371,6 +470,26 @@ export class DeviceRegistry {
...
+  private async refreshedList(sn: string): Promise<EufyDevice | undefined> {
+    if (Date.now() - this.listFetchedAtMs < LIST_REUSE_MS) return this.devices.find((d) => d.sn === sn);
+    this.listInFlight ??= this.getDevices().finally(() => {
+      this.listInFlight = undefined;
+      this.listFetchedAtMs = Date.now();
+    });
+    try {
+      return (await this.listInFlight).find((d) => d.sn === sn);
+    } catch {
+      return undefined;
```
max246, line 505: This blanket `catch { return undefined }` swallows `SessionExpiredError`, defeating this PR's own change to **re-throw** it from `getDevices()` (`:118`/`:126`). Path: `record()`'s `getDeviceParamList` throws `SessionExpiredError` → it isn't `MegaApiError && OWNER_ONLY_CODE`, so it falls to the generic branch → `refreshedList(sn)` → `getDevices()` rejects with `SessionExpiredError` → caught here → `undefined`. With cached devices, `record()` returns **stale params on a dead session**; with none, it throws a generic `device ${sn} not found`. Either way `getDevice`/`inspectDevice`/`commandContext` (eufy-mega.ts:1028/1272/1603) never see the expiry — the opposite of the surfacing goal. Consider letting `SessionExpiredError` propagate here (re-throw it) rather than mapping it to `undefined`.

src/client/device-registry.ts
```diff
@@ -371,6 +470,26 @@ export class DeviceRegistry {
...
+   *
+   * The list is account-wide (`get_house_list` plus one `get_devs_list` per house), so without this a refresh
+   * cycle over N devices would multiply into N of those bursts — and every fetch clears the capability caches,
+   * so they would stop working. One fetch serves every device that wants the same answer.
+   */
+  private async refreshedList(sn: string): Promise<EufyDevice | undefined> {
+    if (Date.now() - this.listFetchedAtMs < LIST_REUSE_MS) return this.devices.find((d) => d.sn === sn);
+    this.listInFlight ??= this.getDevices().finally(() => {
+      this.listInFlight = undefined;
+      this.listFetchedAtMs = Date.now();
```
max246: `.finally()` runs on **rejection** too, so a *failed* `getDevices()` still advances `listFetchedAtMs`. The freshness guard at `:481` (`Date.now() - listFetchedAtMs < LIST_REUSE_MS`) then returns the cached (possibly empty) list for the next 5s **without retrying** — so a transient network failure during a device-less `record()` suppresses recovery even after the network is back. The reuse window is meant to gate *successful* loads; move the `listFetchedAtMs = Date.now()` into the fulfilled path (e.g. `.then()`), leaving `listInFlight = undefined` in `.finally()`.

src/model/capabilities/camera.ts
```diff
@@ -552,6 +584,26 @@ export const CAMERA: CapabilityModule = {
...
+   * caller could learn it had moved, and re-reading cannot say WHEN. Every id that carries the read is
+   * mapped, from {@link enablementReads}, so the event and the getter cannot disagree about which ids
+   * those are or which polarity each reports under; the raw value is normalised to `enabled` because
+   * the two ids report it inverted from each other.
+   *
+   * Neither id is claimed by another capability, so both emit without capability context — which is the
+   * declared behaviour for an uncontested id, and safe here because both have only ever been reported by
+   * camera devices.
+   */
+  events: enablementReads().map(({ paramType, invert }) => ({
```
max246, line 600: Minor / edge case: `enablementReads()` yields both `{1035, invert:true}` and the `2001` read-alias `{2001, invert:false}`, and each becomes its **own** `poll → cameraEnabled` event. A device that reports both ids in one poll diff fires `cameraEnabled` twice for a single state — and since the two derive with opposite `invert`, a raw disagreement would emit **conflicting** enabled values. The member doc says a family reports one or the other (1035 battery/solo vs 2001 standalone), so likely benign in practice — but worth a de-dup (collapse aliases into one event that reads whichever id is present) if any family is found to report both.

src/transport/http/mega-client.ts
```diff
@@ -784,6 +900,114 @@ export class MegaHttpClient {
...
+    );
+    return false;
+  }
+
+  /**
+   * Note that the held session is working. A replacement that keeps serving calls for long enough is not
+   * contention, so the hold-off is forgotten and the next genuine expiry recovers immediately.
+   */
+  private noteSessionWorking(): void {
+    if (this.recoveries > 0 && Date.now() - this.lastRecoveryAt > REAUTH_STABLE_MS) this.recoveries = 0;
```
max246, line 974: Lower-priority nuance on the recovery back-off: `recoveries` only resets after a successful call arrives **>REAUTH_STABLE_MS (10 min)** past `lastRecoveryAt`. So a *genuine* second token expiry within that window is treated as contention — `recoveryDue()` imposes the exponential wait and the resulting `SessionExpiredError` carries `CONTENDED_SESSION_HINT` ("another client may be signed in… give each its own openudid"), even when no second client exists. The back-off itself is reasonable; it's the **hint** appended on the first genuinely-rapid re-expiry that could steer a user toward the wrong fix. Could gate the hint on an actual repeat signal (e.g. only after ≥2 held-offs) rather than every held-off recovery.

src/transport/p2p/live-stream.ts
lenoxys: @-

src/client/device-registry.ts
lenoxys: @-

src/client/device-registry.ts
lenoxys: @-

src/model/capabilities/camera.ts
lenoxys: @-

src/transport/http/mega-client.ts
lenoxys: @-

src/model/capabilities/__tests__/unobservable-members.spec.ts
lenoxys: @-

lenoxys:
Rebased onto current `beta-0.1.0` and worked through the review.

**Conflicts resolved.** #46 landed while this was stacked on it, so GitHub re-targeted this PR to `beta-0.1.0` and it went `CONFLICTING`. Rebased the 22 commits onto current beta; three conflicts, all where beta had moved: the client option added next to `tuyaAllowUnverified`, and the token classifier twice. The classifier is a union of both designs — `EufyCloudErrorCode.SESSION_KICKED` is the single source of truth inside `tokenRejected`, and I adopted your anchored `(?:token|session).*does not exist` from e4abcda rather than my looser matcher; you were right that a bare `does not exist` can clear a healthy session. `MERGEABLE` again.

**Review — three confirmed and fixed, two I am pushing back on, one answered.**

| your comment | outcome |
| --- | --- |
| `record()` misses `homeBaseAttached` | **confirmed, worse than reported** — d033479 |
| `SessionExpiredError` swallowed in the registry | **confirmed, one level higher than marked** — 1eb8392 |
| `.finally()` stamps the reuse window on rejection | mechanism real, **stated consequence cannot happen**; changed anyway — 1eb8392 |
| duplicate `cameraEnabled` from 1035 + 2001 | real structure, **not fixing**: unobserved, idempotent, and the de-dup would invent precedence |
| `CONTENDED_SESSION_HINT` on a first rapid re-expiry | **disagree** — the hint is gated by the 60 s wait, not the 10 min stable window |
| `1035` in the spec fixture | answered, and documented in the fixture — fc4ed6c |

Details in each thread. The two clip findings were the valuable ones: `record()` on a HomeBase camera was starting on the own-session path, which predates this PR and probably meant it never worked there at all.

`npm run verify` green: 132 files, 2535 tests.

max246 (approved):
Re-reviewed against my earlier round — all three correctness findings are fixed, and the two I'd flagged as minor/nuance are acceptably resolved as documented design. Ready to merge.

**Fixed (the real ones):**
- ✅ **`record()` per-camera demux** — `command-router.ts:521` now destructures `homeBaseAttached` from `resolveSession` and passes it into `recordClip`, so the one-shot record path gets the same level/topology gate as live. Covered by `live-start-failure-recovery`, `media-opts-plumbing`, `talkback-session` specs.
- ✅ **`SessionExpiredError` no longer swallowed** — `refreshedList` now does `if (e instanceof SessionExpiredError) throw e; return undefined`, so a dead session propagates instead of degrading to "no such device." Tested: `device-registry.spec.ts:118-141` (`rejects.toBeInstanceOf(SessionExpiredError)`).
- ✅ **Stale-cache-after-failure** — `listFetchedAtMs = Date.now()` moved inside `.then()` (resolve-only); `.finally()` only clears the in-flight handle. A rejected fetch no longer opens the reuse window. The new JSDoc states the invariant explicitly.

**Acknowledged, acceptable to leave:**
- `camera.ts` duplicate `cameraEnabled` (1035 + 2001 alias) — the added JSDoc documents that both ids normalise to the same `enabled` polarity and that 1035/2001 are device-family-exclusive (don't co-occur), so no conflicting emit in practice. Was flagged minor.
- `mega-client.ts` recovery back-off reset (`recoveries` clears only after `REAUTH_STABLE_MS`) — deliberate contention heuristic; a held-off caller still gets a `SessionExpiredError` + hint so the host can drive reauth. Was flagged lowest-priority.

My one open question (unobservable-members fixture using `1035`) is answered inline by the fixture's own JSDoc — it's a realistic value, not a shared transport enum. CI green, MERGEABLE/CLEAN, and the two most-affected spec files pass locally (56/56). 👍

PR #63, opened 2026-08-23
feat(clean): AIoT read surface — WorkStatus sub-states, dock activity, fault codes, DP catalog fix

martijnpoppen:
Replaces the five closed PRs #41–#45, rebuilt as one branch on the current #39 tip. **#44 is dropped entirely** — its one-line `isAiotVacuum` import is already on this base, brought in by the #34 merge.

Four independent read-side improvements. **No write path is added or changed anywhere in this PR.**

## 1 · WorkStatus state 5 → cleaning / paused / docked (DP 153)

State 5 is the vendor's catch-all for "off the dock or servicing mops". Reading only the state field collapsed three physically distinct situations into `"cleaning"` — actively cleaning, user-paused, and parked on the dock running a wash cycle.

| sub-message | activity |
|---|---|
| `go_wash.mode` = `WASHING(1)` / `DRYING(2)` | `"docked"` |
| `go_wash.mode` = `NAVIGATION(0)` | `"cleaning"` — driving to the dock, not on it |
| `station.washing_drying_system` present | `"docked"` |
| `cleaning.state` = `PAUSED(1)`, no `go_wash` | `"paused"` |
| nothing claims it | `"cleaning"`, unchanged |

Dock beats pause: a robot that paused itself to go wash reports both.

Also drops `15 → "paused"`. `WorkStatus.State` ends at `8`, so no device can report it; the entry was unreachable and its own comment said as much.

## 2 · Dock activity (DP 173)

`dockState` comes off `unexposed` and answers a typed `DockActivity`: `idle · washing · drying · descaling · emptyingDust · addingWater · recyclingWater · makingDisinfectant · cuttingHair · unknown`.

A busy subsystem wins over the mop state — they're independent, and `IDLE` is the proto3 zero member so a dock emptying its bin sends no `state` field at all. Reading `state` first would answer `"idle"` for a dock that is visibly working.

Corrects the old doc that described field #1 as "repeated `{id,state}` dock component groups" — it is the auto-maintenance **config**, a different message with the same outward shape.

## 3 · Fault codes (DP 177)

The AIoT line had no fault surface. `errorCode` now points at DP 177 (`ErrorCode` protobuf) with DP 106 as a read alias, discriminating on value shape the way `decodeCleanType` already does. `error[0]`, else `warn[0]`, else `0`.

Both lists are `repeated uint32`, which proto3 **packs** by default — handled both ways, and read as full varints, since the station (6010–6311) and situational (7000–7055) ranges are four digits.

## 4 · DP catalog reads the fields the response declares

`parseDpCatalog` keyed off `entry.type` / `entry.values`. The response declares neither — it's `data_type` and `property`. So `type === "enum"` was never true and `enumRanges` came back **empty for every real response**, which is why `suction.supportedLevels` silently resolved to `undefined`. Also drops the speculative `dp_list` / `id` fallbacks.

## Evidence

Field numbers come from the vendor's own message definitions, not a third-party client's behaviour. What corroborates them is this repo's own live traffic:

- **WorkStatus** — the three states observed on a T2351 (`CHARGING(3)`, `CLEANING(5)`, `GO_HOME(7)`) sit at exactly those positions in the same enumeration, with `state` at field 2 as the existing decoder already assumed.
- **StationResponse** — the member's own doc recorded *"field #5.1 = fill level %"* from a live payload. Field 5 is `clean_water`, typed `Numerical`, whose one field is `value`. That's what a percentage at 5.1 predicts, from a reading taken before the definition was consulted.
- **DP identities are mega-sourced.** 173 = `baseStation`, 177 = `faultAlert` in `CLEAN_PARAMS`, generated from `get_product_data_point`.
- **The catalog fix is self-evidencing** — `CLEAN_PARAMS` carries a `type` and `writable` per DP, which could only have come from `data_type` and `mode`: the exact fields the parser wasn't reading.

**Safety property throughout:** each decode only ever *narrows*. WorkStatus falls through to `"cleaning"`, the dock read to `undefined`, the fault read to `undefined`. A wrong field number yields today's behaviour, never something worse.

## Verified how

- [x] `npm run verify` green locally (Node 24.5.0) — 121 files, **2426 tests**

Adds `__tests__/proto-bytes.ts`: byte-real protobuf fixtures and a schema-less reader mirroring the `RawDpCodec` contract without importing `transport/`. A flat fake cannot exercise these decodes at all — they turn on which sub-message is present and on proto3 omitting zero-valued fields, so "the enum's zero member" and "nothing was said" are the same bytes.

**Combining the branches surfaced a real conflict worth noting:** #41 and #43 had each grown their own copy of that fixture builder, and together they collided (`TS2440`, four symbols). The last commit collapses the inline copy onto the shared helper — 80 lines removed, no fixture semantics changed. This is exactly the follow-up both original PR bodies flagged.

Also publishes `DockActivity` / `DOCK_ACTIVITIES` through the capabilities barrel — `dockState`'s declared return has to be nameable, and the docs workflow runs typedoc with warnings as errors. ESM surface 206 → 208 exports.

## Unverified or deferred

- **Nothing here is driven on hardware.** All four are schema-grounded and corroborated as above, and all degrade safely, which is why they ship as reads rather than behind markers. A capture of a real pause, a mop-wash cycle, and one DP 177 payload would upgrade them from corroborated to confirmed.
- **Only one property per DP.** `clean_level`, `dirty_level` and the `clean_water` percentage decode the same way from DP 173, but `property-id-integrity.spec.ts` allows one property per param per line. Surfacing them needs an `actions()` addition or a schema change — its own decision.
- **Only the first fault code is answered.** A robot reporting several simultaneous faults exposes only the first; the property is one number.
- **The fault code's meaning is not interpreted.** The vendor table runs to several hundred entries with per-model variants; mapping number → text is presentation and belongs to the caller.
- **`errorCode`'s published `paramType` changes 106 → 177**, with 106 moving to `readAliases`. Tuya behaviour is unchanged, but it is schema-visible and the part I'd most want a reviewer's eye on.
- `FAST_MAPPING(4)`, `REMOTE_CTRL(6)` and `CRUISIING(8)` still collapse onto `"cleaning"` — a limit of the `VacuumActivity` union, not of the decode.
- The catalog parser reads past `mode` and `code` without using them. `mode` would be real evidence for a write's verification status; wiring it changes `DpCatalog`'s shape and call sites.

---


src/model/capabilities/vacuum-clean.ts
```diff
@@ -347,22 +432,93 @@ export function decodeCleanType(
...
+ */
+function firstRepeatedCode(fields: readonly RawDpField[], field: number): number | undefined {
+  const found = fields.find((f) => f.field === field);
+  if (found === undefined) return undefined;
+  if (found.kind === "int") return Number(found.value);
+
+  let value = 0;
+  let shift = 0;
+  for (const byte of found.value) {
+    value |= (byte & 0x7f) << shift;
```
max246, line 485: Non-blocking: the manual varint accumulate uses JS 32-bit `<<`, so a packed code above 2³¹ would wrap. Fine for the documented 4-digit station (6010–6311) and situational (7000–7055) ranges — worth a one-line note that this assumes codes stay ≤ ~9 digits, so the assumption is explicit if the fault table ever grows a wider code.

src/model/capabilities/vacuum-clean.ts
```diff
@@ -246,31 +249,113 @@ export function decodeTuyaWorkStatus(raw: ParamValue | undefined): VacuumActivit
...
+  const sub = (field: number): Buffer | undefined => {
+    const found = fields.find((f) => f.field === field);
+    return found?.kind === "bytes" ? found.value : undefined;
+  };
+
+  const goWash = sub(WORK_STATUS_FIELD.GO_WASH);
+  if (goWash && GO_WASH_ON_DOCK.has(subValue(codec, goWash, SUB_MODE_FIELD))) return "docked";
+
+  const station = sub(WORK_STATUS_FIELD.STATION);
+  if (station && codec.nested(station)?.some((f) => f.field === STATION_WASHING_DRYING)) return "docked";
```
max246, line 364: This is the softest of the sub-message reads: it treats *presence* of field 3 (`washing_drying_system`) as "washing/drying", ignoring its value — whereas `go_wash` reads the actual mode. That's consistent with the stated "absent message = idle" rule and degrades safely (falls through to `"cleaning"`), so no change needed — but it's the one branch not corroborated by a live capture yet, so it's the first thing to confirm on-device.

src/model/capabilities/vacuum-clean.ts
```diff
@@ -488,15 +646,26 @@ export const VACUUM_CLEAN_MEMBERS = {
...
+   * maps the number itself.
    */
   errorCode: {
-    param: LEGACY_VACUUM_DP.ERROR_CODE,
+    param: VACUUM_DP.FAULT_ALERT,
     type: "number",
-    kind: "scalar",
     provenance: "mega",
-    description: "Error code, 0 = ok (DP 106, Int ro). Legacy Tuya G-series/X8 clean line.",
+    readAliases: [{ paramType: LEGACY_VACUUM_DP.ERROR_CODE }],
```
max246, line 799: Minor consistency: sibling aliases (`battery` line 444, `cleanType` line 462) gate their DP-106-era reads on `available: isTuyaVacuum`, but this errorCode alias is ungated. I think ungated is actually *better* here — it lets the legacy fault surface regardless of family classification, and `decodeVacuumFault` discriminates on shape (int vs protobuf) so an AIoT device that happens to carry DP 106 still decodes sanely. Just calling out the asymmetry so it reads as deliberate rather than an oversight — a word in the JSDoc would settle it.

max246 (commented):
Reviewed the full decode path against the message layouts and the tests. This is excellent — clean read-only surface, every decode only ever *narrows*, and the fixtures are byte-real (a schema-less `byteCodec` mirroring the real `RawDpCodec` contract rather than a flat-list fake), so absent-field-vs-zero-member is actually exercised — which is the whole subtlety here. I verified the four claims hold:

- **WorkStatus 5 sub-states** — resolution order (dock-beats-pause, NAVIGATION≠arrival, empty sub-message = DOING) matches the doc and only state `5` is refined; `4/6/8` correctly fall straight through to `"cleaning"`. Dropping the unreachable `15` is right (enum ends at 8).
- **Dock activity** — busy-subsystem-wins ordering + proto3-zero `state` absence → `"idle"`, and `undefined` (unbound/no-status) kept distinct from `"idle"` and `"unknown"`. Field #2-not-#1 (status vs auto-maint config) is guarded and tested.
- **Fault codes** — `firstRepeatedCode` handles both packed (length-delimited varint run) and unpacked; `error ?? warn ?? 0`, and `undefined` (didn't decode) ≠ `0` (stated no fault). Multi-byte code path tested against the 4-digit ranges.
- **DP catalog** — reading `data_type`/`property` instead of the never-present `type`/`values` is the real fix; `enumRanges` was empty for every live response before. Dropping the speculative `dp_list`/`id` fallbacks is correct now that the shape is known.

CI green, mergeable, no write path touched. Three non-blocking observations inline — all latent/consistency, none block merge.

max246 (approved):

src/model/capabilities/vacuum-clean.ts
martijnpoppen: Noted in the JSDoc in 293f676 — the width assumption is now stated outright, with the four documented ranges as the reason it holds today.

---

src/model/capabilities/vacuum-clean.ts
martijnpoppen: Agreed, no behaviour change. 293f676 records your point in `resolveCleaningState`'s doc: the station branch is marked as the softest read and the first thing to confirm on-device, with the contrast against `go_wash` (which reads the actual mode) spelled out so the asymmetry doesn't read as an oversight later.

---

src/model/capabilities/vacuum-clean.ts
martijnpoppen: Your reading is the intended one, and 293f676 says so on the member: a fault is the one value worth surfacing even when family classification is wrong or absent, and `decodeVacuumFault` discriminates on shape rather than family, so a device carrying DP 106 decodes sanely whichever line it turns out to be on. "The asymmetry is the point, not an oversight" is now in the JSDoc.

---

PR #66, opened 2026-08-23
feat(capabilities): let several members read one DP's payload

martijnpoppen:
## What changed

A new optional `readsFrom` on `ValueMember`. A member that declares it names the member which owns the `param`, decodes **that** member's stored property, and inherits its evidence gate. It publishes no `PropertySpec` of its own.

- `members.ts` — the field, and ~4 lines in `bindMembers` resolving the owner before computing `prop` and the read gate. A `readsFrom` naming nothing falls back to the member itself, whose gate is false, so a typo costs the getter rather than throwing at bind time.
- `property-id-integrity.spec.ts` — a new guard: a derived member must name a real owner in the same table that declares a `param` and is not itself derived, must declare a `decode`, and must claim no `param`, `readAliases` or `write` of its own.
- `vacuum-clean.ts` — `doNotDisturbActive`, the first reading derived this way, plus `decodeDoNotDisturbActive`.
- `manifest.spec.ts` — `tableReads` re-derives the install rule independently, so it learns the same owner resolution.
- `__tests__/bind.ts` — the helper can now inject a `RawDpCodec`, which it previously hardcoded to `undefined`.

No change to the one-owner rule itself: two *capabilities* claiming one id stays a hard failure, and 1207 stays tracked debt.

## Why

A clean-line DP carries a whole protobuf message, and most of them state more than one thing. DP 157 reports both whether do-not-disturb is switched **on** and whether its quiet window is open **right now**; DP 167 carries the session total beside two lifetime accumulators; DP 176 carries fifteen toggles. Three separate comments in this file already say the extra readings were dropped for one reason — "the property is one number", "because the property schema allows one property per data point".

The obvious fix was to relax that rule and let two members claim one DP. That does not work, and fails silently:

- `Device.resolveInto` builds `specByParam` first-writer-wins (`src/model/device.ts:221`).
- `applyParams` stores each incoming param under exactly one property name (`device.ts:417`).
- A getter reads by *property name* (`members.ts:794`).

So a second spec for DP 157 would be dropped on the floor and its getter would read a name nothing is ever stored under — `undefined` forever, with the one-owner guard green. The guard was never the only thing enforcing one-DP-one-property; the storage layer is.

`readsFrom` is the shape that matches what is actually true: one property, several readings of its payload. It is the inverse of `readAliases` (one value across several wire ids), and it leaves the schema literally correct, since the schema describes what a device *reports* and the device reports one DP.

## Evidence

`UNDISTURBED_FIELD` in this file already recorded `active` as field 1 of `UndisturbedResponse`, beside `undisturbed`(2) — that comment is not new here, only the number's promotion into a constant. The existing `decodeDoNotDisturb` fixture already encodes `active` as a `Switch`-wrapped field 1.

**Stated ambiguity:** whether the vendor wraps `active` in a `Switch` the way `sw` is wrapped, or sends it as a bare bool, is not confirmed. Both shapes are read — a varint at face value, a sub-message opened for its `value`. That is not a guess about which arrives; both readings mean the same flag, so handling either is what removes the guess. Called out in the decoder's own doc.

An absent `active` beside a present `undisturbed` reads `false` (proto3 omits the zero, so "not open" and "said nothing" are the same bytes once the message is recognisable); an absent `undisturbed` is `undefined`. Same rule the child-lock and session-time decodes already use.

AIoT only. The Tuya line's DP 107 is a plain bool carrying the **switch** and states nothing about the window, so a non-protobuf value answers `undefined` rather than borrowing the switch — gated on `paramIds.has(157)` so the owner's DP 107 read alias cannot drag the getter onto a device that cannot answer it.

## Verified how

- [x] `npm run verify` is green locally — 122 files, 2477 tests
- [x] Added or updated specs (offline, synthetic fixtures) — 5 decoder cases and 4 bind-level cases, all on `proto-bytes.ts` byte-real fixtures
- [x] Docs updated — `cd docs && npm run docs:api` clean (typedoc warnings-as-errors is not part of `verify`)

The new guard was checked by breaking it both ways: `readsFrom` naming a non-member, and `readsFrom` beside its own `param`. Both fire, and the second is caught by the existing one-owner guard as well.

## Unverified or deferred

- **No writes, by construction.** A `readsFrom` member cannot declare one, and the guard enforces it. Setting a field inside a shared payload means re-encoding the whole message, which needs an encoder and a captured write this SDK has neither of — the owner keeps the wire.
- **The `active` wire shape** is read defensively rather than confirmed, as above.
- **DP 168 consumables and the rest of DP 154** are what motivated this and are deliberately *not* in this PR: their field numbers come from the vendor `.proto` and are not recorded anywhere in this repo, so shipping them here would be asserting numbers this branch cannot point at. They follow once re-verified.
- **DP 167's `total`(2) / `user_total`(3)** are the same story — the containers are documented, the message type inside them is not.

Stacked on `martijnpoppen/clean-record-list`.

---


max246 (approved):
Traced the whole path — resolution, storage, decoder, guard — against the diff and the code it touches. This is a clean, correct design. Approving.

**The framing is right.** The PR correctly identifies that one-DP-one-property is enforced by the *storage layer*, not just the guard: `resolveInto` is first-writer-wins on `specByParam`, `applyParams` stores each param under exactly one name, and getters read by name — so relaxing the one-owner rule would silently strand a second getter on `undefined`. `readsFrom` (several readings of one payload) is the shape that fits, and is the clean inverse of `readAliases` (one value across several ids).

**Verified:**
- **No PropertySpec, as claimed** — `propertiesOf` skips any member with `param === undefined` (members.ts:530), which a `readsFrom` member always is (guard-enforced). The invariant holds via the *pre-existing* paramless-skip; the PR didn't have to touch `propertiesOf`, and correctly didn't.
- **`bindMembers` resolution** — owner lookup with self-fallback is right: a typo'd `readsFrom` resolves to the member itself, which declares no param, so its gate is false → the getter silently drops (fail-soft at runtime, fail-hard in `property-id-integrity.spec`). The gate uses the *owner's* evidence while `available` stays the *member's* own — which is exactly what makes the AIoT-only narrowing (`paramIds.has(157)`) keep `doNotDisturbActive` off a Tuya-only device whose `doNotDisturb` reports via the DP 107 alias.
- **Decoder** — no field-number collision: `active` is field 1 of the top-level `UndisturbedResponse`, while `sw` is field 1 *inside* the field-2 `undisturbed` sub-message, so `decodeDoNotDisturbActive` (top-level f1) and `decodeDoNotDisturb` (f2 → nested f1) read cleanly separate things. Both `active` wire shapes handled (bare varint / `Switch`-wrapped `value`), and the `undefined` (not an UndisturbedResponse) vs `false` (present-but-zero) distinction matches the sibling decodes.
- **Read-only by construction** — enforced by the guard (no `param`/`readAliases`/`write`) and structurally (setter block needs `m.write`). Chained `readsFrom` is disallowed (owner must not itself be derived), so cycles are structurally impossible.
- **Guard coverage** — checks owner exists, owns a param, isn't itself derived, member has a decode, and claims no param/alias/write. Comprehensive.

Ran `property-id-integrity`, `vacuum-clean`, `manifest` specs locally — 96/96 pass; CI green. The `active` wire-shape ambiguity is read defensively and already flagged in the decoder doc + PR checklist as needing a live capture, which is the right call. This also lays a clean foundation for the deferred DP 168 consumables / DP 154 / DP 167 readings once their field numbers are re-verified. 👍

PR #69, opened 2026-08-23
feat(clean): read a robot vacuum's cleaning history

martijnpoppen:
## What changed

`eufy.getCleanRecords(sn, pageSize, page)` returns one page of completed runs, with the account's total so a caller can page.

Split the way `dp-catalog` already is: `MegaHttpClient.getCleanRecords` does the signed POST and hands back the raw response; `model/clean-records.ts` owns the shape. `transport/` never learns what a record means and `model/` never learns how the call is signed.

## Why

The cloud keeps a record per completed run and nothing in the SDK reached it. This is the list; the per-run detail blob is a separate problem (below).

## Evidence

Endpoint, body and response shape come from the v6 app's own definitions for this call: `POST /app/clean/get_device_clean_record_list` with `{device_sn, num, page}`, answering `{clean_record_list[], total}`, each record carrying `{id, device_sn, user_id, create_time, update_time, file_path, download_url, extend}`.

Three deliberate omissions:

- **The detail blob is not fetched.** Each record carries a `download_url` to the run's map and statistics. The host it points at is unconfirmed, the blob is a framed container rather than a bare protobuf, and this client's binary path is host-allowlisted with SSRF checks by design. The URL is handed over instead of guessed at.
- **`extend` is passed through verbatim.** Its contents are not specified anywhere this SDK can point at, so parsing it would be a wire claim with no evidence behind it.
- **`user_id` and `device_sn` are not surfaced.** The caller supplied the serial, and the account id is not this type's business. A spec pins the exact key set so neither can reappear by accident.

Deletion (`POST /app/clean/delete_device_clean_record`) is left out entirely: it is destructive, and nothing here needs it to be useful.

## Verified how

- [x] `npm run verify` is green locally — 2464 tests
- [x] Added or updated specs (offline, synthetic fixtures) — 9 new tests; fixtures use synthetic serials and a zeroed account id

Parser behaviour, all pinned: a record with no usable `id` is skipped rather than admitted with a fabricated one — the id names the run, so a record a caller cannot act on is worse than an absent one. Every other field defaults, so one odd entry never costs the whole page. Numeric strings are accepted because the gateway is inconsistent about them. An empty list answers a real page rather than the failure sentinel — "no history yet" is not "could not read".

## Unverified or deferred

- **The response shape is app-sourced, not observed on a live call.** The parser returns the empty sentinel on any mismatch and never throws, so a wrong guess degrades to "no history" rather than to an exception.
- **The detail blob decode** is the follow-up, and needs the download host confirmed first.

## Review follow-up (second commit)

Both of max246's non-blocking notes, answered in the docs rather than the behaviour:

- The JSDoc said records come back "newest first" as though the SDK guaranteed it. The parser preserves the response order and never sorts, so that ordering is the **gateway's** contract — now stated that way.
- `total` falling back to the page's own record count can understate when a malformed record was skipped. Recorded on the field, with why that is the safe direction: a caller pages until it reaches the total, so a low figure ends the walk early rather than looping for records that will never arrive.

Rebased onto `claude/clean-read-surface` (#63) now that #71 has merged into it — the previous base, `martijnpoppen/clean-dock-info-childlock`, became an ancestor of #63 with no PR of its own. Same diff, one useful destination.


max246 (approved):
Read the parser, the transport call and the client wrapper end to end. Clean, well-scoped, correctly layered. Approving.

**Layering is right** — same split as `dp-catalog`: `MegaHttpClient.getCleanRecords` does the signed POST and returns raw; `model/clean-records.ts` owns the shape; neither learns the other's job. `post("clean", …)` resolves to `app-clean-${region}.eufy.com` by the same interpolation every sibling call uses (`things`/`devicemanage`/`push`), so the routing is structurally correct — whether that endpoint answers is the app-sourced unknown you already flag, and the parser degrades to the empty sentinel on any mismatch rather than throwing.

**Parser is defensive in the right ways** — id-less records skipped (a run a caller can't name is worse than an absent one), every other field defaults so one bad entry never sinks the page, numeric-string tolerance for the gateway's inconsistency, `EMPTY_CLEAN_RECORD_PAGE` on shape mismatch, never throws. `num()` rejecting non-`^\d+$` also correctly excludes negatives for these unix-seconds/id fields. Empty list → a real page, not the failure sentinel — the right distinction.

**The deliberate omissions are all well-justified**: the `download_url` blob handed over rather than fetched (unconfirmed host + your SSRF-allowlisted binary path — exactly right not to guess), `extend` passed through verbatim, `user_id`/`device_sn` withheld with a spec pinning the key set, and destructive delete left out. The client wrapper's try/catch → EMPTY + debug log keeps it total.

`clean-records.spec` passes locally (9/9), CI green.

**One minor note (non-blocking):** both the client JSDoc and PR say records come back *"newest first,"* but the parser preserves the response order and doesn't sort — so that ordering is the gateway's contract, not something the SDK enforces. Fine as-is (the app relies on the same server ordering), but a caller reading the JSDoc might assume the SDK guarantees it; a word like "newest first *as the cloud returns them*" would set expectations precisely. Also, `total` falling back to `records.length` can understate when a malformed record was skipped — a genuine edge, and the safer direction to err, so just noting it.

PR #71, opened 2026-08-23
feat(clean): four more AIoT clean-line reads (DP 157, 167, 169, 176)

martijnpoppen:
Combines the two PRs that were #67 and #68 — the same kind of work, in the same files, on the same evidence, so one review rather than two. Same commits, unchanged.

## What changed

Four values a T2351 reports and the SDK could not read. Two of them existed as Tuya-only properties and gain their AIoT wire on the **same property name**, rather than a second spelling per platform; two are new.

| Property | Wire | Notes |
|---|---|---|
| `vacuum_clean.doNotDisturb` | DP 157 `UndisturbedResponse` | DP 107 kept as a Tuya read alias |
| `vacuum_clean.clearTime` | DP 167 `CleanStatistics` | DP 109 kept as a Tuya read alias |
| `vacuum_dock.dockFirmwareVersion` | DP 169 `DeviceInfo.station.software` | new |
| `vacuum_clean.childLock` | DP 176 `UnisettingResponse.children_lock` | new |

The two aliased decoders discriminate on the value's **shape** — plain bool / integer for Tuya, protobuf for AIoT — the way `decodeCleanType` and `decodeVacuumFault` already span the two lines on one property.

`vacuum_dock.decodeState` now picks up DP 169 alongside 173; detection still keys on 173 alone, so the capability still means "has a dock".

## Why

A T2351 reported nothing for do-not-disturb or session clean time, and the dock had no version surface at all. One name per feature is what keeps a host from branching on device family to ask the same question.

`dockFirmwareVersion` is deliberately on the **dock** capability and deliberately not called a plain firmware version: `info.firmwareVersion` already reports the **robot's**, off the cloud device record, and a second spelling of that would be one feature under two names. The dock is separate hardware.

## Evidence

All four ids are mega-sourced, from a real `get_product_data_point` response resolved into `CLEAN_PARAMS`: 157 `doNotDisturb`, 167 `cleanStats`, 169 `appAndDevice`, 176 `commonSettings`. Field numbers come from the vendor's own message definitions.

Four decode subtleties, each pinned by a spec:

- **`Undisturbed.sw` is read, not `UndisturbedResponse.active`.** The latter says the quiet window is open *right now*; this property means the feature is switched *on*. A payload with `active=true` beside an empty switch reads `false`.
- **`CleanStatistics` carries three containers with a `clean_duration` at the same inner field number** — `single`(1), `total`(2), `user_total`(3). Only `single` is read; picking the wrong one would report a lifetime figure as the current run.
- **DP 169 also carries the robot's own `software` at the top level.** A spec pins that the decode reads the **nested** station one.
- **DP 176's request counterpart numbers the same settings differently** — only `children_lock` sits at 1 in both — so a reader and a writer of that DP can never share a field table. Recorded in the constant's own doc, because it is the trap the next person extending this would fall into.

Present-but-empty containers read as `false` / `0` rather than as missing: proto3 omits zero values, so "off" and "nothing said about the switch" are the same bytes once the container is there. An absent container stays `undefined` — that is the device not answering.

## Verified how

- [x] `npm run verify` is green locally — 2455 tests at the tip; each commit was verified on its own first
- [x] Added or updated specs (offline, synthetic fixtures) — 22 new tests on byte-real `proto-bytes.ts` fixtures

Members keep their stored `kind` alongside the new `decodedKind`; `value-kinds.spec` pairs `unit` / bool `type` against the stored one.

## Unverified or deferred

- **Reads only.** No captured write for any of these DPs, so no setters are installed. Unlike a Tuya DP write, an AIoT one is not refused by a router guard — it reaches the device — so a setter stays off the surface until a capture exists.
- **`dockFirmwareVersion` is `undefined` for a robot out on a job.** The vendor fills the station block only while the robot is powered on **at** the dock, so that is the normal answer rather than a fault. Documented on the member.
- The lifetime accumulators beside `single`, and the other fourteen DP 176 toggles, are left unread: their field numbers are not recorded in this repo.

Two commits, reviewable separately. Stacked on `claude/clean-read-surface` (#63).

---


max246 (approved):
Traced all four decoders against their documented message shapes. Clean, consistent with the rest of this stack (#63/#66 build on it), no defects. Approving.

**Decoders — each reads the right field and degrades the right way:**
- **`doNotDisturb` (DP 157)** — reads `undisturbed(2) → sw(1) → value(1)`, i.e. the *switch*, not top-level `active(1)`. That's the exact separation #66's `decodeDoNotDisturbActive` depends on (top-level f1) — no collision, verified across both PRs. Tuya bool/`0`/`1`/`true`/`false` short-circuits before the protobuf path.
- **`clearTime` (DP 167)** — reads `single(1).clean_duration(1)` only, deliberately not the `total(2)`/`user_total(3)` accumulators that carry a `clean_duration` at the *same* inner field number. This was the sharpest trap and it's handled correctly.
- **`childLock` (DP 176)** — reads `children_lock(1) → value(1)`; the doc records that the REQUEST message numbers the same toggles differently, which is the right warning to leave.
- **`dockFirmwareVersion` (DP 169)** — reads the *nested* `station(11).software(1)`, not the robot's top-level software — the documented DP 169 trap, avoided.

**The undefined-vs-false/0 discipline is uniform**: an absent container is `undefined` (device not answering); a present-but-empty one is `false`/`0` (proto3 omits the zero). Correct at every nesting level, and none of them throw.

**Capability wiring is right:**
- Dock `detection` still keys on `[VACUUM_DOCK_DP]` (173) alone — I checked it's unchanged — so landing DP 169 into `pickDpParams` doesn't let a 169-only device masquerade as a dock. `dockFirmwareVersion` is evidence-gated on 169 via `reads()`, so the getter only appears once the device has reported it, and reads `undefined` while the robot is off the dock — matching the documented normal case.
- `childLock` gated to DP 176 (AIoT-only, no Tuya equivalent, no alias); `doNotDisturb`/`clearTime` correctly span both lines with the Tuya id as a `isTuyaVacuum`-gated read alias and shape-discriminated decoders — same pattern as `decodeCleanType`/`decodeVacuumFault`.
- Reads only; no setter installed for any of these, correctly flagged since an AIoT write reaches the device (no router guard) and none is captured.

`vacuum-clean` + `vacuum-dock` specs pass locally (86/86), CI green. Naming `dockFirmwareVersion` on the dock capability rather than shadowing `info.firmwareVersion` is the right call. Nothing to change. 👍

PR #72, opened 2026-08-23
fix(p2p): give the level-2 grace and the prebuffer window one owner each

### LINKED ISSUE #64: recordFragments waits ~8 s before attaching to an already-warm shared source, silently
## What happens

`recordFragments` spends a fixed ~8 s between the call and its first fragment, on an **already-warm** shared source as well as a cold one, and emits no log line in that window. `live()` pays the same ~8 s on the first call.

Measured with the SDK logger at debug on a wired T8442 behind a T8010 station, redacted:

```
+1.4s   [mega] key/exchange eufylife (authed) ->
+9.330s [live sn*0237:0] warming (retry=2000ms deadline=20000ms)
+9.437s [live sn*0237:0] first frame — live (640x360 h264, powered=wired)
        warmed: 224 video frames
+24.339s --- requesting a fragment recording ---
+32.375s fragment 0 at 8.035s init=1137 data=0 keyframe=false
+32.375s fragment 1 at 8.035s init=0 data=379546 keyframe=true
```

The live source is streaming and has been for 15 s; its session is resolved and 224 frames have already arrived. `recordFragments` is still 8.035 s from its first fragment, and the gap is silent — the next log line after the request is the fragment itself.

## Why it looks like `resolveSession`

Both `sharedLiveSourceFor` and `recordFragments` reach the transport through `resolveSession(sn, { waitLevel2: "soft" })`. `sharedLiveSourceFor` returns the cached source immediately once it exists, so the 8 s cannot be source construction or warm-up. `~8 s` is also close enough to a fixed soft-wait timeout to be suspicious, and on this account only one station ever logs `level-2 key negotiated` — the one the camera is *not* behind:

```
+1.447s [p2p] sn*2199 level-2 key negotiated (cipher_id 209)
```

So the soft wait looks like it is running to its timeout every time rather than being satisfied.

## Why it matters

It is the dominant latency in a HomeKit Secure Video recording, and it is paid on every recording. HomeKit gives an accessory ten seconds to produce first output; the plugin measured 10.2 s and 10.7 s on two wired cameras, and roughly eight of those are this wait. Nothing on the host side can shorten it, and pre-event media does not help either — draining a retained window does not move the moment the recording attaches.

Verified independently: on a source already warm and already carrying a full four-second prebuffer ring, first fragment still arrived 8.0-8.2 s after the request across four runs on two cameras (T8442, T8400).

## What would fix it

Either satisfy the soft level-2 wait from an already-resolved session instead of re-waiting, or skip it entirely when the shared source for that `${parentSn}:${channel}` is already `live`. A caller that is joining a streaming source has, by construction, the session it is waiting for.

## Environment

`@mega-yfue/eufy-sdk` at `6bbb4eb`, Node 24.19.0, one T8010 station with T8442/T8400/T8410 wired cameras. Consumer is `@homebridge-plugins/homebridge-eufy-security` `beta-5.0.0`, whose HKSV work is homebridge-plugins/homebridge-eufy-security#1000.

### LINKED ISSUE #65: The pre-event window is unbounded, collapses when the stream stalls, and only live() can configure it
Three separate defects in the same window, all in `SharedLiveSource` / `MediaProvider`. Splitting them is fine if you prefer; they are filed together because a host cannot work around any of them independently.

## 1. `bufferedMedia`'s keyframe fallback ignores the requested window

`shared-live-source.ts`:

```ts
private bufferedMedia(seconds: number): TimedMediaFrame[] {
  if (this.preBufferMs <= 0 || !this.ring.length) return [];
  const cutoff = Date.now() - Math.min(seconds * 1000, this.preBufferMs);
  let start = this.ring.findIndex((item) => item.timestampMs >= cutoff && this.isKeyframe(item));
  if (start < 0) start = this.ring.findIndex((item) => this.isKeyframe(item));
  return start < 0 ? [] : this.ring.slice(start);
}
```

When no keyframe sits at or after the cutoff, the fallback returns the **whole ring** — the requested `seconds` is discarded rather than honoured as a bound. Consequences:

- `attachWithPrebuffer(0)` on a source that has a ring drains everything. So `recordFragments({ preBufferSeconds: 0 })`, and omitting the option (`?? 0`), do not mean "no pre-event media" — they mean "all of it".
- `recordFragments({ preBufferSeconds: 2 })` on a source built with `preBufferSeconds: 4` can deliver 4 s.

A host therefore has no way to ask for *less* than the configured window, only to decline configuring one. Confirmed hermetically: `attachWithPrebuffer(0)` returns `[]` when `preBufferMs === 0`, and the whole ring otherwise.

Suggested fix: keep the alignment intent but bound it — fall back to the *newest* keyframe at or before the cutoff rather than the oldest keyframe in the ring, and return nothing when `seconds <= 0`.

## 2. `pushRing`'s alignment trim collapses the ring when the stream stalls

```ts
const cutoff = item.timestampMs - this.preBufferMs;
let firstKeep = 0;
for (let i = 0; i < this.ring.length; i++) {
  if (this.ring[i].timestampMs >= cutoff && this.isKeyframe(this.ring[i])) { firstKeep = i; break; }
  if (this.ring[i].timestampMs >= cutoff) { firstKeep = i; }
}
let align = firstKeep;
for (let i = firstKeep; i >= 0; i--) { if (this.isKeyframe(this.ring[i])) { align = i; break; } }
if (align > 0) this.ring.splice(0, align);
```

If no keyframe is inside the window, `firstKeep` ends up at the newest frame and `align` walks back to the most recent keyframe, so the ring is spliced down to that keyframe onward — near-empty. That is exactly what a stalling stream produces, and these cameras stall: on a wired T8442 one recording showed a 7.45 s gap between fragments, with 118 samples arriving in one fragment afterwards.

Effect on a host: the retained window is unpredictable. Same camera, same 15 s of warm-up, same `preBufferSeconds: 4`, three runs — 3.38 s of pre-event media, then ~0, then ~0. On a T8400 whose stream is steady it was consistently ~3.4-3.7 s across runs.

The trim wants to guarantee the run opens decodable, which is right; dropping most of the window to achieve it is not. Retaining the newest keyframe *and everything after it* even when that keyframe predates the cutoff would keep the window decodable without emptying it.

## 3. No pre-event window on `snapshotLive`, `openReadable` or `talkback`

`command-router.ts` `mediaProviderFor`:

```ts
snapshotLive: async (opts) => { const source = await this.sharedLiveSourceFor(sn, { powered: opts?.powered }); … }
live:         async (opts) => { const source = await this.sharedLiveSourceFor(sn, opts as SharedLiveOpts); … }
openReadable: async (opts) => { const source = await this.sharedLiveSourceFor(sn, { powered: opts?.powered }); … }
```

Only `live` forwards `preBufferSeconds`. Since the window is fixed at source construction and `warnIgnoredLiveOpts` refuses to change it later, whichever egress opens the source decides whether a window exists at all — and three of the four cannot ask for one.

This is not theoretical. A host that serves HomeKit snapshots calls `snapshotLive` constantly; each call opens a ringless source that then lingers 8 s. A recording triggered inside that window gets no pre-event media even though the host asked for it on every call it can.

`MediaProvider.snapshotLive`, `openReadable` and `talkback` should accept `preBufferSeconds` and pass it through, exactly as `recordFragments` does.

## Evidence that the mechanism itself is right

None of this is a claim that the design is wrong — driven directly it works well:

```
SharedLiveSource({ preBufferSeconds: 4 }) + 6 s of synthetic 15 fps frames
  ringBuffer(4).length === 60
  attachWithPrebuffer(4) -> 60 items spanning 3.91 s, first is a keyframe
```

and end to end on a wired T8400, a recording received at least 1.44 s and 1.68 s of media captured before it attached across two runs, every drained fragment opening on a keyframe, versus 0.00 s on the same camera warmed with no window.

## Environment

`@mega-yfue/eufy-sdk` at `6bbb4eb`, Node 24.19.0. Consumer is `@homebridge-plugins/homebridge-eufy-security` `beta-5.0.0` (homebridge-plugins/homebridge-eufy-security#1000). Related: mega-yfue/eufy-sdk#64.

lenoxys:
Fixes #64 and #65. They are one PR because each defect came from putting a window on the caller instead of the object that owns it: the level-2 grace belongs to the session, and the pre-event window belongs to the shared pull.

## Level-2 Grace (#64)

`P2PSession.awaitLevel2Key(graceMs, graceFrom)` replaces three polling loops and owns all waiters, deadlines, terminal outcomes, and diagnostics.

- Best-effort media uses one session grace measured from connect. A late `CMD_GATEWAYINFO` does not restart it, so an already-warm source cannot charge another eight seconds.
- A command that cannot be framed without the key retains its full per-call grace. A delayed gateway reply on an old session can still satisfy that command.
- A key already present and a one-shot negotiation already concluded without one both answer immediately.
- Close settles every waiter with a distinct result and clears its deadline.
- Connection generations own their key and async cipher lookup. Close clears the key and sequence; fulfilled and rejected lookups from an earlier connection cannot mutate a reconnected session; an old waiter cannot observe the new connection's key as its own success.
- Waiting, timeout, terminal no-key, and close are distinct debug diagnostics.

**Live qualification:** on an already-warm wired source, the first source fragment arrived 18 ms after `recordFragments` and the adapted initialization segment followed in 266 ms. A snapshot-first source attached the immediately following recording in 8 ms. The fixed roughly eight-second per-recording delay is gone.

## Pre-event Window (#65)

Retention and drain now share one keyframe rule: start at the newest retained keyframe at or before the cutoff, or at the oldest retained keyframe when the ring reaches no further back.

- A drain covers the requested window instead of beginning inside it.
- A shorter drain cannot fall back to the whole ring.
- Zero, absent, negative, `NaN`, and infinite durations retain and drain nothing.
- A delivery stall or long GOP keeps its complete decodable run until a newer keyframe gives the time-based trim somewhere safe to move to. There is no independent frame-count cap overriding the configured window.
- `SharedSourceHints` exposes `powered` and `preBufferSeconds` on every source-creating egress: `snapshotLive`, `live`, `openReadable`, `recordFragments`, and `talkback`. Each forwards its full option bag to shared-source construction, while `SHARED_LIVE_OPT_KEYS` keeps egress-only options out of conflict diagnostics.

Retention uses transport-arrival time because frames carry no device clock. A drain therefore exceeds its requested duration by one keyframe interval on a steady stream and potentially more after a delivery stall; this is documented rather than presented as a strict duration guarantee.

**Live qualification:** a four-second window on an already-warm wired source delivered at least 4.53 s of pre-event media, every drained fragment keyframe-aligned. When a live snapshot created the source first, the following recording drained at least 1.21 s, again entirely keyframe-aligned.

## API

- New `SharedSourceHints` in `core/contracts`, carrying `powered` and `preBufferSeconds`.
- `MediaProvider.snapshotLive`, `live`, `openReadable`, `recordFragments`, and `talkback` accept those hints.
- New `P2PSession.awaitLevel2Key(graceMs, graceFrom)`; `hasLevel2Key` remains the synchronous state read.
- `SharedLiveOpts` extends `SharedSourceHints`.

Pre-1.0, so no compatibility shim.

## Verification

- `npm run verify`: format, typecheck, four architecture guards, build, package seal, ESM import, examples typecheck, and 2,576 tests.
- Focused tests cover invalid durations, newest-at/before keyframe selection, long GOPs and stalls, every media egress, concurrent waiters, terminal/timeout/close diagnostics, deadline cleanup, delayed gateway replies, and fulfilled/rejected stale lookups across reconnect.
- Two-axis Standards and Spec review: no findings after corrections.


max246 (approved):
This is the careful kind of fix — both defects are "the window lived on the caller, give it to the object that owns it." I read the two hard cores line by line rather than trusting the tests, and they hold up. Approving.

**Level-2 grace (`P2PSession.awaitLevel2Key` / `settleLevel2`) — concurrency is sound:**
- Ordering is right everywhere it matters: `setLevel2Key` sets `level2Key` *before* `settleLevel2()`, so the outcome computes `"key"`; the terminal paths settle with `level2Key` still undefined → `"terminal"`.
- The async cipher lookup captures `generation` and re-checks `this.closed || generation !== this.connectionGeneration` after its one `await`; the rest of that block is synchronous (`deriveLevel2KeyFromGatewayInfo` isn't awaited), so a single guard is sufficient and a stale lookup from an old connection can't mutate a reconnected session.
- `close()` bumps the generation, clears the key/seq, and `settleLevel2("closed")` settles **every** waiter — so no waiter can straddle a reconnect and observe the new connection's key as its own success. New waits after close short-circuit on `this.closed`.
- The waiter/timeout race is clean: `finish`'s `settled` flag + `clearTimeout` + splice-by-index, and `settleLevel2` empties the array with `splice(0)` so there's no double-wake. `deadline.unref?.()` keeps it from holding the loop open.
- Router selection matches the intent: media/soft → `("session", 8s)` measured from connect (so an already-warm source computes `remaining <= 0` and returns instantly instead of re-charging 8s — this *is* the #64 fix), command-framing → `("call", 25s)` per call, and `!soft` still throws. Every old 200ms busy-poll loop is gone.

**Pre-event window (`windowStart` / `durationMs`) — correct in all cases:**
- Traced `windowStart` through: newest keyframe at-or-before cutoff (breaks only once a later keyframe exists *and* one is already recorded); oldest keyframe when the ring reaches no further back (the `&& start >= 0` guard lets the first past-cutoff keyframe be recorded when `start === -1`); and no-keyframe → `0` sentinel, disambiguated in `drain` by the explicit `isKeyframe(ring[start])` re-check. A legitimate index-0 keyframe and the sentinel don't collide.
- Retention and drain share `windowStart`, so the ring is never trimmed tighter than a drain needs. `durationMs` maps absent/NaN/Infinity/negative/zero → 0, matching "retain and drain nothing."

**API + diagnostics:** `SharedSourceHints` (`powered` + `preBufferSeconds`) is cleanly factored onto every source-creating egress with a clear "any egress may open the pull" rationale, and `SHARED_LIVE_OPT_KEYS` correctly lists only source-construction options so per-egress options (`timeoutMs`, `skipKeyframes`, `encoder`, …) don't raise false conflict diagnostics.

Focused suites pass locally (p2p-level2-grace / level2-wait-ownership / shared-live-prebuffer / shared-source-opts — 53/53), CI green (2576). Nothing to change. 👍

PR #74, opened 2026-08-24
feat(clean): the Raw-DP writer, the frames it unblocks, and a live capture that corrected one

martijnpoppen:
Six commits, reviewable separately. The first is the plan's stated prerequisite; the next two are its first uses; the last three fold in a product-catalogue dump and a live DP capture that arrived mid-review.

**The capture found a real bug in commit 3 — see "What the capture corrected" below. That is the part most worth a reviewer's attention.**

## What changed

| # | Commit | Item |
|---|---|---|
| 1 | `RawDpWriter` | §3 — the writer, `encodeModeCtrl` rewritten on it |
| 2 | Dock + mode-control frames | A2, A3 — declared, not installed |
| 3 | Detail blobs, map fetch, area-select | A11, D1, B1/B2 |
| 4 | X10 product catalogue | **D2, D3 answered** |
| 5 | Live DP capture | **consumables fix**, `resumeCleaning` verified |
| 6 | CleanParam mop settings | **A6 complete** |

## Why

Every write so far hand-rolled its bytes, which was tolerable while the only one was two varints in a flat message. It stops being tolerable one level down: a dock command wraps a `oneof` two levels deep, a room-select carries a repeated sub-message. Hand-rolling those is where silent wire bugs come from, and on a fire-and-forget write a bad frame is indistinguishable from success.

## What the catalogue settled (commit 4)

A live `get_product_data_point` dump for the **T2351 (X10 Pro Omni)**. The **T2266 (X8)** returns an empty list for the same call — it is the Tuya line and this endpoint serves the AIoT catalogue.

- **D2 answered** — schedules live on **DP 164 `timing`**, `TimerRequest` / `TimerResponse`.
- **D3 answered** — every previously-unparsed DP comes back named. The new information is the **message per DP**, recorded as `VACUUM_DP_MESSAGE`: data, not code.
- **157 → `UndisturbedResponse`, 168 → `ConsumableRuntime`** confirmed rather than resolved offline.
- **DP 162 is Raw carrying `LanguageRequest`/`LanguageResponse`** — a locale string was never the wire, which is why holding that write was right.
- **`stopRemoteControl` corrected** — it used `STOP_RC_CLEAN(16)`, assumed from the name. The DP 155 note says: enter with `START_RC_CLEAN` or any direction, leave with **`STOP_TASK`**.

## What the capture corrected (commits 5–6)

A live DP capture from a T2351 driven through the app.

**A real bug this PR introduced.** The consumables decode never worked on hardware. `ConsumableRuntime` is wrapped at field 1 of the report; the parts sit one level down. Reading them at the top level finds the wrapper where a part should be and answers `undefined` for all nine counters. **The fixtures encoded the same wrong nesting**, so they passed against a decode that never worked — the failure a synthetic fixture cannot catch alone, since it only ever confirms the assumption it was written from. Fixed, with a fixture trimmed from the real report.

**`RESUME_TASK = 14` captured** → `resumeCleaning` drops `unverified` and becomes a real installed verb, declared as a `method` like the three already-proven ones.

**`seq` is one counter across verbs.** The app sent 124/125/126/127 for start, pause, resume, go-home. The three existing verbs each kept their own `let seq = 111`, which would hand two outstanding requests the same number — the one thing `seq` exists to prevent. They now share one.

**DP 157's open ambiguity is settled** — `active` IS `Switch`-wrapped. The bare-varint branch stays; it costs one comparison.

**A6 completed.** `mop_mode` supplies `mopLevel` and `mopCornerClean`. `decodeCleanParamValue` grew an optional inner field number because `mop_mode` holds level(1) and corner_clean(2) side by side — "first varint" would report edge-hug mopping as ON purely because the water level was high. A test pins that trap.

**A near miss worth recording:** the live report carried `station(14)` with a field present *while cleaning*. Reading any station field as washing/drying would have called that "docked". The decode checks field 3 specifically, so it holds — but the reviewer who called that the softest branch was right, and there is now a test for it.

## Evidence

Field numbers are the vendor's own — cited per constant table — plus the catalogue and capture above. Three places where the honest answer was to **not** assume:

- **The detail blob's length-field width is undocumented.** `unwrapCleanRecordBlob` tries the plausible widths and keeps whichever makes the checksum *and* the declared length agree. The checksum is the oracle, so a wrong guess fails loudly instead of yielding a plausible wrong run.
- **`CleanRecordDesc.Extra`** — its field number within the parent is recorded nowhere. Left undecoded.
- **`mapId` has no default.** eufy-clean defaults it to `3`; on a two-floor home that sends the right room ids against the wrong floor's map.

**Coordinates go through ZigZag**, which is why the writer keeps `sint` as its own call. As a plain varint, −1 encodes as 2⁶⁴−1 and the robot drives somewhere real and wrong.

Two decisions the writer deliberately does **not** make for its caller: a zero-valued field is written when asked, and an empty sub-message is emitted rather than optimised away. Only the capability knows whether "explicit zero" and "absent" are the same claim.

**No real device data anywhere.** The capture carried a serial, MAC, SSID and LAN IP; none reached this diff, and every fixture is synthetic or trimmed to shape only.

## Verified how

- [x] `npm run verify` green locally — 140 files, **2870 tests**
- [x] Ran against a real device — **read-only.** A catalogue API call and a passive DP capture while a human drove the app. **No write was sent by this SDK.**
- [x] Added or updated specs (offline, synthetic fixtures) — 75 new

## Unverified or deferred

Seventeen frames still ship behind `unverified` — no setter installed, intent path refuses. `resumeCleaning` is the one verb a capture promoted.

- **`stopCleaning` remains unverified.** The app's "end clean" sent method **6 (go home)**, not `STOP_TASK(12)`.
- **All six dock verbs remain unverified.** DP 173 only ever showed `StationResponse`; the empty-dust press produced no `StationRequest` echo.
- **B1/B2 encoders** are wired to no member: none driven, and no `mapId` to drive them with.

Still out of scope for lack of recorded evidence:

- **B3 multi-map, B4 dock automation, B2's scene list** — the catalogue names their DPs and messages (171, 172, 180); the field numbers *inside* those messages are still unrecorded.
- **`fan`** is named in the field table and deliberately not surfaced — the capture shows it tracking DP 158, which `suction.level` already publishes.
- **A12 MAC/SSID/IP** — field numbers within `DeviceInfo` not recorded.
- **D1 map decode, D4** — the decoder is the clean-native library, absent from the app package.


max246 (approved):
Read the encoder, both live-capture fixes, and the record-blob unwrap against their tests. This is careful, high-value work — the capture-found bug alone justifies the PR. Approving.

**`RawDpWriter` (the foundation) is correct where it's easy to be subtly wrong:**
- `encodeVarint` uses **division, not bit-shift**, and guards `Number.isSafeInteger` — so it can't wrap above 2³¹ and emit a plausible frame for the wrong value (the exact trap a hand-rolled varint falls into). `tag()` uses `field * 8 + wire` for the same reason and guards field ≥ 1.
- `zigzag` is separate from `encodeVarint` because the choice is the field's declared type, and `sint` routes through it — right call, since −1 as a plain varint drives the robot somewhere real and wrong.
- `sub()` emits an **empty** sub-message rather than optimising it away, matching what the readers on the other side of the file depend on. `finish()` = `varint(len)++body` base64 — and the tests round-trip encoder→`byteCodec` (the reader), so it's a genuine round-trip, not two copies of one guess.

**The capture-found bug is real and the fix is right.** `decodeConsumableHours` now reads the `RUNTIME` (field 1) wrapper and finds parts one level down; the old top-level read found the wrapper where a part should be and answered `undefined` for all nine counters — while the fixtures encoded the same wrong nesting, so they passed against a decode that never worked on hardware. The new fixtures encode the real wrapper (`report = frame(sub(1, parts))`) and one is trimmed from a live report. This is the failure mode a synthetic fixture can't catch alone, called out honestly.

**The other fixes hold:**
- `decodeCleanParamValue` gains an optional inner field number so `mop_mode`'s level(1) and corner_clean(2) are read by number, not "first varint" — otherwise edge-hug mopping reads ON purely because the water level is high. Test pins it.
- One shared `modeCtrlSeq` replaces three per-verb `let seq = 111`; test asserts start/pause/resume seqs are distinct and sequential (124→125→126), which is the whole point of a sequence.
- `resumeCleaning` promoted to an installed `method` (RESUME_TASK=14, captured); `encodeModeCtrl` correctly omits the method field for start.
- `unwrapCleanRecordBlob` uses the checksum as an oracle across length-widths 1/2/4 and refuses a mismatch — a wrong guess fails loudly instead of yielding a plausible wrong run.
- The `station(14)`-present-while-cleaning near-miss (the softest branch from an earlier review) now has a test proving field-3-specificity keeps it out of "docked".

**Restraint is right too:** 17 frames stay `unverified` with no setter (`stopCleaning` because the app sent method 6 not STOP_TASK; all six dock verbs because DP 173 never echoed a `StationRequest`), and `fan` is deliberately not surfaced since it tracks the DP 158 `suction.level` already publishes.

Secrets scan clean (no serial/MAC/SSID/IP/hex in the diff). Ran the 4 key specs locally — 179/179 pass; CI green (2870). Nothing to change. 👍

PR #76, opened 2026-08-24
fix(p2p): recover intermittent live starts on a battery camera (#75)

### LINKED ISSUE #75: Intermittent live source error before first keyframe on T8170
## Downstream

homebridge-plugins/homebridge-eufy-security#1048

## Redacted reproduction

Using `@mega-yfue/eufy-sdk` 0.1.0-beta.31 through a dedicated, unpaired Homebridge 2 instance, I alternated five cold live starts on a battery-powered T8170 and five on a wired T8400 control. VPN was disabled for the complete run. No device name, serial, address, account identifier, media, or raw log is included.

Each start used the same HomeKit selection: High profile, level 4.0, 1280x720 at 30 fps. The controller sent RTCP from session start and retained no media.

## Result

- T8170: 3/5 starts delivered video; successful first-video latency was 2650-5151 ms.
- T8170: 2/5 starts acquired the SDK live handle and then emitted the typed source error before any video keyframe reached the consumer. No adaptation process was started.
- T8400 wired control: 5/5 starts delivered video; first-video latency was 805-1087 ms.
- Every attempt released its HomeKit reservation, adaptation processes, and SDK consumer before the next attempt.

## Expected

After `camera.live()` resolves, the source should either deliver a keyframe within its bounded warm-up/retry policy or expose enough bounded typed failure information to distinguish the failed media-source stage without raw transport data.

## Scope

The downstream plugin does not infer or retry transport behavior. This appears SDK-owned because the live handle was acquired successfully, no source keyframe was delivered, and the SDK emitted its error before plugin adaptation began.

### LINKED ISSUE #1048: Diagnose intermittent live startup on one wired camera
## Parent

#998

## Observation

During real Apple Home verification for #1041, one wired camera was materially less reliable than a second wired control camera. With the VPN disabled, the affected camera sometimes refreshed its snapshot without establishing moving video, and earlier attempts ended with bounded `source-error` or `rtcp-timeout` reasons. A later attempt did stream but appeared lower quality. The control camera then streamed continuously at normal apparent quality.

Both successful process observations used High profile, level 4.0, and `1280x720@30`, so the observed difference is not explained by an under-levelled HomeKit codec selection. The VPN was enabled during the first failed round and can explain RTCP routing failures, but it does not yet explain why this camera remains harder to start than the control camera with the VPN off.

No camera name, serial, address, key material, or media is included here.

## What to establish

- Reproduce repeated cold live starts on the affected wired camera and a wired control camera with VPN state fixed and recorded.
- Attribute each failed attempt to the narrowest stage already owned by the plugin: HAP preparation, SDK source acquisition, first source keyframe, first adapted output, controller RTCP, or cleanup.
- Use the identity-free `live-video-selected` trace from #1041 to record the selected profile, level, geometry, and frame rate for every attempt that reaches start.
- Determine whether the failure is plugin-owned or an SDK media-source defect. If SDK-owned, open or link a minimal redacted upstream reproduction rather than adding transport guesses here.

## Acceptance criteria

- [ ] A bounded repeated-start comparison records pass/fail counts for the affected camera and one control camera with the VPN disabled.
- [ ] Every failed attempt is attributed to a specific lifecycle stage and bounded reason without device identity or media retention.
- [ ] Successful attempts record the HomeKit selection and confirm all adaptation processes and SDK consumers are released after closure.
- [ ] Any plugin-owned defect has a deterministic contract regression test before its fix; any SDK-owned defect is linked to an upstream issue with redacted evidence.
- [ ] Production Homebridge is left active and healthy after live verification.

## Blocked by

- #1041, which adds the identity-free HomeKit selection trace needed to separate controller negotiation from source-start reliability.

## Relationships

- **Next after #1041** in the #998 livestream workstream.
- Related to #1032 because its bounded live-session reasons provide the current failure-stage vocabulary.
- Related to #1035, #1045, and #1047 because first-keyframe delay, short output, or repeated adaptation may look like a source-start failure unless measured separately.

lenoxys:
Closes #75

## What changed

- Own-session `START_LIVE` now carries the fields the current app sends: `msg_id: 1` and `extValue: 1000`.
- Datagram reassembly classifies sequence numbers instead of assuming every non-successor is loss.
  A duplicate or stale datagram is acknowledged and ignored; only a true forward gap invalidates the
  incomplete logical frame. Reassembly state is held per data type, so a control frame arriving mid-burst
  can no longer be appended to a video frame.
- An unacknowledged own-session `START_LIVE` is retained and repeated byte-identically every 150 ms until
  the camera acknowledges it, bounded by 3 s of elapsed time rather than by a send count, then abandoned —
  abandoning it also forgets the channel's started state, so the next tick issues a real start instead of
  nudging a stream that never began. Sequence and GCM ciphertext are unchanged across the repeats. The app
  does the same: 20 identical sends across 272 ms for a camera still waking, one for a camera that
  acknowledged in 9 ms.
- Numbering that restarts mid-connection is followed rather than read as a flood of stale datagrams: a
  repeat is told from a restart by how far back it is numbered (the deepest repeat in the capture is 120
  behind; the bound is 1024), because ignoring a restart would freeze the high-water mark and black-hole the
  data type until the new numbering climbed past it.
- A consumer told why its start failed only gets that `error` when it is listening for one, and gets its
  `stop` either way; the source's teardown runs even if a consumer's listener raises.
- `LiveStreamStartError` reports why a stream failed before its first keyframe: `warm-timeout`,
  `source-error`, or `source-ended`, with `timeoutMs`, `attempts`, and the preserved `cause`.
- Warm-up is bounded until the first **keyframe** rather than the first frame, so a stream that only
  delivers delta frames fails with a typed reason instead of appearing started.
- Bounded, identity-free `debug` traces cover the media command sent, its acknowledgement, retry and
  unacknowledged outcomes, the first inbound video command, first unit, first keyframe, foreign-channel
  media, empty decodes, and dropped datagram gaps. Per-datagram `DATA` logging was removed.

## Why

Redacted evidence from a current app capture and from SDK runs on one battery camera:

- The app's first keyframe is about 150 KB across 148 datagrams and arrives 138 ms after `START_LIVE`.
- In that flow, 890 video datagrams contained 133 immediate duplicates and 41 stale retransmissions,
  with zero true forward gaps.
- The previous logic treated each duplicate or stale datagram as loss and discarded the keyframe being
  assembled, so consumers received only delta frames and waited out the warm-up window.
- Failures that remained after that fix showed a start with no acknowledgement, which the app handles by
  retransmitting the same datagram.

## Verification

- `npm run verify`: 140 files, 2789 tests, plus formatting, typecheck, guards, build, declaration seal,
  ESM smoke test, and example typecheck.
- Hermetic contracts added for duplicate delivery, stale delivery, a genuine forward gap, interleaved data
  types, sequence wraparound, acknowledgement handling, the exact retransmitted bytes, and the app-matched
  payload fields.
- Real-device runs on one battery camera against a wired control, media retained nowhere:
  - SDK-only, before this change: 8/10 with two `warm-timeout` failures.
  - SDK-only, after reassembly fix: 9/10; the single failure started at level-1.
  - SDK-only, with the bounded start retry: 10/10, first keyframe 0.07-1.35 s once warm.
  - Wired control: 10/10 throughout.

## Deferred

- Two host-level failures still occur on that battery camera when a start is never acknowledged and the
  single retransmission does not recover it. That is a separate wake/acknowledgement problem, not the
  reassembly defect fixed here.
- The two payload fields were changed together, so their individual contribution is not isolated.


src/transport/p2p/p2p-session.ts
```diff
@@ -1283,35 +1463,46 @@ export class P2PSession extends EventEmitter {
...
     this.send(addr, RequestMessageType.ACK, buildAckPayload(this.ackTypeHeader(dataType), seqNo));
 
-    // Detect a lost/reordered datagram: sequence numbers are per-dataType and increment by 1. If
-    // this datagram isn't the expected successor and we're mid-reassembly, the in-flight frame is
-    // unrecoverable — drop it rather than splice the wrong bytes in (which corrupts the bottom of a
-    // big keyframe → green/garbled blocks). We resync at the next XZYH frame boundary.
     const prevSeq = this.lastSeqByType.get(dataType);
-    const gap = prevSeq !== undefined && seqNo !== ((prevSeq + 1) & 0xffff);
+    const advance = prevSeq === undefined ? 1 : (seqNo - prevSeq) & 0xffff;
+    if (advance === 0 || advance > SEQUENCE_LOOKBACK) return;
```
martijnpoppen: A large forward jump is misclassified as stale, and because the `return` sits above the `lastSeqByType.set(...)` on line 1489, the high-water mark freezes and the data type is black-holed until the restarted counter climbs back to it.

Any `advance` in `[0x8001, 0xffff]` is read as backwards. That is correct for a genuine retransmission, but it also swallows a counter restart. If the camera restarts its per-datatype counter mid-connection — a fresh `START_LIVE` after a stop — while `prevSeq = P` is below 32768, then for the restarted stream at `seqNo = n`:

```
advance = (n - P) & 0xffff = 65536 - P + n
```

which stays above `SEQUENCE_LOOKBACK` for every `n < P`. So exactly `P` datagrams (up to 32 767) are ACKed and discarded before line 1489 ever runs again and sequencing resyncs. The 3 s warm-up deadline fires long before that.

The pre-PR code always resynced at the next XZYH boundary, so this is a new way to get stuck. `resetInboundSequencing()` is only reachable from `connect()`/`close()`, so nothing clears the frozen mark within a connection.

Moving the `set` above the classification `return` fixes the freeze on its own: a stale datagram would then advance the mark, which is wrong for a true retransmission. The cleaner shape is to keep the ignore for `advance === 0` and small backwards distances, but treat a jump beyond the lookback as a resync — update the mark and drop the pending frame — since neither reading of it can reassemble around the discontinuity.

---

src/transport/p2p/shared-live-source.ts
```diff
@@ -582,20 +618,31 @@ export class SharedLiveSource {
...
+   * An end before the first keyframe is also a failed start, so those consumers get the typed `error`
+   * explaining why nothing played and then the `stop` that closes them — a bare `stop` would look like a
+   * normal end of stream to a caller still waiting for its first frame.
+   */
   private onUpstreamEnd(): void {
     if (this.disposed || !this.stream) return;
     this.logger.debug(
       `${this.tag} upstream ended (station max-duration / reconnect) — notifying ${this.consumers.size} consumer(s)`,
     );
+    if (!this.deliveredKeyframe) {
```
martijnpoppen: Emitting `error` before `stop` can throw out of `onUpstreamEnd` and leave the source leaked.

`c.fail(error)` reaches `ConsumerImpl.fail` (line 243), which does `this.emit("error", err)`. On an `EventEmitter` with no `"error"` listener, Node throws `ERR_UNHANDLED_ERROR` rather than returning. That escapes the `for` loop on line 635, so:

- the remaining consumers never get `fail` **or** the `end()` on line 637,
- `teardown("stopped")` on line 638 never runs, so `onStartFailed()` never fires and the stream, warm-retry interval, deadline timer and ring all leak.

`onUpstreamError` (line 645) has the same exposure, but it is at least an error path already; this one converts a normal station max-duration/reconnect end into a throw for any consumer that only listens for `stop`.

Worth noting that the PR's own tests had to be adjusted around exactly this: `shared_live_source.spec.ts:245` and `readable_egress.spec.ts:71` both gained a keyframe before `emit("stop")`, which routes around the `!this.deliveredKeyframe` branch instead of covering it.

Two ways out, either is fine:

- guard the emit — `if (this.listenerCount("error") > 0) this.emit("error", err)` — so a consumer that only wants `stop` still gets one; or
- wrap the fail loop in `try`/`catch` per consumer so one unhandled emit cannot strand the others or skip `teardown`.

A test where the consumer has no `"error"` listener would pin whichever you pick.

---

src/transport/p2p/p2p-session.ts
```diff
@@ -652,29 +715,54 @@ export class P2PSession extends EventEmitter {
...
+   * An own-session start's encryption follows the session key, not the device family: level-2 GCM once a
+   * key is negotiated, level-1 ECB before that. A key that arrives after a level-1 start makes the next
+   * keepalive tick re-issue the start at level-2, so a camera that only accepts level-2 recovers from a
+   * `live()` that raced ahead of key negotiation.
+   *
+   * Each start is retained until its DATA acknowledgement and repeated once if unacknowledged
+   * ({@link LIVE_START_RETRANSMIT_MS}). Use the `LiveStream` helper for a managed feed with keepalive.
    */
   startLiveMedia(channel: number = STATION_CHANNEL, accountId = "", homeBaseAttached = false): void {
+    this.tracedDatagramGaps = 0;
```
martijnpoppen: This reset makes `MAX_TRACED_DATAGRAM_GAPS` unbounded in practice.

The budget is documented as "per live start", but `startLiveMedia()` is also the keepalive path — `LiveStream` calls it every `DEFAULT_KEEPALIVE_MS` (3000), and it is the warm-retry nudge target too. So `tracedDatagramGaps` is re-armed to 0 roughly every 3 s for the whole life of the stream, and the cap on line 1491 only ever bounds gaps within a 3 s window rather than the log volume it exists to bound.

Resetting it where a start is actually issued — inside the `else` branch on lines 751-753, next to `sendStartLiveOwnSession` — matches the documented intent and leaves the keepalive tick alone.

---

src/transport/p2p/p2p-session.ts
```diff
@@ -652,29 +715,54 @@ export class P2PSession extends EventEmitter {
...
-   * Use the `LiveStream` helper for a managed feed with keepalive.
+   *    START_LIVE, then the 1139 keepalive to hold the stream. A bare 1003 is read as a status query
+   *    there, not a stream.
+   *
+   * An own-session start's encryption follows the session key, not the device family: level-2 GCM once a
+   * key is negotiated, level-1 ECB before that. A key that arrives after a level-1 start makes the next
+   * keepalive tick re-issue the start at level-2, so a camera that only accepts level-2 recovers from a
+   * `live()` that raced ahead of key negotiation.
+   *
+   * Each start is retained until its DATA acknowledgement and repeated once if unacknowledged
```
martijnpoppen: Stale after the last commit: this still describes the count-bounded policy that `f2aa0d7` replaced.

"repeated once if unacknowledged" is no longer what happens — `armLiveStartRetransmit` (line 848) resends every `LIVE_START_RETRANSMIT_MS` (150 ms) until the elapsed-time bound, which across a 3 s deadline is ~19 retransmissions, and considerably more START_LIVE datagrams over a 20 s warm-up once abandonment and re-start cycles are counted. `docs/live-media.md` already documents the real elapsed-bounded behaviour, so the two now contradict each other.

The PR body carries the same stale claim ("retained and retransmitted once, byte-identically, after 500 ms, then abandoned") — worth updating both, since the retransmission volume is the thing a reader would want to know before enabling this against a fleet.

---

martijnpoppen (commented):
Reviewed the transport-recovery work at `f2aa0d7`. The reassembly classification is the right fix for the duplicate/stale problem the capture shows, and the evidence in the PR body is unusually good. Four findings inline, two of them worth fixing before merge.

**Blocking-ish**

1. `p2p-session.ts:1487` — a counter restart mid-connection is classified as stale, and the `return` sits above the high-water-mark update, so the data type is black-holed until the restarted counter climbs back past the frozen mark (up to 32 767 datagrams). The warm-up deadline fires first. The pre-PR code always resynced at the next XZYH boundary, so this is a new failure mode.
2. `shared-live-source.ts:633` — the new `error`-before-`stop` ordering throws `ERR_UNHANDLED_ERROR` for a consumer with no `"error"` listener, stranding the remaining consumers and skipping `teardown()`, so timers and the ring leak. The two test files amended with a keyframe before `emit("stop")` are routing around this branch rather than covering it.

**Non-blocking**

3. `p2p-session.ts:730` — `tracedDatagramGaps = 0` sits on the keepalive path, so the `MAX_TRACED_DATAGRAM_GAPS` budget is re-armed every 3 s and never bounds anything.
4. `p2p-session.ts:726` — the JSDoc still says "repeated once", which `f2aa0d7` replaced with a 150 ms elapsed-bounded repeat; `docs/live-media.md` and the PR body disagree with each other on this.

Checked and found sound: the DATA-vs-VIDEO split in `onAck` (the `0xd1 0x00` / `0xd1 0x01` headers match `buildCommandHeader`'s default and the audio path), the shared `retransmitUnacked` sweep (delete-then-reinsert during `Map` iteration is safe since `sentAt` takes the sweep's captured `now`), per-dataType `pendingByDataType` reassembly, sequence wraparound at 65535 → 0, byte-identical retransmission, and the `acceptsMedia`/`annexbOf` trace insertions (no control-flow change).

---

src/transport/p2p/p2p-session.ts
max246: **Blocking.** This `return` sits above the `lastSeqByType.set(dataType, seqNo)` on the next-but-one line, so a datagram classified stale never advances the high-water mark. That's correct for a genuine duplicate/retransmission, but it misclassifies a **counter restart**: with `SEQUENCE_LOOKBACK = 0x8000`, a restart from prevSeq 20000 to 0 gives `advance = (0-20000)&0xffff = 45536 > 32768` → stale → returned, and the mark stays frozen at 20000, so every restarted datagram (0,1,2,…) is also stale until it climbs back past 20000 — ~20k datagrams black-holed, past the warm-up deadline. Pre-PR this resynced at the next header boundary.

The fix needs to tell a restart apart from a stale retransmit rather than lumping both under `advance > LOOKBACK`: a real retransmit is only a few hundred back, so a large backward jump is a restart and should **resync** (reset `pendingByDataType`, update `lastSeqByType`) instead of being ignored. Bounding the stale window to a plausible retransmit depth, or resyncing on a header-bearing datagram, both get there — but simply moving `.set()` above the `return` is NOT safe on its own, since it would let a true stale retransmission drag the mark backwards. Worth a test that drives a restart and asserts the stream recovers.

src/transport/p2p/shared-live-source.ts
```diff
@@ -582,20 +618,31 @@ export class SharedLiveSource {
...
+      const error = this.startFailure("source-ended");
+      for (const c of [...this.consumers]) c.fail(error);
```
max246: **Blocking.** `c.fail(error)` → `ConsumerImpl.fail` → `emit("error", err)`. A consumer whose EventEmitter has no `"error"` listener makes Node throw `ERR_UNHANDLED_ERROR` synchronously right here, which propagates out of this loop — so the `for (… ) c.end()` below and `teardown("stopped")` never run: the other consumers are stranded and the keepalive timer + ring leak. `onUpstreamError` (a few lines down) has the identical pattern.

Emit `stop`/`end` before `error`, or guard the emit (skip when `listenerCount("error") === 0`, or emit inside a `try`), and make teardown run regardless (e.g. a `finally`) so one throwing consumer can't take out the rest. The two amended tests set `deliveredKeyframe` first, so they route around this branch — a test with a listener-less consumer failing before its first keyframe would cover it.

max246 (changes requested):
The core idea is right — the duplicate/stale reassembly classification is the correct fix for the flood the capture shows, and the evidence in the body is unusually good. But I independently confirmed the two blocking defects martijnpoppen already flagged; both are real and neither is addressed at this head (`f2aa0d7`, threads still open). Requesting changes on those two. The two non-blocking notes (unbounded trace budget on the keepalive path; the stale "repeated once" docs) also hold and are worth folding in.

**1. Reassembly black-holes a data type on a counter restart** (`p2p-session.ts:1487`). `SEQUENCE_LOOKBACK = 0x8000`. A mid-connection counter restart where the drop exceeds half the space computes a wrapped `advance > 32768` → classified stale → the `return` fires **before** `lastSeqByType.set(...)`, so the high-water mark freezes and every subsequent restarted datagram is also stale. Worked example: prevSeq 20000, device restarts at 0 → `advance = (0-20000)&0xffff = 45536 > 32768` → ignored; seqNo 1,2,… stay ignored until the counter climbs back past 20000 (~20k datagrams), long after the warm-up deadline. The pre-PR code resynced at the next header boundary, so this is a NEW failure mode the fix introduces.

**2. `error`-before-`stop` can abort teardown and leak the source** (`shared-live-source.ts:635`, and `onUpstreamError` just below has the same shape). `c.fail(error)` reaches `ConsumerImpl.fail` → `emit("error", …)`; on a consumer with no `"error"` listener Node throws `ERR_UNHANDLED_ERROR` synchronously, which propagates out of the `for` loop so the remaining `c.end()` loop and `teardown("stopped")` never run — other consumers stranded, timers and the ring leaked. The two amended tests set a keyframe first (`deliveredKeyframe === true`), so they skip this branch rather than cover it.

Everything else martijnpoppen checked I also found sound (the DATA-vs-VIDEO `onAck` split, the shared `retransmitUnacked` sweep, per-dataType `pendingByDataType`, wraparound at 65535→0, byte-identical retransmission). Once the two above are fixed — and ideally a test that exercises each (a counter restart that must resync; a listener-less consumer failing before its first keyframe) — this is good to go.

src/transport/p2p/p2p-session.ts
lenoxys: Right, and confirmed by test before fixing: the `return` above `lastSeqByType.set` freezes the mark, and the pre-PR code always resynced at the next boundary, so this branch introduced it. Worse than the review says in one respect — this branch's own abandon-and-restart is exactly what makes a camera begin a fresh stream on a connection already up, so the PR raises the odds of the case it can't survive.

Fixed in f3929a1, telling a repeat from a restart by depth rather than direction, with the depth measured rather than guessed. Re-parsed the app capture behind this PR: across 1758 video datagrams from two own-session cameras, the deepest genuine repeat arrived **120** numbers behind the mark (per-stream max 120 and 38; distances are heavily clustered at 1-4 and ~36-38), so `STALE_RETRANSMIT_DEPTH = 1024` sits an order of magnitude clear of a real repeat.

The depth is also what bounds the remaining cost in both directions, which is why I did not just move the `set`: a repeat mistaken for a restart discards one logical frame, and a restart mistaken for a repeat now black-holes at most `min(prevSeq, 1024)` datagrams instead of up to 32767 — bounded and self-healing rather than terminal. A restart traces under its own `sequence-restart` phase, since a reader chasing a dead stream needs to know which of the two happened.

New test `resynchronizes onto numbering the device restarts mid-connection` (prevSeq 20000 → restart at 0) fails on the old code exactly as described — the 1300 and 1301 frames never arrive — and a companion test pins a 120-back repeat as still ignored, so the depth cannot be quietly shrunk past the evidence.

src/transport/p2p/shared-live-source.ts
lenoxys: Right, and it is worse than a leak: that emit is synchronous, inside a fan-out reached from the transport's datagram handler, so `ERR_UNHANDLED_ERROR` becomes an uncaught exception in the host — a plugin crash caused by a consumer that did nothing wrong but listen only for `stop`. And `onUpstreamEnd`'s `fail` is new in this PR, so a clean end of stream became a throw.

Fixed in 0f9d0c9, taking both of your suggestions where each belongs:

- `ConsumerImpl.fail` skips the emit when `listenerCount("error") === 0`, so the source never manufactures an unhandled error, and that consumer still gets its `stop` from `end()`. Same shape as the talkback fan-out at `command-router.ts:615`, so there is one idiom for this in the module.
- `onUpstreamEnd` and `onUpstreamError` tear down in a `finally`, so a caller's throwing listener cannot leak the stream, the warm timers or the ring. The throw still propagates — a listener that raises is the caller's defect to see, not the source's to swallow — which is why the guard, not the `try`, is the actual fix.

Two tests, both red on the old code: one where a listener-less consumer is attached ahead of a watching one and every signal must still land (the old code threw the `LiveStreamStartError` straight out of `emit("stop")`), one where a consumer's error listener throws and the source must still reach `stopped` with the stream stopped (the old code stayed `warming`). I left the two amended tests as they are — they assert a warmed end-of-stream, which is a case worth keeping — and covered the branch they skipped directly.

Not changed: on `onUpstreamError` a listener-less consumer still gets no `stop`, only the suppressed `error`. That asymmetry predates this branch, and adding an `end()` there would push a `stop` at consumers that already destroyed their egress on `error`. Say the word if you want it in this PR.

src/transport/p2p/p2p-session.ts
lenoxys: Right. Fixed in f3929a1 by re-arming where a start is actually issued — in the own-session start branch and the attached branch, not on the keepalive tick — and the constant's JSDoc now names the keepalive path, since that is the thing that made a budget documented "per live start" bound a 3 s window instead of the log.

src/transport/p2p/p2p-session.ts
lenoxys: Right. JSDoc corrected in f3929a1 to the elapsed-bounded repeat, and the PR body rewritten — it now states the 150 ms cadence, the 3 s elapsed bound, that abandonment forgets the channel's started state, and the app's own 20-sends-in-272 ms evidence for repeating rather than sending once. `docs/live-media.md` was already accurate; it gained only the mid-connection restart resync from the finding above.

src/transport/p2p/p2p-session.ts
lenoxys: Right, and confirmed by test before fixing: the `return` above `lastSeqByType.set` freezes the mark, and the pre-PR code always resynced at the next boundary, so this branch introduced it. Worse than the review says in one respect — this branch's own abandon-and-restart is exactly what makes a camera begin a fresh stream on a connection already up, so the PR raises the odds of the case it can't survive.

Fixed in f3929a1, telling a repeat from a restart by depth rather than direction, with the depth measured rather than guessed. Re-parsed the app capture behind this PR: across 1758 video datagrams from two own-session cameras, the deepest genuine repeat arrived **120** numbers behind the mark (per-stream max 120 and 38; distances are heavily clustered at 1-4 and ~36-38), so `STALE_RETRANSMIT_DEPTH = 1024` sits an order of magnitude clear of a real repeat.

The depth is also what bounds the remaining cost in both directions, which is why I did not just move the `set`: a repeat mistaken for a restart discards one logical frame, and a restart mistaken for a repeat now black-holes at most `min(prevSeq, 1024)` datagrams instead of up to 32767 — bounded and self-healing rather than terminal. A restart traces under its own `sequence-restart` phase, since a reader chasing a dead stream needs to know which of the two happened.

New test `resynchronizes onto numbering the device restarts mid-connection` (prevSeq 20000 → restart at 0) fails on the old code exactly as described — the 1300 and 1301 frames never arrive — and a companion test pins a 120-back repeat as still ignored, so the depth cannot be quietly shrunk past the evidence.

src/transport/p2p/shared-live-source.ts
lenoxys: Right, and it is worse than a leak: that emit is synchronous, inside a fan-out reached from the transport's datagram handler, so `ERR_UNHANDLED_ERROR` becomes an uncaught exception in the host — a plugin crash caused by a consumer that did nothing wrong but listen only for `stop`. And `onUpstreamEnd`'s `fail` is new in this PR, so a clean end of stream became a throw.

Fixed in 0f9d0c9, taking both of your suggestions where each belongs:

- `ConsumerImpl.fail` skips the emit when `listenerCount("error") === 0`, so the source never manufactures an unhandled error, and that consumer still gets its `stop` from `end()`. Same shape as the talkback fan-out at `command-router.ts:615`, so there is one idiom for this in the module.
- `onUpstreamEnd` and `onUpstreamError` tear down in a `finally`, so a caller's throwing listener cannot leak the stream, the warm timers or the ring. The throw still propagates — a listener that raises is the caller's defect to see, not the source's to swallow — which is why the guard, not the `try`, is the actual fix.

Two tests, both red on the old code: one where a listener-less consumer is attached ahead of a watching one and every signal must still land (the old code threw the `LiveStreamStartError` straight out of `emit("stop")`), one where a consumer's error listener throws and the source must still reach `stopped` with the stream stopped (the old code stayed `warming`). I left the two amended tests as they are — they assert a warmed end-of-stream, which is a case worth keeping — and covered the branch they skipped directly.

Not changed: on `onUpstreamError` a listener-less consumer still gets no `stop`, only the suppressed `error`. That asymmetry predates this branch, and adding an `end()` there would push a `stop` at consumers that already destroyed their egress on `error`. Say the word if you want it in this PR.

lenoxys:
All four findings were right, and both blocking ones are fixed at `0f9d0c9` with a test that is red on the old code for each. Threads answered individually.

**1. Reassembly black-holes a data type on a counter restart** — `f3929a1`. Confirmed by test, and the pre-PR code did resync at the next boundary, so this branch introduced it. I did not simply move the `set`: a repeat is now told from a restart by *depth*, and the depth is measured rather than guessed — re-parsing the app capture behind this PR gives a deepest genuine repeat of **120** numbers behind the mark across 1758 video datagrams from two own-session cameras, so `STALE_RETRANSMIT_DEPTH = 1024` clears it by an order of magnitude while capping the black hole at `min(prevSeq, 1024)` datagrams instead of 32767. A restart traces under its own `sequence-restart` phase.

**2. `error` before `stop` strands consumers and skips teardown** — `0f9d0c9`. Also worse than described: the emit is synchronous inside a fan-out reached from the datagram handler, so `ERR_UNHANDLED_ERROR` surfaces as an uncaught exception in the host — a plugin crash caused by a consumer whose only sin was listening for `stop`. `fail` now skips the emit when nothing is listening (matching `command-router.ts:615`) and that consumer still gets its `stop`; teardown moved into a `finally` so a throwing caller listener cannot leak the stream or timers, while the throw still propagates.

**3. Trace budget re-armed on the keepalive path** — `f3929a1`, re-armed where a start is actually issued.

**4. Stale "repeated once" JSDoc** — `f3929a1`, and the PR body rewritten to the 150 ms / 3 s elapsed-bounded policy including that abandonment forgets the channel's started state.

One thing deliberately left alone, flagged in the thread: on `onUpstreamError` a listener-less consumer still gets no `stop`. That asymmetry predates this branch, and pushing a `stop` there would reach consumers that already destroyed their egress on `error` — happy to fold it in if you want it here.

`npm run verify` green: 140 files, 2800 tests, plus format, typecheck, all four guards, build, seal, ESM smoke and example typecheck. No new live-device run — both fixes are transport classification and fan-out ordering, and the restart case is not one I can provoke on demand against the fleet, so the evidence for it is the capture measurement plus the hermetic contracts.

max246 (approved):
Re-reviewed at `0f9d0c9` — **both blockers are correctly fixed, and both the way I'd hoped.** Clearing my earlier CHANGES_REQUESTED.

**Finding #1 (reassembly counter-restart) — fixed, and it avoids the unsafe shortcut.** The new classification adds `STALE_RETRANSMIT_DEPTH = 1024`: a datagram beyond `SEQUENCE_LOOKBACK` is a **restart** (resync — high-water moves to the new numbering) only when it is more than 1024 behind, and a stale retransmit within that depth still `return`s **without** advancing the mark. That's exactly the distinction I flagged — it does NOT just move `.set()` above the return, so a genuine stale retransmit can't drag the mark backward. Worked example holds: prevSeq 20000, restart at 0 → advance 45536, backward 20000 > 1024 → resync (was: black-holed ~20k datagrams). The depth is well-justified (measured deepest repeat 120 behind; 1024 is an order of magnitude clear) and the cost-asymmetry reasoning is right (a repeat misread as a restart drops one frame; the reverse black-holes, so the bound caps the expensive error). Pinned by `data-reassembly.spec`'s "resynchronizes onto numbering the device restarts mid-connection".

**Finding #2 (listener-less consumer strands the source) — fixed, defense in depth.** `ConsumerImpl.fail` now emits `"error"` only when `listenerCount("error") > 0`, so a consumer that only wanted `stop` no longer throws `ERR_UNHANDLED_ERROR` out of the synchronous fan-out; it still gets its `stop` via `end()`. And `onUpstreamEnd`/`onUpstreamError` now wrap the notify loops in `try { … } finally { teardown("stopped") }`, so even if a consumer's own error-listener throws, the source's stream/timers/ring are released regardless — while the throw still propagates as the caller's defect to see. Both levers I suggested, and the finally is the important one.

**Non-blocking items also handled:**
- #3 — the `tracedDatagramGaps` reset moved off the own-session keepalive branch (it now resets only where an own-session start is actually issued), so the `MAX_TRACED_DATAGRAM_GAPS` budget bounds the log again on the path this PR targets. (The attached branch still resets per call, but that's a HomeBase-topology log-volume detail, not the battery-camera case here.)
- #4 — the "repeated once" JSDoc is corrected to the elapsed-bounded retransmit (`LIVE_START_RETRANSMIT_MS` until ack, bounded by `LIVE_START_ACK_DEADLINE_MS`).

50/50 in the affected p2p specs locally (`data-reassembly`, `shared-live-start-failure`, `shared_live_source`, `live-start-ack`); CI green. Everything else I'd already found sound is unchanged. Good to merge — which also unblocks #79 and #90. 👍

PR #78, opened 2026-08-25
feat(clean): schedules off DP 164, and DP 162 is a voice pack rather than a locale

martijnpoppen:
Stacked on #74 — review that one first, or this diff will read as containing its six commits too.

Three commits. The middle one is a **behaviour fix on a property that has been wrong since it was added**, and is the part most worth a reviewer's attention.

## What changed

| # | Commit | Item |
|---|--------|------|
| 1 | Doc-comment line breaks | 24 JSDoc blocks across the two vacuum capabilities |
| 2 | Schedules read | **D2 closed** — `TimerResponse` on DP 164 |
| 3 | Voice pack | **§2's DP 162 correction**, finally applied to the member |

## Why

**D2 was the plan's last "genuinely needs a capture" item, and it did not need one.** The product catalogue dump on #74 named the DP; the vendor's `timing.proto` had carried the whole message all along. Only the number joining them was missing.

`TimerResponse.timers` is a repeated `TimerInfo` four containers deep with a `oneof` for what each timer runs — deeper than anything else on this line, which is why it decodes in its own `src/model/vacuum-schedules.ts` rather than as one more field reader inside an already 2,100-line `vacuum-clean.ts`. It reads a payload and answers a list; it never asks the device for one and holds no DP number of its own.

Three members come out of it. `scheduleCount` owns DP 164, `activeScheduleCount` reads the same payload through `readsFrom`, and `schedules()` answers the list a scalar property cannot hold — a `method` member with `answers: true`, so it is not offered as a button by anything enumerating the surface.

`enabled` and `valid` are separate answers on purpose: a timer pointing at a deleted scene is kept and reported invalid rather than removed, which is why "three schedules" and "three that will fire" are different numbers, and why a host showing only the first explains nothing when a run does not happen.

## The DP 162 fix (commit 3)

§2 flagged this, the catalogue confirmed it, and the **member was never changed**.

`language` published DP 162 as `type: "string", kind: "text"` described as a locale code. DP 162 carries a `LanguageResponse`, so what a caller actually read was the base64 of a protobuf message. A string property answering a string — which is exactly why nothing looked wrong.

The message holds four numbers and no language tag anywhere, so the property is replaced rather than repaired:

```
language  →  voicePack          LanguageResponse.current_id
             defaultVoicePack   .default_id — the pack the firmware shipped with
             voicePackState     .state — idle / updating / success / failure
             voicePackVersion   .version
```

**Breaking:** `dev.vacuumClean().language` is gone. It never answered a locale, so nothing that reads it works today — a host showing it was showing base64.

The test for it had the same shape as the consumables bug #74 fixed: it fed `{ value: "en" }` and asserted `"en"`, a value the device does not send. It now feeds real `LanguageResponse` bytes.

`voicePackState` earns its place. A pack is downloaded from a CDN and md5-checked by the device, so a selection is neither instant nor guaranteed, and this is the field that says which happened.

## Evidence

Field numbers are the vendor's own — `timing.proto` and `language.proto`, cited per constant table — plus the T2351 product catalogue from #74 for both DP numbers. Three readings this message makes easy to get wrong, each pinned by a test:

- **The action comes from the `oneof` branch, never from `Action.type`.** `SCHEDULE_AUTO_CLEAN = 0` is the enum's first member, so proto3 omits the type on every auto-clean timer there is. Reading it alone reports a scene timer as an auto-clean the moment a firmware leaves the type out. `type` is consulted last and only to name a branch this version does not know.
- **Midnight and "no time set" are the same bytes.** `hours` and `minutes` are both zero-valued and both omitted for a 00:00 timer, so absent has to read as zero rather than as missing.
- **Rooms sit in one of two places.** X9-era firmware numbers the list 6 inside `General`; X10-era numbers it 3 inside `Custom`. The vendor kept both, neither is a fallback for the other, so both are read and `Custom` wins.

`TimerInfo.Addition` is deliberately **not** decoded. It carries the account ids of whoever created and last edited each timer. A host showing a schedule needs the time and the days, not who set it, and those ids are not a value to hand a caller without a reason — R7.

Commit 1 is `n/a` on evidence: 24 JSDoc blocks carried a literal backslash-n where a line break belonged, so the prose ran together in typedoc output and in editor hover. They arrived through a GitHub API commit path that escaped the newlines once too often. Same text, breaks restored, re-run through prettier.

## Verified how

- [x] `npm run verify` is green locally — 141 files, **2918 tests** (+15)
- [ ] Ran against a real device — **no.** Both decodes are offline, against synthetic byte fixtures built from the vendor schema.
- [x] Added or updated specs (offline, synthetic fixtures) — 19 for the schedules decode, 4 for `LanguageResponse`, plus the surface assertions

## Unverified or deferred

- **No timer write.** Adding, deleting, opening or closing a schedule is a `TimerRequest`, and no capture has shown the device accepting one. The frame is not built either — declaring an encoder nobody can call is not worth the surface until a capture exists.
- **No voice-pack setter, and not for want of a capture.** Selecting a pack sends a `LanguageRequest.Desc` carrying a CDN url, an md5 the device verifies, a size and a version. There is no single value a `setVoicePack(x)` could take, so the id is reported raw and the id→name table stays the host's, as the plan's Tier C says it should.
- **`schedules()` answers from state already received.** The robot pushes its whole timer list on boot and after any change, so this sends nothing — but a device that has not reported DP 164 in this session has no schedules to read, and the member is gated off entirely rather than answering an empty list it cannot stand behind.
- **`scheduleCount` warns once per update on a raw payload.** It declares `type: "number"` over a base64 wire, so `coerceByType` logs "declared number but wire value is not numeric — keeping raw" before the getter decodes it. Pre-existing behaviour shared with all nine consumable counters and every other Raw-DP member; worth a follow-up across the line rather than a special case here.

---


max246 (approved):
Reviewed the schedules decoder and the DP 162 replacement against their tests. Ready — stacked on #74 (already approved), CI green, 174/174 in the affected specs locally.

**Schedules decode (`vacuum-schedules.ts`) — the deep-nesting traps are all handled:**
- `each()` walks repeated fields via `flatMap` rather than `find`, so a repeated `TimerInfo` (and a repeated room list) isn't collapsed to its first element. This is the thing most likely to go wrong at four containers deep, and it's right.
- The `oneof` is read **by branch presence** (rooms/cruise/scene), with `Action.type` consulted **last** and only to name an unknown future branch — so `SCHEDULE_AUTO_CLEAN = 0` being proto3-omitted doesn't make every timer read as auto-clean. Pinned by a test.
- Rooms read `Custom` then `General` with the correct per-generation field numbers (3 vs 6), neither a fallback for the other.
- `enabled`/`valid` kept as separate answers; `undefined` (unreadable) vs `[]` (genuinely no schedules) kept distinct.
- `TimerInfo.Addition` (creator/editor account ids) deliberately not decoded — right call (R7).

**DP 162 fix is correct and the breaking change is justified.** `voicePack` owns DP 162 (`current_id`); `defaultVoicePack`/`voicePackState`/`voicePackVersion` read the same payload through `readsFrom` (the #66 mechanism). The removed `language` property only ever published the base64 of a `LanguageResponse` typed as a locale string — the same never-worked-but-tests-passed shape #74's consumables fix exposed, and the old test fed `{value:"en"}`, a value the device never sends. Replacing rather than repairing is right, and it's flagged Breaking; on a pre-1.0 line where the property returned garbage, that's fine.

**One minor, non-blocking doc note:** the comment on `enabled: flag(status, OPENED)` says *"an absent `Status` is a timer the device... considers usable, so the zero-omission default reads the right way round"* — but an absent `Status` container makes both `enabled` and `valid` read `false`, i.e. not-usable. The field-level reads are correct (a real enabled/valid timer sends `OPENED`/`VALID = 1`); it's just the comment's *container-absent* justification that reads backwards from what the code does. Worth a word so it doesn't mislead the next reader. No code change needed.

No device data in the diff; every fixture synthetic.

PR #79, opened 2026-08-25
fix(camera): correct the on/off wire and make enablement answerable (#77)

### LINKED ISSUE #77: A direct-connect T8410 never delivers a live frame: its level-2 key never arrives and it sends an unmodelled 0xf121 message
## Observation

One camera on this account never delivers a live frame, on any attempt. It is not intermittent: four consecutive attempts, two through the plugin's adaptation seam and two through a direct SDK probe, all ended identically.

```
LiveStreamStartError: live stream failed to start
  (warm-timeout at awaiting-first-frame after 10 attempts; 20000ms deadline)
```

`snapshotLive()` fails with the same cause underneath it, and `snapshotStored()` reports `StoredSnapshotUnavailableError: No stored snapshot is available`, so this camera produces **no media at all** through any path.

The device is a direct-connect `T8410` (Indoor Cam Pan & Tilt), `deviceType` 31, firmware `2.3.2.4`, hardware `P3`, mains powered. Two other wired cameras on the same account streamed 292 and 439 access units in sessions minutes either side of these attempts, so the account, session, and transport were all healthy.

## Trace

Full `logger` output for one attempt, serials reduced to model plus last four:

```
[p2p] T841..197F sendLookups: cloud -> <addr>:32100, p2p-sgp.anker-in.com:32100, p2p-par.anker-in.com:32100 self=<addr>:42956
[p2p] T841..197F beginCheckCam -> <addr>:26444 (+/-3)
[p2p] T841..197F connected <addr>:26444
[p2p] T841..197F UNHANDLED payload hex: f121000400000000
[p2p] T841..197F UNHANDLED payload hex: f121000400000000
[p2p] T841..197F waiting up to 7958ms for the level-2 key
[p2p] T841..197F UNHANDLED payload hex: f121000400000000
[p2p] T841..197F level-2 key did not arrive within its grace
[live T841..197F:0] warming (retry=2000ms deadline=20000ms)
... 17 live-start traces, 10 START attempts ...
[live T841..197F:0] live stream failed to start (warm-timeout at awaiting-first-frame after 10 attempts; 20000ms deadline) (consumers=1)
```

186 inbound datagrams arrived during the window, almost all `f1e0`/`f1e1` ping-pong with 16 `f1d0`/`f1d1` data and ack. So the camera is connected and talking; it just never sends media.

## Two findings, and only one of them is the cause

**`0xf121` is an unmodelled response type.** `ResponseMessageType` in `src/transport/p2p/codec.ts` models `f140`, `f141`, `f142`, `f182`, `f184`, `f1e0`, `f1e1`, `f1d0`, `f1d1` and `f1f0`. This camera sends `f1 21` with a declared length of 4 and a zero body, three times, straddling the level-2 key wait, and `p2p-session.ts` falls to the `UNHANDLED payload hex` branch each time. Nothing consumes it.

**The level-2 key grace expiring is *not* the discriminator, despite appearances.** A working HomeBase-attached camera on this same account logs the same line and streams fine:

```
[p2p] T844..0237 waiting up to 7579ms for the level-2 key
[p2p] T801..2199 level-2 key negotiated (cipher_id 209)
[p2p] T844..0237 level-2 key did not arrive within its grace
[live T844..0237:0] first keyframe — live (640x360 h264, powered=wired)      <- 439 frames follow
```

Its *station* negotiates the key and the camera's own grace expiring costs it nothing. The failing camera owns its session, so there is no station to negotiate on its behalf and nothing else supplies the key. That difference — direct-connect versus attached — is where I would look first, and it is why I am not filing this as "the grace is too short".

## What to establish

- What is `0xf121`? A zero body and a length of 4 look like an acknowledgement or a state notice, and it arrives exactly around the key exchange. If it is the camera's answer to the key request, the handshake is stalling on a message the SDK discards.
- Does a direct-connect camera need a different level-2 key exchange from an attached one, or the same one addressed differently? The attached path demonstrably works.
- Does the app get media from this camera? A capture would settle both questions at once, as it did for #75.

## Why this is not #75

#75 is an intermittent start on a `T8170` that negotiates its keys and does deliver frames, where the START was not acknowledged and retransmitted keyframes were being discarded. This camera negotiates no key, sends an unmodelled message, and has never delivered a frame in any attempt. The `f2aa0d7` repeat-until-taken fix is visibly active here — it made the 10 attempts — and does not help.

## Reproducing

Against a copy of a persisted storage root, with the plugin instance stopped:

```js
const client = new EufyMega({ /* persisted session */, logger });
const actions = (await client.getDevice('T8410XXXXXXXXXXX')).camera();
await actions.live();   // resolves, then errors at awaiting-first-frame after 20s
```

No device or account identity, address, key material, or media is included above.

lenoxys:
Stacked on #76 (base: `fix/live-start-transport-recovery`). Review that first; the diff here is one commit.

## The camera in #77 is switched off

Neither hypothesis in the issue was the cause. Both are disproved by probing the account, and both disproofs are recorded in-tree where the lines that prompted them are emitted.

- **`0xf121`** did not recur across later probes of that camera. What *does* reach the UNHANDLED branch is `0xf169`, a relay-pool listing — on the sessions of cameras that stream and cameras that do not alike.
- **The level-2 grace** is not the discriminator. Own-session cameras of two device types negotiate a key; three others never do, and cameras from that second group stream normally at level-1 — including one of the **same firmware** as the camera that delivers nothing.

The camera accepts the start and answers without video: 234 audio frames, no video frame, no stream-status report, **zero datagram gaps**, control reply byte-identical to a working camera's. Turned on: 217 video access units. Turned back off: none.

Establishing that, then capturing what the app actually does, corrected four things this SDK had wrong.

## 1. `audio-only` — a failed start says whether the source was silent

`LiveStreamStartStage` gains a member, naming only what the source *did* and never why: a device with no picture to send and one whose video this build cannot read both reach it.

## 2. Enablement converges instead of freezing

Nothing pushes it, so a write left the reported value **frozen**:

```
setEnabled(true)  -> succeeded; camera on, streamed 172 access units
enabled  +0s / +2s / +7s / +22s   ->   false  false  false  false
```

Publishing "this camera is off" meant publishing something that could never come back. Enablement now declares an `observation`, as guard mode already did — and two gaps in that contract closed first, because it is the first member whose **written wire and reported wire differ**:

- **`reflects(value, ctx)`** replaces `expected(value)` + the implicit `m.param`, so the readback names the param the device *actually reports*. Measured: 5 cameras report the enablement param and never `2001`, 3 report `2001` and never it, **none both**. Answering nothing means no readback exists, so the command dispatches unobserved instead of waiting out an unsatisfiable timeout.
- **`observed`** separates the decoded expectation from the raw one. Both checks compared one figure against different things; guard mode's coincide so it never surfaced. A disable bit reports `0` for a property that reads `true` — and the first live run proved it: **the value converged and the transition event never fired.**

Verified on both wire families, both directions: converges in **2–6 s** on the same bound object, `cameraEnabledChanged` fires **exactly once**, and immediately after the write it still reads the old value — delivery is not convergence.

## 3. The on/off wire was wrong — corrected against the app's own frames

Camera power is `CMD_DEVS_SWITCH` (1035) on **every** family. Across six cameras of four device types and both topologies, every on/off the app sent was `1035` carrying one body — `[u32 channel][u32 value][account_id]`, the channel selecting an attached camera. **The capture contains no `6250` frame at all.**

This module diverted the mini / S350 / outdoor-PT families to that privacy envelope on the strength of the decompiled V6 parsers. The app does not use it, and the envelope has no level-1 form — so it could not be sent at all on a session whose one-shot negotiation concluded without a key, observed refusing the write to turn an outdoor-PT camera back on and leaving it dark.

Writing the param this member already **reads** is also what makes the value track its own setter, so `readReflectsWrite` and the `unreflectedMembers` entry it produced are gone, and the observation covers every family. Write polarity is confirmed by the same frames: `1` turns a camera on.

## 4. The encryption level follows the session's key, not topology

| peer | topology | key | `1035` | `1350` |
| --- | --- | --- | --- | --- |
| HomeBase (+4 attached) | attached | yes | `sign8` | `sign8` |
| own-session camera A | **own** | yes | **`sign8`** | **`sign8`** |
| own-session camera B | **own** | yes | **`sign8`** | **`sign8`** |
| own-session camera C | own | no | — | `sign1` |
| own-session camera D | own | no | `sign1` | `sign1` |

Every peer used **one seal for every command family it sent**. Topology mispredicts the two keyed own-session cameras — which is where a level-2-only wire was being chosen for a session holding no key. One decision point, renamed for what decides it.

**And a required level-2 key is now asked for twice.** The negotiation is one-shot per connection, so a gateway reply that never landed refused every later level-2-only command on that connection although a fresh session over the same device negotiates normally. Bounded to one extra ask; never for a caller with a level-1 path.

## Confirmation reads what the device volunteered — opportunistically

`CMD_CAMERA_INFO` carries a camera's own params as a **root-level** array, which this session parsed and discarded; a station's notify nests them under `payload`, so each id is read only in its measured shape. The readback then asks before fetching where there is a concrete value to compare — a push has no expectation and must still fetch, since "converged" there means only "differs from before".

Measured: one camera's write went from 4 cloud requests to **0**, another from 16 to **0**, and the worst case from 54 without converging to 8.

**This is opportunistic, not a replacement.** The app received no P2P param report across 12 toggles, so the cloud readback stays the dependable path and is untouched.

## Deliberately not done

- **The privacy param's alias.** `6250="1"` is now observed with no video, alongside the `"0"` already known — the polarity the code asked for — and it is reported by the outdoor-PT family and no other camera measured. The alias is still **not** taken: the power param moved in the *same* step, so the reading cannot say whether power and privacy are one state or two, and the app drives `1035`.
- **An up-front refusal of video egresses.** An earlier revision of this PR did that. Measurement killed it: the reading never converges, so it would have withheld video from an **on** camera for the bound object's whole lifetime. Re-reading first does not help either — the cloud lagged **5.6 s** behind the write, and per-device param reads are owner-gated on this account so most "fresh" reads are served from cache in 1–3 ms.
- **`snapshotStored`.** Its `not-observed` answer here is correct: a camera that has been off recorded no events and banked no thumbnail.
- **`LiveStreamStartStage` gains a member.** No consumer in this repo switches on it exhaustively.

## Open, and not blocking

Whether the security MQTT broker **pushes** camera param state — which would make confirmation free rather than merely cheaper. A listen test was inconclusive: several subscribes returned errors, so an empty result cannot distinguish "no push" from "not subscribed", and the app's real topic shapes are still unknown. It changes only the *cost* of confirmation; no correction here depends on it.

## Verified

`npm run verify` green: 140 files / 2814 tests, all guards. Baseline 2796 → +18. Every new assertion mutation-checked — polling the written param instead of the reported one, dropping the audio record, dropping the per-generation reset, and mis-reading the enablement property each fail the suite.

Live: the switched-off camera reports `audio-only` through the public capability path; enablement converges and fires once on both wire families; three cameras across attached and own-session topologies stream 145–178 access units unaffected. The fleet was audited back to its original state after every write.

Closes #77.


max246 (approved):
The work here is excellent and verified on its own merits — approving the content. **But it is stacked on #76, which is not mergeable: #76 still carries two unaddressed blocking findings (the reassembly counter-restart black-hole at `p2p-session.ts:1487`, and the `error`-before-`stop` `ERR_UNHANDLED_ERROR` leak at `shared-live-source.ts:633`). #79 builds on #76's transport code in the same files and does not fix either, so this cannot land until #76 is corrected and merged and this rebased onto `beta-0.1.0`.** Flagging that up front so the green check isn't read as "ready to merge now."

On #79's own diff:

**The on/off wire correction is the headline, and it's right.** Power is `CMD_DEVS_SWITCH` (1035) on every family, sealed by the session's key; the 6250 privacy envelope is dropped entirely. This matches the app's own frames as captured here — and independently matches what we validated end-to-end in eufy-mega (1035, signCode by key, no 6250 for power). The old family-diversion to 6250 came from the decompiled V6 parsers, and killing it removes a wire that had no level-1 form and couldn't be sent on a keyless session — the exact reason an outdoor-PT camera couldn't be turned back on.

**The observation contract generalisation is sound.** `reflects(value, ctx) -> { param, expected, observed } | undefined` replacing `expected(value)` + implicit `m.param` correctly lets the readback name the param the device *reports* (measured: 5 cameras report the enablement param and never 2001, 3 the reverse, none both) rather than the one written — so a write to a param the device never echoes dispatches unobserved instead of timing out. Separating raw `expected` from decoded `observed` is the fix for a disable-bit reporting 0 on a property that reads true, which is why the value converged but the transition event never fired. `undefined` → dispatch-unobserved is the safe fall-through.

**Encryption level by key, not topology** — `sendBySessionLevel` keying on `session.hasLevel2Key` matches the capture (every peer used one seal per session; topology mispredicted the two keyed own-session cameras). The bounded one-shot `repromptLevel2Key` re-ask is gated to non-soft callers only. Both consistent with the #72 grace ownership.

112/112 in the affected specs locally (camera, level2-grace, level2-wait-ownership, param-report, shared-live-start-failure); mutation coverage described is credible. No device data in the diff. Restraint on the 6250 alias (power moved in the same step, so it can't be attributed) and on the up-front video refusal (would withhold video from an ON camera) is the right call.

Merge order: land #76 (after its two blockers are fixed), then this. 👍

PR #81, opened 2026-08-25
feat(clean): the scene list off DP 180, and the map id it settles

martijnpoppen:
Stacked on #78, which is stacked on #74. Review those first — this diff is two commits.

The second commit is the feature. **The first is a bug in a test double that every capability spec on this line depends on**, and is the part most worth a reviewer's attention.

## What changed

| # | Commit | Item |
|---|--------|------|
| 1 | `byteCodec` varint range | test-double correctness — it disagreed with the shipped codec past 2³¹ |
| 2 | Scene list read | **B2's read half**, and it corrects what the plan expected of B3 |

## Why

The starting point was a plan error worth stating plainly: **all four messages the plan filed as "the field numbers inside them are unrecorded" are in the vendor proto set** — `DeviceInfo`, `MultiMapsManage`, `SceneResponse`, `StationRequest`. Same story as `timing.proto` on #78. The evidence was never missing; nobody had looked.

`SceneResponse.infos` is a repeated `SceneInfo` carrying each scene's id, name, map, order, type and whether it still works. Three members mirroring the schedules trio: `sceneCount` owns DP 180, `usableSceneCount` reads the same payload through `readsFrom`, `scenes()` answers the list.

## The map id, and what B3 actually is

The plan says *"Tier B after B3 gives B1 a real `map_id`."* **B3 cannot give it one.** `multi_maps.proto` states it outright: a `MAP_GET_ALL` or `MAP_GET_ONE` response travels over **p2p**, not the data point. DP 172 carries only a method, a sequence number and a result — the acknowledgement of an operation, never the map list.

`SceneInfo.mapid` is on the DP. So is a scheduled rooms-clean's `map_id`, decoded on #78. Two DP-side reads now hand a caller the id the area-select frames need — no p2p transfer, and no default invented for it.

That does not promote B1. Its encoders stay `unverified` because no capture has shown a room-select accepted; what changes is that the argument problem is solved. **B1 waits on write evidence, not on B3.**

## The test-double bug (commit 1)

`byteCodec` stands in for the real `RawDpCodec` in every capability spec on the clean line. Its varint reader accumulated with `|=` and `<<`, and JavaScript's bitwise operators truncate to a **signed** 32-bit int — so `0xFFFFFFFE` read back as `-2` where the shipped codec, via protobufjs's `uint64()`, answers `4294967294n`.

That range is not hypothetical: the vendor writes `-2` into a `uint32` as its "no map" sentinel, and `SceneInfo.mapid` does exactly this. A decode written against the double would have been checked on bytes no device sends, and passed.

**Same class as the consumables nesting #74 fixed** — a double that agrees with the assumption it was built from proves nothing about the wire. I did not take the fix on trust either: both codecs were round-tripped side by side over the boundary values (`1, 127, 128, 300, 0x7FFFFFFF, 0x80000000, 0xFFFFFFFE, 0xFFFFFFFF`) and now agree. That comparison cannot ship as a spec — a spec under `src/model` may not import `transport/`, which is the decorrelation guard doing its job — so `proto-bytes.spec.ts` pins the double's own behaviour instead.

No existing spec depended on the mangled reading; nothing else moved.

## Evidence

Field numbers are the vendor's own — `scene.proto` and `multi_maps.proto`, cited per constant table. Four readings named because each could plausibly have gone the other way:

- **The no-map sentinel** → `undefined`, together with the proto3 zero. A scene tied to map 0 and one that said nothing are the same bytes, and "no map" is the reading that cannot aim a clean at the wrong floor.
- **An invalid scene is kept, not filtered out.** The device does not delete a scene whose map went away — it reports it unusable with a reason, and that reason is the only account of why a scene stopped working.
- **`estimatedRuntime` has no stated unit.** Every other duration on this line is seconds, which is a reason to expect seconds and not a reason to claim it. Reported as sent, named for what it is.
- **The vendor's retired `DEFAULT = 5`** invalid-reason is still named, so a device sending it does not read as a decode failure.

## Verified how

- [x] `npm run verify` is green locally — 143 files, **2955 tests** (+37)
- [x] Added or updated specs (offline, synthetic fixtures) — 15 for the scene decode, 7 for the codec double

## Unverified or deferred

- **No `SceneRequest`.** Add, delete and modify are not built: no capture has shown one accepted, and a scene's tasks are only ever sent to the robot and never reported back, so there is nothing to round-trip an encoder against.
- **B3 is not built, and now reads as out of scope for the DP path.** What DP 172 can offer is an operation acknowledgement; the map list is a p2p transfer, which lands next to D1 and D4 rather than in Tier B.
- **A12's MAC / SSID / IP are unblocked on evidence and blocked on structure.** `app_device_info.proto` gives every field number. DP 169's one owner is `dockFirmwareVersion` in `vacuum_dock`, and the one-owner rule is per *line*, so the robot's network fields would have to hang off `vacuumDock()` — the wrong object for them. Closing it means either moving DP 169's ownership (breaking a merged surface) or letting `readsFrom` name an owner in another module of the same line. That is a mechanism decision, not a decode, so it is not made here.
- **`sceneCount` warns once per update**, as `scheduleCount` does — a `type: "number"` member over a base64 wire trips `coerceByType`'s "not numeric — keeping raw" log before the getter decodes it. Pre-existing across every Raw-DP member on this line; still worth a follow-up for the whole line rather than a special case.

---


max246 (approved):
Ready, nothing to address. Stacked on #78 → #74 (both already approved on their merits), so the only gate is landing the stack in order — no buggy base in this chain, unlike #79.

**The test-double fix (commit 1) is the important part and it's correct.** `byteCodec`'s varint reader now accumulates as `BigInt` instead of `|=`/`<<`, so `0xFFFFFFFE` reads back `4294967294n` — matching the shipped codec's protobufjs `uint64()` — where the old signed-32-bit reader answered `-2`. The encoder switched to division/modulo for the same range safety, matching #74's `encodeVarint`. This is the same bug-class as the #74 consumables nesting: a double that agrees with the assumption it was built from validates a decode against bytes no device sends. Pinning the double's own behaviour in `proto-bytes.spec.ts` (rather than importing `transport/` across the decorrelation guard) is the right way to lock it.

**And it's load-bearing for commit 2.** `SceneInfo.mapid`'s no-map sentinel is exactly `-2` written into a `uint32`; `sceneOf` reads it via `Number(hit.value)` → `4294967294` and checks `=== NO_MAP (0xFFFFFFFE)`. Under the old double that value would have been `-2`, slipped past the check, and shipped a decode that aims a clean at the wrong floor. The fix and the feature are correctly coupled.

**Scene decode (commit 2) is sound:**
- `each()` walks the repeated `infos` via `flatMap`, same correct handling as #78's schedules.
- No-map sentinel AND proto3 zero both → `mapId: undefined` — the reading that can't misaim a clean.
- An invalid scene is kept with its `invalidReason`, not filtered — the only account of why a scene stopped working; the retired `DEFAULT = 5` reason is still named so it doesn't read as a decode failure.
- `estimatedRuntime` reported as-sent (0 → undefined), not claimed to be seconds it never stated.
- `undefined` (unreadable) vs `[]` (genuinely no scenes) kept distinct.

Good correction to the plan, too: DP 172 carries only an operation ack, so the map list is a p2p transfer — `SceneInfo.mapid` (and #78's rooms-clean `map_id`) are what actually hand a caller the id the area-select frames need, without promoting B1 (which correctly still waits on write evidence).

154/154 in the affected specs locally; no device data in the diff. Merge order: #74 → #78 → #81. 👍

PR #82, opened 2026-08-25
feat(clean): A12's network fields, via a readsFrom that can name another capability

martijnpoppen:
Stacked on #81 → #78 → #74. Review those first; this diff is two commits.

The second one **adds a form to the member contract**, which is the part most worth a reviewer's attention — it is the only change in this stack that touches how capabilities are declared rather than what they decode.

## What changed

| # | Commit | Item |
|---|--------|------|
| 1 | `PropertySpec.raw` | the Raw-DP log noise flagged as a follow-up on #78 and #81 |
| 2 | `DeviceInfo` robot fields | **A12 complete** — MAC, SSID, IP, hardware revision |

## Why A12 was stuck, and what actually unstuck it

The plan filed A12's network fields as *"field numbers within `DeviceInfo` not recorded"*. They are recorded — `app_device_info.proto` has every one. Same mistake as D2, which this plan sent to hardware while `timing.proto` sat in the same directory.

The real blocker was structural. `DeviceInfo` is **one message spanning two capabilities**: the dock's firmware version at field 11, the robot's MAC, SSID, IP and hardware revision beside it. The one-owner rule is per product **line**, so DP 169 has exactly one owner — `vacuumDock().dockFirmwareVersion`, first to claim it and already shipped. With nowhere else to put them, the robot's network facts would have had to hang off the dock object.

Three ways out, and why this one:

| Option | Cost |
|---|---|
| Put them on `vacuumDock()` | `vacuumDock().wifiIp` — the robot's IP filed under the wrong thing |
| Move DP 169 to `vacuum_clean` wholesale | breaks a merged surface, and puts a dock reading on the clean object |
| **Let `readsFrom` name an owner in another module** | one new declaration form; both objects stay correct |

`readsFrom` now takes `{ property, param }` alongside the existing sibling-name string. It works because `Device` keys state by **name** in one flat namespace shared across capabilities — the mechanism was already there, only the way to say it was missing. Both halves are stated rather than looked up, since a member cannot see the other module's table.

**The guard is what makes that safe.** `property-id-integrity` checks the pair against the line's real owner and rejects a property owned by nothing, by two modules, on a different param, or by the *same* module (name the sibling instead). A rename or a repointed id on the far side fails the build rather than leaving these getters reading a name nothing is stored under — `undefined` forever and silently, which is the failure the whole guard exists to catch.

`borrowedBy()` is now the single resolution both forms go through, called by `bindMembers` and by all three guards that police this mechanism, so the declaration and its checks cannot drift. Widening the type is what found those three guards: each failed to compile. That is the outcome worth having, and the reason to widen the type rather than add a second field.

## The log-noise fix (commit 1)

Flagged as a follow-up on both #78 and #81, so here it is. Every structured payload on this line is declared by what its getter **answers** — nine consumable counters are `type: "number"` over one base64 string, because a caller asking for hours gets a number. Ingest saw the mismatch and logged `declared number but wire value is not numeric — keeping raw`, once per property per push. Ten Raw DPs, ten warnings a time, which is how a warning that means something goes unread.

The value was never wrong: a base64 string cannot be coerced to a number, so it already passed through intact. Only the complaint was wrong. `PropertySpec.raw` marks a payload-carrying property and `coerce` keeps the pass-through while dropping the warning.

**Deliberately narrow.** The obvious larger fix — give such specs an identity decode so storage skips coercion entirely — changes behaviour for members whose wire is sometimes a plain scalar: `suction.level` decodes only `typeof raw === "number"`, so a string-valued family would start reading `undefined`. Suppressing a log line does not justify that risk. The warning still fires for a genuinely misdeclared scalar, and a test pins both halves.

## Evidence

`app_device_info.proto`, cited per constant. Four fields read: `device_mac` (3), `hardware` (5), `wifi_name` (6), `wifi_ip` (7).

**`last_user_id` (8) and `video_sn` (2) are in the message and deliberately unread** — an account id, and a serial the caller already has by another name. Same reasoning as `TimerInfo.Addition` in the schedules read on #78: neither is a value to hand out for the sake of completeness.

An absent or empty string field answers `undefined` rather than `""` — proto3 omits an empty string, so "did not say" and "said nothing" are the same bytes, and an empty SSID is not a network name.

## Verified how

- [x] `npm run verify` is green locally — 144 files, **2980 tests** (+25)
- [x] Added or updated specs (offline, synthetic fixtures) — 6 for the cross-capability read, 6 for the storage flag, plus the three widened guards

## Unverified or deferred

- **Read-only, and not for want of a capture.** DP 169's downlink is `AppInfo` — the phone describing itself to the robot — so there is no write direction for any of this to have.
- **These four reads are as current as the robot's last report.** `DeviceInfo` is sent when it comes online and again when its IP or the account changes, so the values are absent on a robot that has not reconnected since binding, and stale in the way any cached fact is. Stated in the member docs rather than papered over with a refresh this SDK has no way to trigger.
- **`hardwareVersion` has no scale.** Useful for telling two builds of one model apart, not for comparing models.
- **The new `readsFrom` form has exactly one user.** It is a mechanism change made for one DP, which is worth a reviewer's judgement — the alternative was one of the two costs in the table above, and both are harder to undo than this is.

**No real device data anywhere.** The fixture MAC, SSID and IP are `00:00:00:00:00:00`, `example-ssid` and a TEST-NET-1 address.

---


max246 (approved):
Ready, nothing to address. Stacked on the healthy #81 → #78 → #74 chain (all approved), so the only gate is merge order — no buggy base here.

**The cross-capability `readsFrom` (commit 2) is the mechanism change, and it's done safely.** Widening `readsFrom` to `string | { property, param }` is the minimal way to let `vacuum_clean` read the robot's MAC/SSID/IP/hardware out of `DeviceInfo` (DP 169), whose one owner is `vacuumDock().dockFirmwareVersion` — the alternatives (robot IP under `vacuumDock()`, or moving DP 169 wholesale) are both worse and harder to undo. What makes it safe:
- **`borrowedBy()` is the single resolution** both `bindMembers` and all three guards go through, so the declaration and its checks can't drift — and widening the type is what surfaced those three guards (each failed to compile), which is the right way to find them.
- **The evidence gate keys on the borrowed param** (`ctx.paramIds.has(from.param)`), so a borrowing member in another module installs exactly when DP 169 was reported.
- **The integrity guard checks the `{property, param}` pair against the LINE's real owner** and rejects all four ways it can be wrong: owned by nothing, by two modules, on a different param, or by the same module (name the sibling instead). A far-side rename or repoint fails the build rather than leaving these getters reading a name nothing is stored under.
- The borrowing members import `VACUUM_DOCK_INFO_SOURCE` from the owner module rather than hardcoding `"dockFirmwareVersion"`, so the far-side name lives in one place — and the guard still verifies it.

The reads themselves are correct: the robot fields are top-level (3/5/6/7), distinct from the dock's nested `station.software` (11), so no collision; an absent/empty string answers `undefined` not `""`; and `last_user_id` (8) + `video_sn` (2) are deliberately unread — same PII restraint as `TimerInfo.Addition`.

**The log-noise fix (commit 1) is appropriately narrow.** `PropertySpec.raw` (set for any member with a getter-side `decode`) keeps `coerceByType`'s pass-through for a base64 payload while dropping the per-push "not numeric — keeping raw" warning. It deliberately does NOT add an identity decode, which would break members like `suction.level` whose wire is sometimes a plain scalar — right call, and a test pins that the warning still fires for a genuinely misdeclared scalar.

The one-user mechanism change is a fair reviewer judgement call, and I think it lands the right side of it: it's a smaller, more reversible change than either structural alternative, and the guard contains the risk. 764/764 in the affected specs locally; fixture data is `00:00:00:00:00:00` / `example-ssid` / TEST-NET-1. Merge order: #74 → #78 → #81 → #82. 👍

PR #83, opened 2026-08-25
refactor(clean): one DP table for the Tuya line, and C2's answer recorded

martijnpoppen:
Stacked on #82 → #81 → #78 → #74. One commit, no behaviour change.

## What changed

C2 asked for a diff of `jeppesens/eufy-clean`'s `LEGACY_DPS_MAP` against this SDK's coverage. **Nothing came back to port.** What the diff surfaced instead was drift on our side, and that is what this fixes.

The Tuya clean line was spelled in three places:

| Where | What |
|---|---|
| `LEGACY_VACUUM_DP` | four ids, grounded on "protocol inspection" |
| `TUYA_VACUUM_DP` | the same four among twenty, grounded on the product schema and a live read — with three entries carrying notes saying which `LEGACY_VACUUM_DP` key they duplicated |
| `locate.ts` | DP 103, named privately, in neither table |

The weaker table is gone, its two `readAliases` repointed to the identical ids in the survivor, and `look_for_sweeper` joins the table so `locate` cites it rather than respelling it.

## Why

Two tables for one product line is how they come to disagree while each stays individually plausible — the same failure the `members` table exists to prevent, a level down. Three of `TUYA_VACUUM_DP`'s entries already carried "Shared with `LEGACY_VACUUM_DP.X`" notes, which is the duplication being acknowledged in prose rather than removed.

**No id, meaning or behaviour changes.** The removed table's four entries are the same four numbers with the same meanings; `106` was `ERROR_CODE` there and is `FAULT_REPORT` here, one DP, one reading. The tests are unchanged except for the two references that had to move.

## Evidence

The diff, which is the deliverable and is now a test rather than a paragraph:

- **All nine of `LEGACY_DPS_MAP`'s ids** — 2, 3, 5, 15, 101, 102, 103, 104, 106 — are in `TUYA_VACUUM_DP`, live-confirmed on a real X8 Pro through `thing.m.device.dp.get`, and each carries its enum value set, which that map does not. A test asserts the nine are covered, so the comparison fails rather than going stale if one is ever dropped.
- **DP 103 was already grounded by us**, not by the third party: `look_for_sweeper`, `observed: true` in the clean param dictionary, and read by `locate` as a Tuya alias since before this. It was only ever missing from the table.
- **`SCALAR_DPS` is deliberately untouched**, and now has a guard saying so. It is a different device class reusing these numbers for unrelated things — a brush-detangle trigger on 153 where this line has the work status, a schedule blob on 151 where this line has power. It can only be told apart by value SHAPE, never by DP number, so a test asserts none of its nine scalar-only ids leak into the Tuya table. That boundary is what C1 will have to reckon with.

## Verified how

- [x] `npm run verify` is green locally — 144 files, **2984 tests** (+4)
- [ ] Ran against a real device — **no**, and none is needed: no wire, id or dispatch changes. The ids involved were live-confirmed on an X8 Pro in earlier work.
- [x] Added or updated specs (offline, synthetic fixtures) — 4, pinning the diff and the class boundary

## Unverified or deferred

- **Not public API.** Both tables are `@internal` and neither appears in the built ESM surface (241 exports, verified by the ESM smoke), so nothing outside this repo could name the one that went. Not marked breaking for that reason.
- **Write direction for the Tuya line stays unverified**, unchanged by this. `TUYA_VACUUM_DP`'s header still says so: the schema's `writable: true` is not a captured write, and no `publishDps` capture has been made.
- **C1 is not started and is not this.** The scalar device class is a second protocol over the same transport, its evidence is one third-party capture of one device, and per Tier C it would ship behind `unverified` and credited — which also means it cannot be verified without a G50-class robot to point at.

---


max246 (approved):
Ready, nothing to address — a clean, genuinely-no-op consolidation. Stacked on the healthy #82 → #81 → #78 → #74 chain, so the only gate is merge order.

**Verified the repoint is a true no-op.** I checked each removed `LEGACY_VACUUM_DP` id against its `TUYA_VACUUM_DP` replacement: `PLAY_PAUSE 2`, `GO_HOME 101`, `BATTERY_LEVEL 104`, and `ERROR_CODE 106 → FAULT_REPORT 106` (rename, same DP) — every id identical, so no alias or dispatch moved to a different number. `locate`'s private `103` becomes `TUYA_VACUUM_DP.LOOK_FOR_SWEEPER` (still 103). This is the exact failure a delete-and-repoint refactor risks (a reference silently landing on a different number), and it doesn't happen here.

**The consolidation is the right call.** Two tables for one product line is how they drift while each stays individually plausible — the same reason the `members` table exists, a level down; three `TUYA_VACUUM_DP` entries already carried "Shared with LEGACY_VACUUM_DP.X" prose acknowledging the duplication. Both tables are `@internal` and absent from the ESM surface, so dropping one isn't breaking.

**The tests keep it honest:** the cover-the-9 assertion pins that all of `LEGACY_DPS_MAP`'s ids (2,3,5,15,101,102,103,104,106) are in the survivor (so the comparison fails rather than going stale if one is dropped), `LOOK_FOR_SWEEPER === 103`, and the `SCALAR_DPS` no-leak guard pins the class boundary — a different device class reusing these numbers (work-status on 153 vs a detangle trigger, etc.), tellable only by value shape, which is the boundary C1 will have to reckon with.

Good outcome on the C2 diff itself: nothing to port from the third-party map — it surfaced our own drift, which is what got fixed. 141/141 in the spec locally; no wire/id/behaviour change. Merge order: #74 → #78 → #81 → #82 → #83. 👍

PR #85, opened 2026-08-26
fix(clean)!: the five reads a coverage audit found missing, and one that was wrong

martijnpoppen:
One commit, onto `beta-0.1.0` now that #74/#78/#81/#82/#83 have all merged.

This came out of a deliberate coverage audit: every entity `jeppesens/eufy-clean` exposes, checked against this SDK, with each claim verified against the vendor protos and `eufy_decompiled` rather than taken from that integration.

**Three of the six findings are in items this plan had already marked done.** That is the part worth a reviewer's attention — not the code, which is ordinary.

## The bug

```proto
Numerical dust_full_remind = 8;   // 尘满提醒的时长，单位：分钟
```

`dustFullRemind` declared `type: "bool"` and described itself as "whether the robot warns when its dust bag is full". It is a **duration in minutes**.

`Numerical { uint32 value = 1 }` and `Switch { bool value = 1 }` are the same two bytes on the wire, so `decodeUnisetting` accepted it and reported a thirty-minute setting as `true`. Nothing errored. The on/off reading was accidentally right — `0` is off — and the number was gone.

Replaced by `dustFullRemindMinutes`. A test keeps the old reading beside the new one on the same bytes, as the demonstration:

```ts
expect(decodeUnisettingNumber(payload, byteCodec, DUST_FULL_REMIND)).toBe(30);
expect(decodeUnisetting(payload, byteCodec, DUST_FULL_REMIND)).toBe(true);   // what it used to say
```

Same class as the DP 162 fix on #78 and the consumables nesting on #74: a plausible read that cannot look wrong, because two shapes agree on a field number.

## The gaps

| Read | Vendor field | Plan status |
|---|---|---|
| `chargeState` | `WorkStatus.charging` (3) | **A1 promised it** |
| `triggerSource` | `WorkStatus.trigger.source` (20) | **A1 promised it** |
| `petMode` | `UnisettingResponse` 15 | **A9 named it** |
| `poopAvoidance` | `UnisettingResponse` 14 | not named |
| `wifiSignal` | `ap_signal_strength` (11) | not named |
| `doNotDisturbStart` / `End` | `Undisturbed.begin`(2)/`end`(3) | not named |
| `resetConsumable` | `ConsumableRequest.reset_types` | **A7 was "read AND reset"** |

Details that decided a reading:

- **`chargeState` answers `undefined` while not charging.** The vendor omits the whole `charging` message rather than sending a "no", so absence *is* the reading. It also separates `ABNORMAL` — contacts touching, nothing flowing — from the contented `"docked"` that `activity` reports for exactly that state.
- **`wifiSignal` is a bare `uint32` at the top level**, the only field of `UnisettingResponse` that is not a wrapper, which is why neither existing reader reached it. Until now an X10 reported no signal at all, since `rssi` is the Tuya line's DP 134. Reported as the vendor states it, 0–100%. eufy-clean converts it with `(value / 2) - 100` to look like dBm — a plausible figure with no basis in anything the device sends, so it is not copied.
- **The DND window is one `"HH:MM"` string per end.** Four numbers for one window is four things a host has to reassemble, and the pieces are meaningless apart. `undefined` means no window configured; `"00:00"` is midnight and real, and has a test.

## The reset numbering trap

`ConsumableRequest.Type` is **not** `ConsumableRuntime`'s field numbers:

```
report:   side_brush = 1  …  dirty_watertank = 10, dirty_waterfilter = 11   (8 and 9 unused)
request:  SIDE_BRUSH = 0  …  DIRTY_WATERTANK = 7,  DIRTY_WATERFILTER = 8    (no gap)
```

Nine parts, two schemes, one message pair — the same trap `UNISETTING_FIELD` carries. Borrowing the reader's table would reset the wrong part, silently, because the device never says which it cleared. A test asserts the whole mapping rather than spot-checking, since the two tables agreeing by accident on one part is exactly how this would go unnoticed.

## Breaking

`dustFullRemind` is gone. It answered a boolean for a field that is not one, so nothing reading it was getting the value.

## Evidence

Vendor protos throughout — `unisetting.proto`, `undisturbed.proto`, `work_status.proto`, `consumable.proto`, `common.proto` — cited per constant. The reset is additionally corroborated in the decompile: `DeviceHomeModule.resetAccessory(deviceId, int accessory, Callback)` calling through to `resetAccessories`, taking exactly this kind of integer part id.

**Deliberately not built, and why:** eufy-clean also exposes **off-peak charging** (a switch and two times). Its own source says *"Field 23 is not in the compiled proto, so we parse the raw bytes directly."* That is their observation of a device, not the vendor's schema — Tier C by R5, so it is recorded as a candidate rather than built as if grounded. **Auto-Return**, **Activity Log** and **Cleaning Pattern** are `SCALAR_DPS` on the G50-class line, which is C1.

## Verified how

- [x] `npm run verify` is green locally — 144 files, **3029 tests** (+45)
- [x] Ran against a real device — **no.** Offline, against synthetic byte fixtures built from the vendor schema.
- [x] Added or updated specs (offline, synthetic fixtures) — 22, including the numbering trap and the old-vs-new reading of `dust_full_remind`

## Unverified or deferred

- **`resetConsumable` ships declared and not installed**, like the other seventeen. The message, field and enum are the vendor's and the app has the feature, but no capture has shown the frame accepted — and an AIoT DP write is fire-and-forget, so a wrong one would look like success while discarding service history the device never recomputes.
- **Two `UnisettingResponse` fields are still unread**, now the only ones: `unistate`(10) and `wifi_data`(12). `Unistate` is worth a follow-up — it carries `mop_holder_state_l/r`, `mop_state` and `map_valid`, which is the AIoT answer to the mop-pad sensor this SDK only has on the Tuya line. Left out of this PR to keep it to the audit's agreed scope rather than growing it mid-flight.
- **No write for any of the new reads.** DND times, pet mode and poop avoidance are all settable in the app; none has a captured frame.

---


max246 (approved):
Ready, nothing to address. Straight onto `beta-0.1.0` now the stack has merged — no dependency gate. Verified every finding against its test.

**The bug fix is right and demonstrated.** `dust_full_remind` is a `Numerical` (minutes), not a `Switch`, and since both wrap `value = 1` the old `type: "bool"` reader accepted it and reported 30 minutes as `true`. `dustFullRemindMinutes` reads it as the number via `decodeUnisettingNumber`, and the test keeps both readings on the same bytes (`… Number → 30`, `… Unisetting → true`) as the proof — same class as the DP 162 fix (#78) and the consumables nesting (#74). Breaking removal of `dustFullRemind` is justified: it never returned the real value.

**The reset numbering trap is handled correctly** — the sharpest thing here. `CONSUMABLE_RESET_TYPE` is the 0-based contiguous REQUEST enum (0–8), NOT `ConsumableRuntime`'s report field numbers (1–11 with 8/9 unused); a test asserts the whole `CONSUMABLE_PARTS → [0..8]` mapping rather than spot-checking, since two tables agreeing on one part by accident is exactly how a wrong reset would go unnoticed. And `resetConsumable` ships `unverified`/uninstalled (`setResetConsumable` undefined), which is right for a fire-and-forget write that would silently discard service history — the corroboration in `DeviceHomeModule.resetAccessory` is a nice extra.

**The new reads each use the correct decoder:**
- `wifiSignal` reads the bare TOP-LEVEL `uint32` (field 11) via `decodeUnisettingTopLevel`, not the wrapper reader — the test proves the wrapper reader returns `undefined` on those bytes, which is why neither existing reader reached it. Reported as the vendor states it (0–100), not eufy-clean's baseless `(v/2)-100` dBm.
- `chargeState` treats the omitted `charging` message as "not charging" (`undefined`), a present-but-empty one as `charging`, and separates `ABNORMAL` → `fault` from the `docked` that `activity` reports for the same physical state.
- DND window renders `begin`/`end` as `"HH:MM"`, `undefined` for no window, `"00:00"` real (tested).
- `triggerSource` maps the source enum with the vendor's `UNKNOWN` for an omitted value; `petMode`/`poopAvoidance` are standard wrapper bools.

Good scope discipline too — off-peak charging left as a Tier-C candidate (their raw-byte observation, not the vendor schema), `unistate`(10)/`wifi_data`(12) flagged as the only remaining unread fields rather than pulled in mid-audit. 160/160 in the spec locally; no device data (synthetic fixtures throughout). 👍

PR #86, opened 2026-08-26
feat(clean): read unistate, and leave the three fields whose polarity is unknown

martijnpoppen:
Stacked on #85. One commit.

The audit's own leftover — the last unread part of `UnisettingResponse` except `wifi_data`. **The interesting half of this PR is what it does not build.**

## What changed

`unistate`(10) is device STATE where the rest of that message is user toggles, which is why it sits a level down and the existing toggle reader never reached it.

| Read | Field | Why it earns a property |
|---|---|---|
| `hasValidMap` | `map_valid`(4), an `Active` | the precondition for every area-select frame |
| `mapLayers` | `live_map.state_bits`(6.1) | a bitfield — the vendor says the values combine |
| `cleanStrategyVersion` | `clean_strategy_version`(7) | diagnostic only |

`hasValidMap` is the one worth having: a room or zone clean sent at a robot with no usable map is a request it cannot honour, and this is the device's own answer rather than an inference from whether some scene happens to name a map. `LIVE_MAP_BITS` names the bit positions so no caller hard-codes a shift.

## What it does not build, and why

`mop_holder_state_l`(1), `mop_holder_state_r`(2) and `mop_state`(5) each carry a bool the vendor annotates 已安装或已取出 — ***installed or removed*** — without saying which boolean is which. `custom_clean_mode`(3) has the same shape and the same problem.

I went looking for the answer rather than assuming one:

- **The decompile** has `mopHolderStateL` / `mopHolderStateR` on the T2080 UI model — but only as data-class members and a `toString`. No branch anywhere settles the polarity.
- **The reference integration** does not read `unistate` at all, so it offers nothing either.

A bool whose polarity is a coin flip is the most guess-shaped value this SDK could publish, and mop presence is exactly the reading a user acts on: getting it backwards says the pad is fitted while it sits on the bench. **One capture settles all three at once** — pop a pad off and watch the bit — so they wait for that rather than for a better guess.

A test asserts no member surfaces them, so the omission reads as deliberate rather than forgotten.

## Evidence

`unisetting.proto`, cited per constant. One structural trap has its own test: **`ai_see` is field 4 at the top of the message and `map_valid` is field 4 inside `unistate`.** A reader that stepped one level would answer one for the other — which is the whole reason this decode exists separately from `decodeUnisetting`, and the test pins both readings of the same payload.

## Verified how

- [x] `npm run verify` is green locally — 144 files, **3048 tests** (+19)
- [x] Added or updated specs (offline, synthetic fixtures) — 8, including the field-4 collision and the deliberate omission

## Unverified or deferred

- **Three fields await one capture**, as above. Worth adding to the capture session already planned for the eighteen unbuilt writes: removing and refitting a mop pad is a few seconds and settles `mop_holder_state_l`, `_r` and `mop_state` together.
- **`wifi_data`(12) is now the only other unread field** of `UnisettingResponse`. It is a nested `Ap` structure carrying frequency band and connection result — worth a look, but it is network diagnostics rather than anything a host surfaces, so it is not chased here.
- **No writes.** Nothing in `unistate` is settable; it is state the robot reports.

---


max246 (approved):
Ready, nothing to address. Stacked on #85 (approved), so the only gate is merge order.

**The field-4 collision trap is handled correctly** — the reason this decode has to exist separately. `decodeUnistateFlag`/`decodeUnistateNumber` step into `unistate`(10) FIRST, then read field 4 inside it, so `map_valid` can't be confused with the top-level `ai_see`(4). The test pins both readings of one payload, which is the right way to lock a same-number-different-level trap.

**The reads are right:** `hasValidMap` reads `unistate.map_valid`'s `Active` value (the genuine precondition for an area-select, vs inferring it from whether a scene names a map); `mapLayers` reads the `unistate.live_map(6).state_bits(1)` bitfield with `LIVE_MAP_BITS` naming the positions so no caller hard-codes a shift; `cleanStrategyVersion` reads `unistate.clean_strategy_version(7)`. All three borrow DP 176 via `readsFrom: "childLock"` (the sibling owner) — correct sibling form since they're in the same module.

**The restraint is the best part.** Leaving `mop_holder_state_l`/`_r`/`mop_state`/`custom_clean_mode` unread because the vendor annotates them "installed or removed" without saying which bool is which — and confirming neither the decompile (data-only, no branch) nor the reference integration settles it — is exactly right: mop presence is a value a user acts on, and shipping a coin-flip polarity that says "pad fitted" while it's on the bench is worse than not reading it. A test asserting no member surfaces them keeps the omission deliberate. One capture (pop a pad, watch the bit) settles all three, and folding that into the already-planned write-capture session is the sensible next step.

168/168 in the spec locally; synthetic fixtures throughout; `wifi_data`(12) correctly left as the only remaining unread field (network diagnostics, not host-facing). 👍

PR #87, opened 2026-08-26
feat(clean): read the map stream off biz/…/res — hex framing, LZ4 blocks, channel routing

martijnpoppen:
Stacked on #86. Steps 1–3 of the map-stream spec — the plumbing that gets a robot's map from the broker to a decodable frame. No `stream.proto` decoding yet; that is step 4.

## What changed

**`core/raw-dp-hex.ts`** — `hexToRawDp(hex)`. The map topic sends the same `varint(len) ++ protobufMessage` frame every Raw DP carries, only hex-encoded instead of base64, so this re-encodes rather than adding a second codec.

**`core/lz4-block.ts`** — `lz4BlockDecompress(block, size)`. Pixel planes arrive LZ4-compressed with no flag saying so: the message carries the bytes and a `pixel_size`, and the two disagreeing *is* the flag. What travels is a bare LZ4 *block* — no magic number, no frame header, no checksum, no stored size — which every npm LZ4 package treats as the awkward case behind its frame API. Sixty lines, no dependency.

**`transport/mqtt/biz-stream.ts`** — `parseBizMapFrame`, `BIZ_PROTOCOL`, `BIZ_CHANNEL`, `bizChannelName`. Reads the envelope and unwraps a frame to its bytes.

**Client** — a new `mapFrame` event carrying `{ deviceSn, frame }`. Until now every message on this leg fell through the DP parsers, which correctly refuse it, and vanished.

## Why

The plan filed live maps under "needs the clean-native library, absent from the app package". That was wrong, and the spec that corrects it is [here](https://claude.ai/code/artifact/754e4950-b5d7-4a78-8485-624ccb8f4d96). The map is published to `biz/eufy_home/<model>/<sn>/res` — cloud MQTT, same broker and session as the DP channel, on a topic `subscribeTopics()` has been subscribing all along.

## Evidence

The envelope is read from the **app**, not inferred from a capture or borrowed from a third-party client:

- `aiot_sdk/device/BaseRobovacServiceImpl.dispatchMessage` — switches on `payload.protocol`: **41 is map data, 43 is a live photo**. That field is where the reference integration's unexplained "protocol-41" comes from. Also the object-or-string `payload`, and the per-device dedupe.
- `clean_native_bridge/…/aiot/AiotMapManager` — the map listener requires **all six** of `channel_id`, `clear_type`, `data_type`, `offset`, `data`, `len` and returns early without them; and it routes `channel_id == 0` to a *scene* event rather than a map one. `requestSceneList` confirms it a second time: `requestMapData(1000, 1, 0)`.
- `aiot_api/bean/AIOTMapDataBean` — every envelope field, including the paging ones.
- `thingclips/sdk/mqtt` — the `protocol` key itself.
- `subscribeBizDevice` subscribes at QoS 1 and publishes nothing, so no request is needed to start the stream.

Channel ids 1–10 are `Metadata.ChanIds` from `stream.proto`. Their trap is recorded at the table: each ChanIds **field** names a message and its **value** is the id the device chose, so pairing each name with its field number is the vendor's default numbering and not an authority — a firmware that renumbered would make the table wrong without making it look wrong. Anything that decodes a channel's bytes must prefer a device-declared mapping.

LZ4 vectors are liblz4 1.9.4's own output, not hand-rolled bytes.

## Verified how

- [x] `npm run verify` is green locally — 147 files, 3082 tests, 247 ESM exports. `docs:api` clean.
- [ ] Ran against a real device — no. Every path here is read-only and no write was added.
- [x] Added or updated specs (offline, synthetic fixtures) — 34 new tests across the three modules. Serials are `T8000P0000000000`.
- [x] Docs updated — module docs carry the reasoning; no guide page needed until there is a surface.

Two checks beyond the committed specs, both run once and discarded:

- **Framing.** Built a frame with the shipped `rawDp` writer, re-encoded it as hex, read it back through the shipped `rawDpCodec` — identical field for field. Can't live in the tree: `core/` sits below `transport/`.
- **LZ4 agreement.** 800 blocks from liblz4 (fast and HC modes, sizes 0–20 000, five data shapes) through this decompressor: zero mismatches.

Three deliberate refusals worth a reviewer's eye, each guarding a failure that would otherwise be silent:

1. **Hex validates.** `Buffer.from(s, "hex")` never throws — it decodes until it meets a non-hex character and returns what it got. For base64 that is harmless because the length prefix won't match a shortened body. Hex truncates from the END, so `"02" + "aabb" + <garbage>` yields a prefix of 2 followed by exactly two bytes, **passes** the length check, and hands back a message with the rest quietly gone.
2. **LZ4 must reach `size` exactly.** LZ4 has no checksum and no terminator, so a truncated block doesn't fail — it stops early. A lenient reader returns a half-drawn map that looks entirely real. The reference implementation does exactly that.
3. **A frame short a field is refused whole.** Same rule the app applies.

## Unverified or deferred

- **`offset` / `len` semantics.** A large map is split across frames — new information, and the reference integration doesn't account for it. Whether `len` is this chunk's length or the whole message's, and whether `offset` counts bytes or chunks, is decided in the app's React Native layer, which ships as Hermes bytecode. Both are carried **verbatim and uninterpreted**. Step 4 must treat a frame with a non-zero `offset` as a fragment, not a message.
- **Whether `Metadata` ever arrives.** The string `ChanIds` does not appear anywhere in the decompiled app either, so whatever consumes the mapping lives in the same Hermes bundle. Until a capture says otherwise, `BIZ_CHANNEL` is a documented default, not a declared mapping.
- **A second, PULL route exists** — `requestMapData(num, page, channel_id)` paged 1000 at a time, and `requestMultiMapData(int[] channel_ids)`. Neither the plan nor the spec's first pass knew about them; they may be the honest answer to the bulk enumeration filed under P2P. Not touched here.
- **No writes.** Nothing in this PR sends anything to a device.

---


max246 (approved):
Ready — approved. Base is `beta-0.1.0` (#86 already merged), so no stack gate; and this is the base #88 needs, so landing it unblocks the map chain.

Traced the three modules against their tests; the plumbing is solid and every "silent wrong answer" refusal is real.

**LZ4 block decompressor — correct, and this is the part most worth scrutiny.** I walked the whole loop: the extend-nibble chain (saturate at 15, add bytes until one is < 255, `undefined` off-end), the three overrun guards (`pos+literals>end`, `written+literals>size`, `written+length>size`), the `offset===0 || offset>written` back-reference guard, the byte-at-a-time overlap copy (offset 1 / length 200 = RLE, which is why it can't be a block copy), and the closing `written === size ? out : undefined`. That exact-size check is the load-bearing one — LZ4 has no checksum or terminator, so without it a truncated block hands back a plausible short map; the reference impl does exactly that. `MAX_OUTPUT` caps a hostile `size` before allocation. 800 liblz4 vectors run clean (discarded, since `core/` can't import liblz4) plus committed spec vectors.

**`hexToRawDp` — the one thing the base64 codec genuinely can't catch.** The guard `length===0 || length%2 || !/^[0-9a-fA-F]+$/` runs before `Buffer.from(hex,"hex")`, which otherwise silently decodes up to the first non-hex char and drops a trailing half-byte — and because hex truncates from the END, `"02"+"aabb"+garbage` yields a self-consistent length-2 frame that passes the codec's length check. Reusing `RawDpCodec` for the framing rather than writing a second codec is the right call.

**`parseBizMapFrame` — refusal #3 is enforced, not just documented.** Object-or-string payload handled (matches the app's `dispatchMessage`), `protocol !== 41` refused (43 live-photo goes elsewhere), and all six of `channel_id`/`clear_type`/`data_type`/`offset`/`len`/`data` required — any missing → `undefined`. `count()` takes number-or-numeric-string, safe-integer ≥ 0.

**The evidence discipline is exemplary** — the envelope is read from the app (`BaseRobovacServiceImpl.dispatchMessage`, `AiotMapManager`, `AIOTMapDataBean`), not inferred, and the `Metadata.ChanIds` trap is recorded at the `BIZ_CHANNEL` table: the default numbering is the vendor's, not an authority, so a decoder must prefer a device-declared mapping. The deferred items are honest and load-bearing: `offset`/`len` carried verbatim (step 4 must treat non-zero `offset` as a fragment — which #88 does), `Metadata` arrival unconfirmed, and a second PULL route (`requestMapData`/`requestMultiMapData`) noted as possibly the real bulk-enumeration answer.

34/34 specs locally; `verify` green (3082); serials `T8000P0000000000`, LZ4 vectors liblz4 1.9.4's own; read-only throughout. Map stack: #86 ✓ → **#87** → #88. 👍

PR #88, opened 2026-08-26
feat(clean): decode the map — stream.proto messages, cell lookups, and a live map per device

martijnpoppen:
Stacked on #87. Steps 4–6 of the [map-stream spec](https://claude.ai/code/artifact/754e4950-b5d7-4a78-8485-624ccb8f4d96), which completes it: the frames #87 routes now decode, and `eufy.mapFor(sn)` hands back a live map.

## What changed

**`model/vacuum-map.ts`** — the six `stream.proto` messages (`Map`, `MapInfo`, `RoomOutline`, `RoomParams`, `RestrictedZone`, `MapDescription`), plus `MapBackup` and `DynamicData`.

**`model/map-pixels.ts`** — cell lookups over both planes, ending at `roomAtPoint`: which room is the robot in.

**`model/vacuum-map-store.ts`** — holds one device's map, assembled from pieces that arrive on five channels at their own pace.

**`client/map-channels.ts`** + `eufy.mapFor(sn)` + a `map` event.

**`model/proto-read.ts`** — lifts the five field readers three decoders had each grown their own copy of. The interesting part isn't the code but the rule each encodes (proto3 omits a zero, so absent and zero are the same bytes), and a rule spelled out in three places drifts in two. Behaviour-neutral: the existing 2140 model tests passed unchanged before anything new was added.

## Why

Structured data rather than a rendered image, per the spec's §4 decision. A host that wants a picture can paint one from this; a host given only a picture can never get back to the data — and "which room is the robot in" wants a room id, not a PNG.

## Evidence

`stream.proto`, `common.proto` and `clean_param.proto` for the schema; the decompiled app for the envelope (see #87). Where the reference integration and the protos disagree, the protos win.

Four traps, each of which produces a **confident wrong answer** rather than a failure — which is why they're called out rather than left in the diff:

1. **`Point` and `Pose` are `sint32`.** The codec reads every varint unsigned, because the wire doesn't record which of `int32`/`sint32`/`uint32` a field was declared as — only the message's author knows, and that's the decoder. Left un-zigzagged, an origin at `-150` reads as `299`, and every wall, zone and room label placed against it lands on the wrong side of the room. A shared `signed` reader undoes it, pinned in a spec against the exact inverse in core's writer — the two halves of one encoding now live in different layers with nothing else holding them together.

2. **A room's suction is not the DP 158 scale.** I started to reuse `SuctionLevel` and caught it: `clean_param.proto` declares `Fan.Suction` with five values ending at `MAX_PLUS = 4`; DP 158's scale has six, with **BoostIQ at 4** and Max Pro at 5. They agree 0–3 and diverge exactly where the live capture behind the existing clean-parameter read didn't reach — it only covered 0 and 2. Reusing it would report "BoostIQ" for a room the user set to Max+. `CleanType`, `MopMode` and `CleanExtent` *do* come from the same proto that read already uses, so those vocabularies are reused rather than re-spelled.

3. **The two planes pack differently and don't share a grid.** The map is four cells to a byte at two bits each; the room outline is one byte per cell whose low two bits are a sub-type, so `0x05` is room 1 and a reader taking the whole byte calls it room 5. Each plane also carries its own `origin` *and* `resolution`, so a cell index from one is not a cell index in the other — everything converts through world coordinates, pinned with one world position landing on cell `(3,0)` of the map and `(1,0)` of the outline. Rows count **up** with `y`; the vendor's renderer flips vertically before drawing, which is a display choice, and reading the flip back into the coordinates puts every lookup in the wrong half of the map.

4. **Pieces from different maps must not be mixed.** Every piece carries the `mapId` it belongs to and a `releases` counter that advances on each edit. A store ignoring those answers a room lookup by indexing last week's outline with this week's coordinates. So a piece naming a different map replaces everything, and an older revision of the same map is dropped. Revisions are per piece and cleared on a switch — carrying them over would drop the new map's first frames as stale.

Also: `docks_v2` wins outright over `docks` (a device sending the newer list sends the older one too, so reading both reports every dock twice, once without its kind). Plane length is checked `>=` the grid, not `==`: a longer plane is padding, a shorter one answers a lookup near its end out of whatever bytes followed.

## Verified how

- [x] `npm run verify` is green locally — 152 files, 3158 tests, 270 ESM exports. `docs:api` clean.
- [ ] Ran against a real device — no. Read-only throughout; no write path was added.
- [x] Added or updated specs (offline, synthetic fixtures) — 76 new tests. LZ4 vectors are liblz4 1.9.4's own output. All fixtures synthetic.
- [x] Docs updated — module docs carry the reasoning; no guide page until the capture settles the two unknowns below.

The channel table is pinned against `BIZ_CHANNEL` by a spec, since either side could move and the symptom would be frames silently going unread rather than anything failing.

## Unverified or deferred

**Three deliberate refusals**, each a place where decoding *something* would have been easy and wrong:

- **`MAP_INFO` (channel 1) is unrouted.** It carries geometry that names no map — the vendor retired that message's `map_id` — so a standalone one can't be stamped against what's held, and geometry attached to the wrong plane misplaces every lookup. The geometry the store uses arrives inside the `Map` frame it belongs to. `PATH`, `OBSTACLE_INFO`, `TEMPORARY_DATA` and `CRUISE_DATA` have no decoder; `SCENES` already arrives on DP 180.
- **`RestrictedZone.suggestion` is unread.** Those are zones the robot *proposed* and the user hasn't accepted. Folding them in with the real ones claims a restriction the robot isn't enforcing.
- **No invented `resolution` default.** The reference substitutes 5 without saying why; a guessed scale silently misplaces everything on the map. An unnamed room likewise stays unnamed, with `scene` + `sceneIndex` exposed so a host can build "Bedroom 2" the way the app does.

**Two things the capture still changes**, both additive to what's here:

- **`offset`/`len` reassembly.** A frame with non-zero `offset` is skipped without being decoded — whatever `offset` counts, non-zero means "not the start", and handing a fragment to a protobuf reader is how a decoder produces a confident wrong answer. The codec would reject a fragment anyway on its length prefix, but relying on that makes correctness an accident of the framing rather than a decision. **Consequence worth knowing: anything split across frames is currently dropped, so a large map may never populate.**
- **The channel numbering is the vendor's default, not a device declaration.** No `Metadata` frame has been observed, and `ChanIds` appears nowhere in the decompiled app either. A device that renumbered would be decoded wrongly and would not look wrong — which is why that risk is written at the table rather than left in a commit message.

`Map.frame` distinguishes full from incremental; how a `P` frame is applied is still unknown, so the type says so and a consumer that can't answer it should keep `"full"` frames and drop the rest.

---


max246 (approved):
Exemplary work — I traced the load-bearing decoders against their tests and every one of the four flagged traps is genuinely handled. Content approved. Merge gate is just the stack: it's on #87 (`clean-hex-frame`, still OPEN), so #87 lands first — a normal stack, not a blocked one.

**The four traps, verified:**

1. **sint32 zigzag** — `signed()` is `raw%2===0 ? raw/2 : -(raw+1)/2`, the exact inverse of core's writer `zigzag`: writer(-1)=1 and signed(1)=-1 round-trip. Reading coordinates unsigned would put an origin at -150 at +299 and misplace every wall/zone/room. Pinned against the writer.
2. **Room suction ≠ DP 158** — `ROOM_SUCTIONS` is its own 5-value enum ending at `maxPlus`(4); DP 158's scale has BoostIQ at 4 / Max Pro at 5. They agree 0–3 and diverge at 4, so reusing `SuctionLevel` would report BoostIQ for a Max+ room. Caught mid-reuse, documented, and a test pins Fan.suction 4 = Max+.
3. **Two planes, different packing, no shared grid** — the map is 4 cells/byte @ 2 bits (`(byte >> ((i&3)*2)) & 3`), the outline is 1 byte/cell with the low 2 bits a sub-type (`id = byte >> 2`, so 0x05 → room 1). Each plane carries its own origin AND resolution, so everything converts through world coordinates (`cellAtPoint`/`pointAtCell` round-trip), and the vendor's vertical display-flip is deliberately NOT read back into the coordinates. Pinned with one world position landing on (3,0) of the map and (1,0) of the outline.
4. **Piece assembly** — the store checks map-switch FIRST (a piece naming a different `mapId` replaces everything and `clear()`s revisions) then per-kind `releases` revision (an older revision is dropped), with the pose kept across a switch because it belongs to the world. Exactly the "don't index last week's outline with this week's coordinates" failure, avoided.

**Also verified:** plane length checked `>=` grid (tests both ways — a short plane refused, a long one accepted as padding, so a lookup near the end can't read trailing garbage); `docks_v2` wins over `docks`; `proto-read.ts` lifts the five shared readers so the proto3-zero rule lives once (behaviour-neutral, the existing model tests passed unchanged).

**The restraint is as good as the decoding** — MAP_INFO unrouted (names no map), suggestion zones unread (the robot isn't enforcing them), no invented `resolution` default. And the deferred items are honestly load-bearing: **`offset`/`len` reassembly isn't done, so anything split across frames is currently dropped and a large map may not populate** — that's the one operational limitation worth a consumer knowing, flagged plainly; the channel numbering being the vendor default (no `Metadata`/`ChanIds` observed) is written at the table, not buried.

76/76 in the map specs locally; `verify` green (3158); LZ4 vectors are liblz4 1.9.4's own; no device data. Merge order: #87 → #88. 👍

PR #90, opened 2026-08-26
fix(client): answer a write the device acknowledged and never applied (#77)

### LINKED ISSUE #84: Read a device's params over P2P, so confirming a write costs no cloud request
## Observation

Confirming a write costs an account-wide cloud read, and the device can answer the same question itself over the session that is already open.

`refreshEventState` polls `registry.getDevices()` in a loop until the written param reflects the expectation. That call is not one request: it is `get_house_list` plus `get_devs_list` per house plus the bare call, with no coalescing, and it returns the whole fleet each time. Measured per confirmed write, across a fleet of 8 cameras:

| camera | topology | cloud requests to confirm one write |
| --- | --- | --- |
| own session, keyed | own | 8 |
| own session, keyed | own | 16 |
| attached | via station | 4 |
| attached, a fourth device type | via station | **54, never converged** |

Concurrency makes it worse rather than better: `enqueueStateTransition` serialises per `sn:param`, so N writes to different params run N independent polling loops. A host setting six properties in one action issues roughly seventy account-list requests in a burst.

This is the reason [#77](https://github.com/mega-yfue/eufy-sdk/pull/79) turned enablement into an observed write but did **not** extend observation to the other ~25 writable members: the traffic does not scale.

## The device already answers this over P2P

The current app solicits a device's own params over the P2P session it already holds, and the reply carries the state a write needs confirming against.

Captured from the app, one session per device: **67 outbound `CMD_CAMERA_INFO` (1103) requests** across 5 peers of 4 device types and both topologies. The request body is 8 bytes and, decrypted, is constant:

```
[u32 LE 255]  [u32 LE <selector>]
```

with exactly two selector values ever observed — `429` (`0x01AD`) and `941` (`0x03AD`) — always both, always in that order, per device. 63 of the 67 decoded (the other 4 went to a peer whose gateway reply was not in the capture, not a different shape).

The frame the app puts them in matches primitives this SDK already has, byte for byte:

| seal | app's frame header | existing primitive |
| --- | --- | --- |
| level-1 | `magic=0100 ch=0 sign=1 declared=16` | `session.sendIntStringCommand(1103, selector, 255, "", 0)` |
| level-2 | `magic=0800 ch=0 sign=8 declared=40` | `session.sendRawLevel2Bytes(body, 0, 1103, 8)` |

At level-1 the 8-byte body is zero-padded to one ECB block, which is what `buildIntStringCommandPayload` with an empty string value already produces. Nothing new has to be built on the wire.

The reply arrives as `CMD_CAMERA_INFO` with a **root-level** `params` array — the shape #79 already decodes (a station's notify nests its array under `payload`; a camera's own report does not, and each id is read only in its measured shape). Replies are large and span many datagrams: declared lengths of 515, 10132 and 36427 bytes were observed, so this needs the session's reassembly, not a per-datagram read.

## Measured against the fleet

Driving the app's own request through the existing primitives, read-only:

| camera | topology | seal | cloud requests | params returned | carries enablement |
| --- | --- | --- | --- | --- | --- |
| outdoor-PT | own session | **level-2** | **0** | **104** | **yes** |
| battery cam | attached | **level-2** | **0** | **147** | **yes** |
| indoor-PT | own session | level-1 | 0 | 8 | no |
| wired cam | own session | level-1 | 0 | 0 | no |

A station's single reply covered **four attached cameras at once** — in the capture, 599 params with four separate enablement entries. One round trip confirms a whole attached group.

## What this changes

- **Confirmation becomes local.** No cloud request, no account-wide payload, no rate-limit exposure, and it works on the LAN.
- **It scales to every member.** 104–599 params per reply is not an enablement-specific channel; it is the device's state. The cost argument that blocked generalising observation disappears.
- **Cheaper than the alternative.** A secure-MQTT push would also avoid the polling, but it is still a cloud round trip with a credential and a broker dependency. This is the session the SDK already owns.

## What is NOT established

Three gaps, to be declared rather than inferred:

1. **What selector `429` returns.** Every decodable reply in the capture answered `941`. `429` is observed being sent to every peer but no reply was attributable to it.
2. **Why level-1 answers so little.** Both unkeyed sessions returned 8 params or nothing, with no enablement. Whether `941` is simply the wrong selector at level-1, or those devices report less, or the request needs a different shape there, is unknown. Consequence: **a session with no level-2 key still needs the cloud readback**, including the camera from #77.
3. **The semantics of `255` and `941`.** Both are constants observed in every sample. Per the precedent of `buildLookupWithKeyPayload` (which reproduces `02 05 01 05` verbatim with its meaning declared unconfirmed), they should be reproduced verbatim and documented as unexplained — not named as though their meaning were known.

## Proposal

Make the readback ask the device first and fall back to the cloud:

1. A P2P **param read** on the session — one request, the reply landing through the existing `applyRealtimeState` path so the device's params update the live model exactly as a volunteered report does.
2. `refreshEventState` tries it when the session holds a level-2 key, and keeps the existing cloud loop when it does not. The cloud path is the fallback, never removed.
3. Because the request rides the same session, it benefits from the level-2 re-prompt added in #79: a session whose one-shot negotiation had settled without a key recovers instead of being stuck on the cloud path forever. Measured during this investigation — a camera the capture shows using level-2 fell back to level-1 here purely because its negotiation had already concluded.

### Acceptance

- [ ] A param read is issued on the session, sealed by the same rule as every other write (`sendBySessionLevel`), not by topology.
- [ ] Its reply updates the live device through the existing realtime path; no second state owner.
- [ ] Confirming a write on a keyed session costs **zero** cloud requests, asserted by a test that counts them.
- [ ] An unkeyed session still confirms through the cloud, asserted rather than assumed.
- [ ] Reassembly is exercised: a reply larger than one datagram must land whole, since the observed replies reach tens of kilobytes.
- [ ] The two constants are reproduced verbatim with their semantics documented as unconfirmed, and `429` is not sent until something is known about it.
- [ ] Live: a keyed own-session camera and an attached camera each confirm a write with no cloud request; an unkeyed camera still converges.

### Out of scope

Generalising `observation` to the other writable members. That becomes affordable once this lands, but each member's readback param and expected value is its own evidence question — see `reflects()` in #79.

## Provenance

The wire facts come from a capture of the current app's own frames against a live account, decrypted locally: level-1 with the derivable per-station key, level-2 by deriving the session key from the gateway reply present in the capture. No device identity, address, account id or key material is reproduced here.

### LINKED ISSUE #1052: Present a switched-off camera as disabled in HomeKit, instead of refusing the stream it offers
## Parent

#981

## Observation

A camera that is switched off is refused, but HomeKit is never told, so Apple Home keeps asking.

The streaming adapter already gates on enablement, and correctly — its contracts pin `refuses a live session while the admitted enabled observation says the camera is disabled`, `refuses a start for a camera observed disabled after its session was prepared`, and `terminates an active session and stops its media when the camera is later observed disabled`. That is the right layer: HomeKit representation and product policy are the plugin's, and the SDK reports the evidence.

What is missing is the other direction. The camera services this plugin publishes are `Lightbulb`, `Microphone`, `Speaker` and `Switch` — there is no Camera Operating Mode service, so no `ManuallyDisabled` and no `HomeKitCameraActive`. A disabled camera therefore looks live in Apple Home: the tile is offered, a tap starts a stream request, and the plugin refuses it. The user sees a camera that fails rather than a camera that is off.

Two smaller gaps sit beside it:

- **The enablement change is not consumed.** The runtime subscribes `connect`, `disconnect`, `deviceAdded`, `deviceRemoved`, `deviceCapabilities`, `error` and the catch-all `event`, but nothing acts on an enablement change by name. The gate therefore learns the state on a HomeKit `onGet` rather than when it moves.
- **The failure reason is generic.** When a disabled camera is asked for media anyway — a session prepared before the state was observed, or a stale reading — the reported reason does not say the camera is off.

## Why now

Both halves of this were unsafe to build until the SDK's enablement value could be trusted, and it now can (mega-yfue/eufy-sdk#79):

- **It could lie.** On the outdoor-PT / S350 / mini families the SDK wrote the privacy envelope while reading the on/off param, so `enabled` could report the opposite of reality — that member was named in `unreflectedMembers` for exactly that reason. Those families now write the param they read, and the entry is gone.
- **It froze.** Nothing pushes enablement, so after any write the reported value never converged — measured on one bound object, a successful `setEnabled` left the reading unchanged at 0s, 2s, 7s and 22s while the camera was on and streaming. It now converges in 2–6 s and emits a change event exactly once when the write lands.
- **A new start stage names the signature.** `audio-only` distinguishes a source that is silent from one delivering sound and no picture, which together with a `false` enablement reading is what a switched-off camera looks like.

Publishing a wrong "off" state would withdraw a working camera from a user's home, so the value's reliability is the precondition, not a detail.

## What to build

1. **Publish the operating mode.** Add the Camera Operating Mode service to an admitted camera and drive its disabled characteristic from the SDK's enablement observation, so Apple Home presents the camera as off instead of offering a stream that will be refused. Keep the existing refusal: the characteristic is presentation, the gate is policy, and neither replaces the other.
2. **Update it when the state moves**, from the SDK's enablement change event, rather than only when HomeKit reads. The event fires once per landed write.
3. **Name the reason.** When media is refused or a start fails on a camera whose enablement reads false, report a bounded diagnostic that says the camera is off, distinct from a transport failure. The `audio-only` stage is the SDK-side corroboration where a reading is stale.

## Evidence gate

Enablement is only trustworthy where the SDK says it is. Two constraints, both already expressed by the SDK surface:

- Only act on an **exactly evidenced boolean** — an absent reading is not a claim that a camera is off, and the existing contract already pins that.
- Honour `unreflectedMembers`. It is empty for enablement after mega-yfue/eufy-sdk#79, but a future family whose read stops tracking its write would reappear there, and this feature must decline rather than publish a value the SDK does not trust.

There is also a residual case to handle rather than assume away: a camera whose session never negotiates a level-2 key still confirms its state through a cloud read, so a reading taken immediately after a write elsewhere can lag. Prefer the event over polling, and do not treat a single stale reading as authority for withdrawing a camera.

## Acceptance criteria

- [ ] A camera the SDK observes as disabled is presented to HomeKit as disabled, so Apple Home does not initiate a stream for it.
- [ ] A camera observed enabled is presented normally, and the cameras that already stream are unaffected.
- [ ] The presented state follows the SDK's enablement change event, verified by a contract test that moves the observation and asserts the characteristic follows — not by a timer.
- [ ] An absent or non-boolean enablement reading publishes nothing and does not withdraw the camera.
- [ ] A member the SDK names in `unreflectedMembers` is declined, asserted rather than assumed.
- [ ] Refusing media on a disabled camera reports a bounded reason naming that, distinguishable from a transport failure, and the allowlisted diagnostic vocabulary is extended for it.
- [ ] The existing streaming refusal and session termination behaviour still holds.

## Out of scope

- Changing the SDK's enablement wire or its readback. Done upstream; treat the value as given.
- The cost of the SDK's confirmation path (mega-yfue/eufy-sdk#84). It affects how fast a change is observed, not how it is presented.
- Any other HomeKit characteristic on the operating-mode service beyond the disabled state.

## Reproducing

Switch a camera off in the vendor app, then open Apple Home. The tile is offered and a tap fails, instead of the camera being shown as off. With mega-yfue/eufy-sdk#79 the plugin's own `enabled` reading is correct and converges, so the state to present is available.

lenoxys:
## What changed

A command resolves once the transport has carried it. Whether the device *applied* it is decided separately, by the observation a member declares — and where that observation timed out, the result was thrown away: the caller's promise had already resolved, so the failure reached the generic `error` bus as a bare `Error` with no serial, no member and no values.

A host could not tell an ignored write from a delivered one. On a camera that means telling the user the power is off while it is still filming.

The outcome now reaches the caller as **`commandUnconfirmed`**, carrying the serial, the property, the param, what was asked for and what the param actually read.

## Why not on `error`, and why not by changing `dispatch`

Both follow the precedent `commandAck` already set in this client:

- It is an **outcome, not a fault**. Nothing went wrong in this library, and on a device that was asleep the same write may still land later.
- A host wanting convergence visibility must not have to pattern-match `error`, and **the dispatch contract must not change shape** to give it. Making `dispatch` await the observation would make every observed write take up to its full deadline and would fail callers that bound their own, shorter one.

## Two exits had to be attributed, not one

The loop ending was the obvious exit. The one that **actually fires** is the last poll outliving the window, whose rejection escaped first as `"semantic event refresh timed out"` — which is exactly why every real failure showed up as an unattributable `Error`.

A private `RefreshWindowClosedError` sentinel now separates *the window closed* from *a dependency failed*: the first falls through to the attributed `StateConvergenceError`, the second still surfaces as the fault it is.

I found this only by running against real hardware — the first version passed every hermetic spec and did nothing live, because fake timers drive the loop to its natural end while reality exits through the poll. There is now a spec for the real path.

## Live qualification

Against the account with Homebridge stopped, on a `T8170` battery camera:

```
15:29:41.790  Jardin: acknowledged after 11511ms
15:30:01.790  >>> commandUnconfirmed {"sn":"T8170…37D0","property":"enabled","param":1035,
                                      "expected":1,"observed":"0","timeoutMs":20000}
15:30:01.812  Jardin: outcome=UNCONFIRMED, still reads enabled=true
error: 0
```

20.02 s after the acknowledgement, fully attributed, and nothing on the error bus. `expected: 1 / observed: "0"` is consistent with param 1035's disable-bit convention.

The same camera had **accepted** the identical write three minutes earlier (converged in 5.4 s), so the failure is intermittent rather than a property of the model — which is precisely why the caller needs to be told each time rather than the behaviour being documented as a model trait.

## Verification

- `npm run verify` green: format, typecheck, all four guards, build, seal, ESM (226 exports), examples typecheck, **2817 tests**.
- Three new specs: the attributed outcome, the real window-closed path, and a genuine post-acknowledgement fault still reaching `error`.
- One existing spec updated — it pinned `reportError` being called for a non-converging write, which is the behaviour this changes.
- `docs/events.md` documents the event alongside `sessionExpired`.

## Note for the consumer

`homebridge-eufy-security` already has the machinery to react to this (its bounded operation issuer reverts HomeKit's state and raises a diagnostic on a failed write); it was wired to a signal that never arrived. Consuming this event needs no new mechanism there.


max246 (approved):
The work is correct and well-reasoned — approving the content. **Gate first: #90 is on the #76 → #79 → #90 chain (base `fix/disabled-camera-no-video`), and #76 is CHANGES_REQUESTED with two unfixed blockers. So this cannot merge until #76 is corrected, #79 rebased, and this rebased onto `beta-0.1.0`.** The green check is "content is good," not "ready to merge now."

On #90's own diff:

**The core fix is right and the insight behind it is the valuable part.** There were two ways out of the convergence wait, and only one was attributed: the loop ending naturally (which hermetic tests reach, because fake timers drive the loop to its end), versus the last poll's own deadline firing first (which reality reaches). The second escaped as a bare `"semantic event refresh timed out"`, which is exactly why every real non-convergence surfaced as an unattributable `Error`. The private `RefreshWindowClosedError` sentinel + the `try { await beforeDeadline(...) } catch { if (!RefreshWindowClosedError) throw; break }` routes both exits to the attributed `StateConvergenceError`. Candid write-up of the hermetic-test blind spot, and there's now a spec for the real path.

**Verified the mechanics:**
- `beforeDeadline` has exactly ONE call site (the convergence loop), so changing its timeout rejection from a bare `Error` to `RefreshWindowClosedError` has no collateral — the concern a shared helper would raise doesn't apply here.
- `reportUnacknowledged` routes a `StateConvergenceError` to the new `commandUnconfirmed` event and anything else to `reportError` — so a genuine post-acknowledgement fault still surfaces as the fault it is (its own spec pins this), and the dispatch contract is untouched (an observed write doesn't start blocking to its full deadline). The `commandAck` precedent for a separate outcome channel is the right one to follow.
- The event payload in `DeviceEventMap` (`sn`, `property`, `param`, `expected?`, `observed?`, `timeoutMs`) matches `StateConvergenceError`/`StateConvergenceFailure` exactly, and `expected` (raw asked) vs `observed` (what the param actually read) is the attribution a host needs to tell an ignored write from a delivered one — which on a camera is telling the user the power is off while it is still filming.

The live qualification (T8170: acknowledged at 11.5s, `commandUnconfirmed` at 20.02s with `expected:1 / observed:"0"` matching 1035's disable-bit convention, nothing on the error bus — and the same camera having accepted the identical write 3 min earlier) is exactly why this belongs as a per-write signal rather than a documented model trait. `npm run verify` green (2817), and 15/15 + 155/155 in the client specs locally.

Merge order: fix #76 → #79 → #90. 👍

PR #94, opened 2026-08-27
feat(client)!: make the event pre-warm opt-in, and let the caller pick its power tiers

### LINKED ISSUE #93: Event pre-warm is on by default and, in the default configuration, only ever wakes standalone battery cameras — for 328 s per unwatched detection
## Observation

The `SessionManager` exists to keep battery cameras asleep — its own module doc opens with it:

> It exists to stop battery-powered cameras draining: a persistent P2P session runs a 5 s PING heartbeat forever (keeping the device awake), so instead of opening every station eagerly and holding it open, this opens a station's session **on demand** … and **auto-closes** it after an idle window whose length depends on the station's power tier.

Event pre-warm is on by default and, in the default configuration, is the **only** thing that opens a battery station without anybody asking for it — and the one it opens is a battery station by construction.

## Why battery is the only tier it can touch

| tier | state before an event | what `prewarm` does |
| --- | --- | --- |
| wired station (HomeBase / mains camera) | already open — `warmWiredP2P` warms every wired station at login (`eufy-mega.ts:1401`) and `wiredIdleMs` defaults to `Infinity`, so it never detaches | nothing: `acquire` returns the live session, `addUser`/`releaseUser` cancel and re-arm a timer that was never armed |
| battery camera attached to a wired HomeBase | the HomeBase session is open and persistent | nothing: `stationKeyOf` resolves to the base's serial, so the pre-warm targets the wired base, not the child |
| **standalone battery camera** (SoloCam and friends) | **detached, by design** | **opens a P2P session to the camera itself and heartbeats it** |

So the feature is a no-op everywhere except on the exact device class the whole on-demand design was built to protect.

## The cost is not the pre-warm window

`prewarmMs` defaults to 28 s, and the docs describe the end of it as a detach:

> If you open a live view (or, later, talkback) within the window it starts with no spin-up; if nobody watches, the session detaches. — `docs/connectivity.md:82`

The release does not detach. `prewarm` releases its user (`command-router.ts:226`), `releaseUser` sees the count hit zero and calls `armIdle` (`session-manager.ts:172`), which arms the station's tier window — `BATTERY_IDLE_MS = 300_000` (`session-manager.ts:29`). One unattended pre-warm therefore holds a standalone battery camera awake for:

```
28 s (prewarm window) + 300 s (battery idle window) = 328 s
```

at `HEARTBEAT_MS = 5_000` (`p2p-session.ts:63`) — roughly **66 PINGs for an event nobody watched**, 11× the window the option names and the docs describe. Any second qualifying detection inside that tail cancels the idle timer and starts again, so a camera that sees a person more than once every ~5.5 min never sleeps at all: the persistent behaviour the module was written to avoid, arrived at by default.

## What triggers it by default

```ts
// eufy-mega.ts:130
const DEFAULT_PREWARM_EVENTS = ["doorbellPress", "personDetected", "petDetection", "packageDelivered"];
```

Fired from the push handler whenever `autoRealtime !== false`, i.e. by default (`eufy-mega.ts:1863`):

```ts
if (this.opts.autoRealtime !== false && this.prewarmEvents.has(out.event)) {
  const dsn = (out.payload.deviceSn as string | undefined) ?? ev.deviceSn;
  if (dsn) void this.p2p.prewarm(this.p2p.stationKeyOf(dsn), this.opts.prewarmMs);
}
```

Raw `motion` is excluded, and the stated reason is frequency:

> Raw `motion` is deliberately excluded (a battery camera sees it constantly, which would defeat the idle-detach).

That reasoning does not survive contact with how these cameras are configured. A standalone battery camera is routinely set to report **human detection only** — that is the eufy app's own recommendation for battery life. In that configuration `personDetected` *is* the constant event, and the frequency argument that excluded `motion` applies to it unchanged. The SDK has no per-installation evidence to tell the two cases apart, and the default currently assumes the favourable one.

## Reproduction

On a standalone battery camera, with a `logger` that shows `debug`, walk past it once and read the session lines — no code needed, the manager narrates it:

```ts
const eufy = new EufyMega({ email, password, countryCode, logger }); // defaults only
await eufy.login();
// walk past a standalone battery camera, then watch the log:
//   [session <sn>] connecting now (battery, on demand)   <- nobody asked for this
//   [session <sn>] connected
//   ... 5 s PINGs ...
//   [session <sn>] idle (0 users) — detaching in 300000ms unless reused
//   ... 5 s PINGs for another 5 minutes ...
```

The last two lines are the report: the pre-warm window closed, and the session did not.

### Fleet the numbers above were checked against

14 devices, 8 cameras, read through the SDK with `autoRealtime: false`, resolving each camera's station key and that station's tier exactly as `stationPower` does:

| camera | topology | station tier | pre-warm effect |
| --- | --- | --- | --- |
| T8170 | standalone | **battery** | **opens + heartbeats the camera** |
| T8171, T8210, T8114 ×2 | attached to a T8010 | wired (homebase) | none |
| T8400, T8410, T8442 | standalone | wired | none (already persistent) |

One camera in eight is exposed here — and it is a SoloCam, i.e. the whole standalone-battery product line, not an exotic topology.

## Proposal

1. **Don't speculatively open a `battery` station by default.** The facade already resolves the tier (`stationPower`, `eufy-mega.ts:1437`) and already uses it to decide what to warm eagerly; the pre-warm path is the one place that ignores it. Decline there, and let a host opt in explicitly (e.g. `prewarmBatteryStations?: boolean`, default `false`) for a mains-plugged camera the SDK still classifies as battery-capable, or for someone who values instant tap-to-view over runtime. Keeping the event list intact and gating on the tier avoids guessing per-event frequency, which varies per installation and cannot be evidenced from here.
2. **An expired speculative hold should close, not inherit the idle window.** The battery idle tail is a sensible tradeoff after real use — a stream ended, a command burst ran, the user is engaged, keep it warm. After a pre-warm that nothing ever attached to, there is no user to keep it warm for, and the option that named the window already expired. `hold`/`prewarm` should be able to say "if I was the last user, close now", cutting an unused pre-warm from 328 s to 28 s. This one stands on its own even if the default in (1) stays as it is.
3. **Document the cost where the knob is.** `prewarmMs`, `prewarmEvents` and `docs/connectivity.md:82` all describe a detach at the end of the window; the 5-minute tail is not named anywhere. The "Battery cameras behind a wired HomeBase" section reassures readers about the topology where pre-warm is free and says nothing about the one where it is not. Both should name the tier, the tail, and how to turn it off (`prewarmEvents: []` disables it entirely today).

### Acceptance

- [ ] With defaults, a `personDetected` / `petDetection` / `packageDelivered` / `doorbellPress` push for a **standalone battery** camera opens no session, and a test asserts the station stays cold.
- [ ] With the opt-in enabled, the same push pre-warms exactly as today.
- [ ] A pre-warm targeting a **wired** station behaves as it does now (no regression for doorbells behind a base, and no change for the eagerly warmed stations).
- [ ] A pre-warm that expires with nothing attached closes the session at the end of `prewarmMs` rather than arming the tier's idle window; a pre-warm that a live stream attached to keeps the existing warm tail when that stream ends.
- [ ] `prewarmMs`, `prewarmEvents` and the connectivity guide state which tier is pre-warmed by default, what the session costs on a battery camera when nothing attaches, and how to change it.
- [ ] The `motion`-exclusion rationale is restated so it no longer implies `personDetected` is rare on a battery camera.

lenoxys:
## What changed

`prewarmEvents` now defaults to `[]` — an empty list is what disables the speculative pre-warm, so it
ships off and a caller opts in per event. A new `prewarmTiers?: PowerTier[]` narrows the same opt-in to
the station power tiers whose sessions may open (default: both), and `PowerTier` is exported so the
option is nameable. The decision moved out of the push handler's inline branch into one private seam,
`prewarmForEvent`.

Everything else here is documentation of behaviour that already existed: what a pre-warm actually costs,
and which station it opens.

## Why

Work out which station a pre-warm's session belongs to and the old default only ever cost, never paid:

| what fired the event | station whose session opens | effect |
| --- | --- | --- |
| camera attached to a HomeBase | the base (`stationKeyOf`) | none — already open, wired, drains nothing |
| standalone wired camera | itself | none — warmed at login, `wiredIdleMs` is `Infinity` |
| standalone battery camera | itself | opens it, and heartbeats it every 5 s |

So the only station the feature genuinely opened is a standalone battery camera — the one device class
`SessionManager` exists to let sleep, per its own module doc.

And an unwatched pre-warm does not cost `prewarmMs`. `prewarm` releases its hold, `releaseUser` sees the
count hit zero and calls `armIdle`, which arms the tier's window — so with the defaults one detection
nobody watched held the session for 28 s + 5 min (~66 heartbeats), and a second detection inside that
tail cancelled the timer and started again. A camera detecting more often than that never slept.

The in-code rationale for excluding raw `motion` ("a battery camera sees it constantly") applies
unchanged to `personDetected` on a camera configured to report human detection only — which is the
battery-saving setting a caller is most likely to meet in the field. "Rare" is a property of the
installation, not of the event name, so the SDK cannot pick that list for a caller; hence opt-in.

`prewarmTiers` defaults to both tiers rather than to `["wired"]`, because the event list is the off
switch: a caller who names an event asked for that session, and silently skipping the only tier a
pre-warm can act on would make an explicit opt-in do nothing.

## Evidence

n/a for the wire — no frame, parameter or command changes. The reasoning above is read off this
repository's own code: `DEFAULT_PREWARM_EVENTS` (now removed), `warmWiredP2P` + `wiredIdleMs: Infinity`,
`stationKeyFor`, `prewarm` → `releaseUser` → `armIdle` with `BATTERY_IDLE_MS`, and `HEARTBEAT_MS`.

Topology exposure was checked read-only against a live account (14 devices, 8 cameras, `autoRealtime:
false`, resolving each camera's station key and that station's tier exactly as `stationPower` does):
one standalone battery camera station, four cameras attached to a wired base, three standalone wired.
One camera in eight was the only one the old default could act on — and it is a standalone battery
model, i.e. a whole product line rather than an exotic topology.

## Verified how

- [x] `npm run verify` is green locally (145 files, 3065 tests), plus `npm -C docs run docs:api && npm -C docs run guard:docs`
- [ ] Ran against a real device — n/a: this removes a speculative session open. The read-only fleet
      topology check above is the only live step, and it wrote nothing.
- [x] Added or updated specs (offline, synthetic fixtures) — `src/client/__tests__/prewarm.spec.ts`, 17
      cases: every gate, the tier resolved for the station rather than the reporting device, the
      fail-closed path, and one that pins the push handler still routes each decoded event through the
      seam (a removed call site fails it). Each assertion was checked by re-introducing the defect.
- [x] Docs updated — `docs/connectivity.md` (the pre-warm section rewritten around the caller's decision:
      what naming an event buys, then the three things it costs), `docs/realtime.md`,
      `examples/08-connectivity.ts`, and the JSDoc on `prewarmEvents` / `prewarmMs` / `PREWARM_MS` /
      `P2PCommandRouter.prewarm`, which all described the window's end as a detach. Neither the guide nor
      the JSDoc enumerates the event names: the option is typed to `keyof DeviceEventMap`, so the editor
      and the generated reference already carry the list, and the guide points at `/events` — one fact,
      one place. What both DO state is the constraint you cannot read off the type: a pre-warm rides the
      **push** channel, so a poll-carried event is inert however it is listed.

## Unverified or deferred

- **The 28 s + idle-window tail is documented, not fixed.** An expiring speculative hold still inherits
  the tier's idle window. Cutting it needs `SessionManager` to tell an expired speculative hold from a
  consumer that finished, and `bumpCommand` shares the same `hold` primitive, so it belongs in its own
  change. It is proposal (2) of #93 and stays open there.
- **`prewarmTiers: ["wired"]` is close to a no-op by construction** — wired stations are warm already, so
  it only bites after a session drops. Documented as such rather than hidden; it exists so a caller can
  keep an opt-in that can never touch a battery.
- **No ADR.** The decision's rationale lives in the `prewarmEvents` JSDoc and the connectivity guide,
  which are where a caller meets it; say the word and I'll add `docs/adr/0002`.

Refs #93


max246 (approved):
Approved. This makes event pre-warm opt-in at the SDK level — exactly the right default — and the correctness that matters most here is airtight: **with no options set, nothing pre-warms.**

**Default-off, verified and pinned.** `this.prewarmEvents = new Set(opts.prewarmEvents ?? [])` defaults to empty, and `prewarmForEvent`'s gate `if (!this.prewarmEvents.has(event)) return;` short-circuits on an empty set before `p2p.prewarm` is ever reached. `DEFAULT_PREWARM_EVENTS` is gone. It's not just the gate that's tested — `it.each(events)("does not pre-warm %s with default options")` builds a client with no options and asserts each former high-intent event pre-warms nothing, so a regression that restored the old default would fail the suite.

**The seam and its four gates are right:** `autoRealtime !== false` → an opted-in event → the station is one the account reports → its tier is allowed. Two details worth calling out as correct:
- The tier is resolved for the **station whose session would open**, not the reporting device — so an attached camera is judged by its base (`stationKeyOf` → `stationPower`), pinned by "resolves the tier of the station it would open, not of the device that reported."
- Fail-closed on an unknown station (`registry.list().some(sn === station)`), tested by "declines a child whose station has no record."

**The design calls are well-reasoned.** `prewarmTiers` defaulting to both rather than `["wired"]` is correct: the event list is the off switch, so a caller who named an event asked for that session, and defaulting tiers to wired-only would silently make an explicit opt-in a no-op (the only tier a pre-warm can actually act on is a standalone battery camera). And the topology analysis justifies the breaking change — the old default only ever *cost*: the sole station it genuinely opened was a standalone battery camera, the exact class `SessionManager` exists to let sleep. Removing raw `motion` from the rationale and extending it to `personDetected` on a detection-only battery camera ("rare is a property of the installation, not the event name") is the honest reason the SDK can't pick the list for a caller.

Good candour on the deferred item too — the 28 s + idle-window tail is documented rather than hidden, correctly punted to its own change since it needs `SessionManager` to tell an expired speculative hold from a finished consumer.

17/17 in `prewarm.spec` locally; `verify` green; docs + example + JSDoc all updated to match (and correctly state the constraint you can't read off the type — a pre-warm rides the push channel, so a poll-carried event is inert). No wire change.

Side note: this supersedes the workaround we shipped in the bridge (passing `prewarmEvents: []` explicitly) — that stays correct and harmless, but once this lands the bridge inherits off-by-default for free. 👍

PR #97, opened 2026-08-28
feat!: property announcements, coded stream configuration, per-camera media isolation, and consumer-paced delivery (#95, #98, #99, #100)

### LINKED ISSUE #95: Declare poll events for the writable members nothing announces
## Observation

A host cannot learn that a writable member changed outside its own writes, for most members. Counting the declared events per capability against their member counts:

```
camera            15 members  |  cameraEnabled                     (poll)
lock              14 members  |  lockState                         (push)
battery           13 members  |  batteryLevel, batteryAlert
motion            11 members  |  motion, crying, sound, vehicle, dog
doorbell           9 members  |  doorbellPress, petDetection, package*
contact            5 members  |  contactState
arming             2 members  |  armingModeChanged, alarm

light              4 members  |  NONE
audio              6 members  |  NONE
siren              9 members  |  NONE
smart_light       10 members  |  NONE
ptz               10 members  |  NONE
keypad, storage, leak, smoke, co, suction, vacuum_*  |  NONE
```

So for a camera's `statusLed` and `nightVision`, a floodlight's `isOn` and `brightness`, and a device's microphone, speaker and volume, there is no announcement of any kind — neither push nor poll. A caller can only re-read, and re-reading cannot say *when*.

## Why this matters to a caller

`camera.enabled` already has the answer, and its own doc states the reasoning:

> *"Enablement only ever arrives as a cloud param — no id pushes it — so re-reading was the only way a caller could learn it had moved, and re-reading cannot say WHEN."*

That is why the poll-sourced `cameraEnabled` exists, mapped from every id that carries the read, with the polarity normalised. The same argument applies unchanged to `statusLed`, `nightVision`, `light.isOn`, `light.brightness`, and the audio members: they are cloud params with reads, the poll already diffs params, and `pollOnce` already turns a param diff into semantic events. What is missing is the declaration.

Concretely, for `homebridge-eufy-security`: HomeKit subscribes to notifications and otherwise reads a characteristic only while its app is open. A member whose value the host cannot be told about shows the old state in HomeKit until something happens to read it — so a user who changes the status light or the floodlight in the vendor app sees the previous value, indefinitely.

## What would help

For each writable member whose value arrives as a cloud param, declare a poll-sourced event the way `camera.ts` does:

```ts
events: enablementReads().map(({ paramType, invert }) => ({
  source: "poll" as const,
  match: paramType,
  emit: "cameraEnabled",
  derive: (s) => (s.source === "poll" && s.to !== undefined ? { enabled: asBool(s.to) !== invert } : {}),
})),
```

Mapping every id that carries the read, and normalising polarity, is what keeps the event and the getter from disagreeing — the same property that made the enablement one trustworthy.

## A related question, possibly separate

`pollOnce` emits the semantic event for a param diff, but the enablement event declares no `refresh`, and `applyParams` is not called on the poll path — it appears only in `applyRealtimeState`, in `refreshEventState`'s settled check, and in device resolution. So it is not obvious to a reader whether the live `Device` getters are fresh at the moment the poll's event fires, or only once something else applies the params.

Measured once, this worked end to end: a camera switched on in the vendor app produced `cameraEnabled` and the getter read the new value. Measured once the day before, on the same member, the getter stayed stale for 11 minutes across poll boundaries. I have not isolated the difference, so I am reporting it as a question rather than a defect — but if the poll's event can fire while the getter still reads the old value, a host told "it changed" would read the previous value and conclude nothing had.

### LINKED ISSUE #98: A wired camera's live source alternates coded geometry many times per session, reported only as changed frame dimensions
## Observation

A wired mains-powered camera (T8442, Indoor/Outdoor Cam 1080p) changes the coded geometry of its live
stream repeatedly within one session. The SDK surfaces each change only as different `width`/`height` on
subsequent `LiveVideoFrame`s, with no event, no observation, and nothing in the contract that says it can
happen.

Measured directly against the SDK at `4ec0458` — no plugin, no FFmpeg — by counting consecutive frames per
`${codec} ${width}x${height}` from the `video` event. One camera, one account, three runs:

**Run 1, `live()` alone, 30 s — 445 frames, 1 change**

```
h264 1920x1080   88 frames
h264 1280x720   355 frames
```

**Run 2, `live()` + `recordFragments()` on one shared source, 45 s — 656 frames, 6 changes**

```
h264  640x360     8 frames
h264  960x540    77 frames
h264 1920x1080   94 frames
h264  960x540    77 frames
h264 1920x1080  250 frames
h264  960x540    75 frames
h264 1920x1080   68 frames
```

**Run 3**, earlier, measured through a consumer rather than at the event: `1280x720` for 8 access units then
`1920x1080` for 31.

So the geometry is not merely negotiated once and kept. It ramps (`640x360` → `960x540` → `1920x1080`,
consistent with the camera opening at a low resolution and climbing), and then **oscillates between
`960x540` and `1920x1080` roughly every 70-250 frames**. Run 1 also shows a `1280x720` that no other run
produced, so the ladder itself is not fixed.

`live()` and `recordFragments()` each completed all three runs **without error**. Nothing here is a crash;
the stream is healthy and the frames decode. The issue is that a change of this kind is invisible in the
contract.

## Why it matters

An H.264 encoder cannot change input geometry mid-stream. Any consumer adapting this source to a fixed
output has to tear down and rebuild its encoder on every change, because the encoder it opened for
`960x540` cannot accept a `1920x1080` frame. In run 2 that is **seven encoder lifetimes in 45 seconds**.

Downstream that is not a subtle cost. A consumer feeding HomeKit re-initialises its transcode seven times,
and each re-initialisation restarts analysis and discards the frames in flight. A consumer that assumed one
geometry per session — which the contract currently permits it to assume — instead feeds mismatched frames
into an encoder that was opened for something else.

`SharedSourceHints` has no resolution hint, and `LiveVideoFrame` documents `width`/`height` without saying
they may change between frames of one session, so today a consumer either rediscovers this empirically or
gets it wrong.

## What I am asking the SDK to decide

Not a bug report with an obvious fix — a contract question, hence the detail above.

1. **Is this the camera's own adaptive ladder, or something the SDK provoked?** The ramp from `640x360`
   suggests the former, but run 2's extra oscillations appeared only once a second egress joined the shared
   pull, so the number of consumers may be influencing it. If the SDK can pin a resolution at start, that
   would settle it.
2. **Should a geometry change be an event rather than a silent field change?** A consumer that must rebuild
   an encoder needs to know a change is coming, or at minimum that one has happened, without diffing every
   frame's dimensions itself. Something like a `geometry` event, or a flag on the first frame of a new
   geometry, would let a consumer act once instead of comparing per frame.
3. **If neither, the contract should state it.** `LiveVideoFrame` saying that `width` and `height` may change
   within a session, and that a change always begins on a keyframe, would be enough for a consumer to
   handle it correctly. (In these runs each new geometry did begin on a keyframe, but that is an observation
   from three runs, not a guarantee I can rely on.)

## What this does not explain

Being explicit, because it is why I went looking. A host on this account sees live sessions and HKSV
recordings fail with its own `source-error` classification, raised when the `recordFragments` async iterator
or the live consumer's `error` event fires. **I could not reproduce that here.** Three runs, 30-45 s each,
alone and concurrent, all completed with `error=none`. So the failure that led me here is either
intermittent, or needs a condition these runs did not create — and it should not be assumed to be the same
thing as this ticket. Filed separately if and when it reproduces.

## Reproducing

Count consecutive frames per `${codec} ${width}x${height}` off the `video` event of `camera().live()` on a
wired 1080p camera for 45 s, and start a `recordFragments()` on the same device a few seconds in. No
transcoding needed; the dimensions on the event are enough.

### LINKED ISSUE #99: LiveTrace and LIVE_TRACE_MESSAGE are unreachable to consumers, so a host must hardcode the phase vocabulary and silently drops any new one
## Observation

`live-trace.ts` tells a consumer to match on this vocabulary:

> `src/transport/p2p/live-trace.ts:5-7` — "the message and its phase vocabulary have one owner here rather
> than a literal repeated at each call site. **A caller reads these to tell one startup outcome from
> another, and matches on `LIVE_TRACE_MESSAGE` plus a `LiveTrace` phase to do it.**"

Neither is reachable from outside the package.

- Exported only from `src/transport/p2p/index.ts:25`:
  `export { LIVE_TRACE_MESSAGE, type LiveTrace } from "./live-trace.js";`
- `package.json` declares a **single** export condition — `"."` → `dist/index.d.ts` / `dist/index.js`. There is
  no subpath export, so `@mega-yfue/eufy-sdk/transport/p2p` does not resolve.
- `grep -c LiveTrace dist/index.d.ts` → **0**. The package root does not re-export it.

So the only way a consumer can honour that instruction is to retype the message literal and the phase names
by hand.

## Why it matters

A host that must bound what it logs cannot pass SDK trace detail through unfiltered, because the same debug
stream carries material it may not retain. Redacting per phase is the right shape and it is what the trace
was designed for — but with the type unreachable, the host's phase allowlist is a hand-copied list with no
relationship to the one that owns it.

That has already gone wrong once, in exactly the way it must. `sequence-restart` was added in `f3929a1`
(*"resynchronize onto numbering a camera restarts"*), the eleventh phase beside ten already handled. A
consumer's allowlist kept the ten, so `sequence-restart` fell through its default branch and was reduced to
an unstructured line for as long as it existed. The information lost is not incidental: a data channel whose
numbering restarts mid-connection discards the frame being reassembled, which is one of the few pieces of
evidence available when a live session or recording fails for a source-side reason.

Nothing on either side could have caught it. The SDK widened a union it owns, which is correct and needs no
consumer coordination; the consumer's list is not typed against that union, so it compiled and passed its own
tests while dropping the new phase.

## What I am asking for

Publishing `LiveTrace` and `LIVE_TRACE_MESSAGE` from the package root, so a consumer can write

```ts
} satisfies Record<LiveTrace["phase"], true>
```

and fail to compile the next time a phase is added, instead of silently discarding it.

Both are already public in intent — the JSDoc instructs callers to use them, every field is documented as "a
fixed label, a boolean, a data-type id, or a sign code", and `LiveTrace` names no serial or address, so it
carries nothing a consumer should not see. This is a re-export, not a new surface.

If exposing the whole union is not wanted, exporting just `LiveTrace["phase"]` as a named string-literal type
would be enough for the allowlist case, which is the one that breaks.

## Not asking for

A change to the trace content or to when it fires. The phases are well chosen and the redaction discipline in
them is right; the only problem is that the vocabulary cannot be depended on from outside.

## Related

- mega-yfue/eufy-sdk#98 — the geometry alternation found in the same investigation. Independent of this.

### LINKED ISSUE #100: Two cameras on one station: a starting stream permanently adopts its sibling's video and audio
## Observation

Switching a live view between two cameras attached to the same HomeBase serves the **first camera's video and audio on the second camera's stream**, for the whole life of that stream. Reported from a real account, reproducible by hand:

- Open camera A, close it, open camera B within ~10 s → B's tile shows A's picture. Heard once: A's **audio** while B's video was on screen.
- Wait ~30 s between the two → correct.
- Intermittent in between, roughly two in three attempts landing on the wrong camera.
- The consumer's snapshot for B is then A's picture too, correcting itself only after B's next successful live refresh.

## Mechanism

Established from the code, not inferred from the symptom.

### The leak: a channel filter that switches itself off

`src/transport/p2p/live-stream.ts:265-287`. Every `LiveStream` on a station sees every camera's frames and `acceptsMedia` is the demultiplexer. It has an escape hatch:

```ts
if (this.ownFrames > 0) return false;
if (++this.foreignFrames < FOREIGN_FRAME_TOLERANCE) return false;
this.logger.warn(
  `[live] station tags media channel ${frame.channel}, not the started ${this.mediaChannel} — taking every ` +
    `frame from here (another camera streaming on this station would interleave with this one)`,
);
this.mediaChannel = undefined;
return true;
```

The hatch exists for a station that tags an attached camera's frames with something other than the channel that was started — its JSDoc says so, and says "Nothing about the fleet this was measured on needs that path".

But the condition it actually tests is **"30 media frames arrived for another channel before any of mine"**, and that is precisely what a station serving one camera at a time produces during a handover. `FOREIGN_FRAME_TOLERANCE` is 30 — two seconds at 15 fps — and video and audio share the counter, so it is reached in about a second. `ownFrames > 0` is the only guard and a starting stream has none yet.

Once it fires, `mediaChannel` is cleared permanently: that stream emits **every** frame on the station, video and audio, until it is torn down. The warning even names the consequence.

### The trigger: two channels re-issuing a start at once

`DEFAULT_KEEPALIVE_MS = 3000`, and for a HomeBase-attached camera the nudge is not a ping — `live-stream.ts:72-76`:

> "a HomeBase-attached camera has no such state and **re-sends the full media start — a genuine restart on that path**"

Leaving camera A does not stop A's stream. `SharedLiveSource.onDetach` → `arm()` only moves to `lingering` and arms an 8000 ms timer (`shared-live-source.ts:706-711`, `:391`); `stream.stop()` happens solely in `teardown`. So for 8 s after leaving A, **A's stream is still re-issuing a channel-A media start every 3 s**.

Open B inside that window and both channels are restarting on a station that serves one stream. B's first frames are A's, `ownFrames` is 0, the hatch fires, and B is bound to A's media for its whole session.

After ~8 s A has torn down, B's first frames are its own, `ownFrames > 0` holds the filter, and everything is correct — which is exactly the "~10 s wrong, ~30 s fine, intermittent" shape reported.

### Why the consumer's snapshot is wrong too

A consumer takes its live still from the same source, so the still is A's picture retained under B's identity. It persists until B's next live refresh. Nothing on the consumer side is at fault: its session, ports, processes, snapshot retention and capacity accounting are all per accessory and keyed by serial.

## No arbitration exists, and one comment says it does

`command-router.ts:658-659` states several cameras behind one station "may be warm at the same time", while `media.ts:140` describes "the one-live-channel-per-station constraint the shared path enforces". **There is no such constraint in the code.** `SharedLiveSourceOptions.concurrentCap` is documented "surfaced for observability only", is never set by any caller and is never enforced.

## What I think has to change

1. **The hatch must not fire on contention.** Frames tagged for a channel another stream on this station actually started are evidence of contention, not of mis-tagging. The hatch should be reachable only for a channel nobody asked for — or removed, since it is unneeded on the measured fleet and a stream that never delivers is a *detectable* failure while serving the wrong camera silently is the worst outcome a security camera can produce.
2. **A sibling channel starting should end the lingering pull** rather than leaving it to re-issue its start for 8 s.
3. **Decide whether a station serves one live channel at a time and enforce it in one place**, so the two comments above stop disagreeing.

## Related, separate

The vendor app exposes a station **Working Mode**: `Standard` (default) and `Performance (Beta)` — "Enhance stability and fluency when connecting multiple devices". That may be the setting that lets a station serve several streams. It is not modelled: the SDK's existing `WORKING_MODE` (1246, `battery.ts:31`) is the battery power mode, unrelated. Worth its own investigation and its own param evidence; it is an enhancement, not a substitute for the fix.

## Not established

Whether the station truly serves exactly one live stream at a time — that is the consumer's model of it and the observed behaviour is consistent with it, but no measurement here proves it.

### LINKED ISSUE #1038: Bound adaptation backpressure and keep media work off the main thread
## Parent

#981

## Observation

An adaptation process is a separate OS process, so encoding itself cannot block the event loop. Three
things still can.

- Frames are written to the adaptation process stdin with the return value discarded, and there is no
  `drain` handler anywhere in `src/`. When a child stops reading, the pipe fills and every subsequent
  frame accumulates in plugin heap for the life of the session. Measured on a host with a 2.5-core cgroup
  quota: at six concurrent sessions the peak stdin backlog reached **1.49 MB** and sustained throughput
  fell from 30 fps to 9.2 fps, while the event loop p99 rose from 10.8 ms to 85 ms.
- This defeats a policy the SDK already implements correctly. `SharedLiveSource` drops its backlog and
  resynchronises at the next IDR past its queue bound, but that path only arms when a consumer pauses,
  and the plugin never pauses. The buffering is relocated from a keyframe-aware SDK queue into a
  byte-blind pipe buffer.
- `LastSuccessfulImage` replacement performs `mkdirSync`, `chmodSync`, `writeFileSync` with `flush: true`
  and `renameSync` synchronously on the main thread for an image bounded at 10 MiB. Its JSDoc currently
  defends this as costless. A successful snapshot while several adaptations saturate the disk stalls
  everything, including motion delivery.

Motion itself holds no lock: the path from the SDK push handler to `updateCharacteristic` awaits nothing
and shares no mutex, queue or timer with a live session. The coupling is the single event loop and the
single host, which is exactly what the three items above consume.

## What to build

- Pause the SDK media consumer when a write reports backpressure and resume on `drain`, so the SDK's own
  keyframe-aware drop-and-resync policy arms instead of a second buffer policy being invented here. Keep
  one source of truth for media backpressure.
- Replace the synchronous retained-image write with an asynchronous atomic replacement, preserving the
  temporary-file-plus-rename guarantee and the directory and file modes.
- Cover both with contract specifications: a consumer that cannot keep up is paused rather than buffered,
  and no media path performs synchronous file work.

## Acceptance criteria

- [x] A stalled adaptation process does not grow plugin heap without bound.
- [x] A slow consumer resynchronises at a keyframe rather than replaying a stale backlog.
- [x] No synchronous file write remains on a media path.
- [x] Event loop delay under concurrent sessions is measured before and after.

## Blocked by

Nothing. Ready to start.

## Relationships

- Blocks nothing, and does not block #998.
- Complementary to admission control: this bounds the damage of an overloaded host, admission control
  prevents reaching that state.
- **Changes an input #1039 measures.** #1039's capacity signal is read partly from the stdin backlog that
  this ticket stops accumulating, so once this lands that signal becomes "the consumer was paused" rather
  than "the buffer grew". Land this first and #1039 is measured against the behaviour it will ship with;
  land it after and #1039's thresholds have to be re-measured.
- **Do this third of the live-adaptation set.**

### LINKED ISSUE #1054: A camera condition never says which camera, and its accessory alias cannot be resolved to one
## What happens

A camera condition names a count and never a camera. On an account with eight cameras, this is what a failed
recording looks like:

```
[camera-recording-failed] A HomeKit Secure Video recording did not complete. 1 accessory is affected.
  No action is needed; recording is attempted again on the next trigger.
[HDS …] Encountered unexpected error for recording stream 1:
  Error: recording produced no further usable output: source-error
```

Neither line says which camera, and the owner cannot act on that without guessing.

## The alias does not close the gap

The structured record does carry an identity:

```json
"affectedAccessoryCount": 1,
"accessoryAliases": ["accessory-0d44fb71-bb05-4434-9a30-08a56c2940e5"]
```

But `src/diagnostics.ts:2293-2300` mints it as `accessory-${randomUUID()}` into an in-memory `Map`. So:

- it is **regenerated every restart**, and cannot correlate failures across one;
- **nothing resolves it.** Grep for `accessoryAlias` outside `src/diagnostics.ts` returns nothing — no
  persistence, no UI surface, no support-archive manifest entry.

Its only present use is correlating conditions with each other inside one process lifetime. As device
attribution it is a dead end, and `AGENTS.md` describing diagnostics as owning "support-case accessory
aliases" reads as an intent that was never finished.

## Why this is not just "log the serial"

Two constraints make it a design question rather than an obvious change.

**The two outputs have different audiences.** `DiagnosticConditions.write` (`src/diagnostics.ts:2302-2347`)
emits the human console line and the structured JSONL record from separate statements. The console line goes
to the owner's own log, where Homebridge already prints accessory names everywhere. The JSONL record goes into
a support archive intended to be shared. `AGENTS.md` lists serials and device names as sensitive, which is
right for the archive and arguably over-strict for the console. Nothing in the code currently distinguishes
the two, so the archive's rule is applied to both.

**`diagnostics.ts` has no internal dependencies,** by the module graph in `AGENTS.md`. It cannot look up a
model or a name from a serial, so it cannot enrich a condition by itself. Any identity has to be supplied by
whoever already knows the device — the reconciler or `platform.ts` — which means a new parameter threaded
through a composition root rather than a local change.

## Options

1. **Product model on the console line only.** `scripts/README.md:62` already establishes accessory id,
   product model and power class as this repo's safe identity vocabulary — they "identif[y] the subject
   without naming rooms". `[camera-recording-failed] … for a T8442` would answer the question on a
   heterogeneous fleet with no new archive exposure. Ambiguous where two cameras share a model.
2. **Make the alias stable and resolvable.** Derive it deterministically instead of from `randomUUID()`, and
   expose the mapping where the owner can read it (custom UI, or a manifest in the support archive). Keeps
   the archive pseudonymous and survives restarts. More work, and it does not help someone reading the
   console line in the moment.
3. **Both.** Model narrows it immediately, the stable alias disambiguates duplicate models and correlates
   across restarts.
4. **Serial fragment,** as the maintainer scripts already print (`sn*0237`,
   `scripts/eufy-camera-session.mjs:24`). Unambiguous, but that convention was written for tooling output a
   maintainer reads, not for shipped logs and archives, so adopting it here is a deliberate widening of what
   the plugin retains.

Option 3 is the one worth building; option 1 alone would already remove most of the pain.

## Also worth deciding

The `Error` a HomeKit controller surfaces through HDS (`src/media/recording.ts:124-138`) has the same problem
and a tighter constraint: `media/` is handed a source, not a device, so it has no identity to name even if the
policy allowed one. Whether that error should carry a caller-supplied label, or whether the condition line is
the right place to attribute and the HDS error should stay anonymous, should be settled with the above rather
than separately.

## Acceptance criteria

- [ ] A camera condition in the console names which camera it is about, in whatever vocabulary is chosen.
- [ ] The structured record's identity survives a restart, or the record states that it does not.
- [ ] Whatever is chosen is stated as a rule in `docs/architecture.md`, because it changes what every
      installation retains about its devices.
- [ ] `test/contracts/diagnostic-conditions.test.ts` covers the console line and the structured record
      separately, since the point is that they may now differ.

lenoxys:
Closes #95, #98, #99, #100.

Four issues, one prerelease. They ship together because a host adopting any one of them lands in the others. The media fixes all sit on the same live path, and the property work changes what a host can stop polling for.

## Breaking

**Every readable property announces.** `batteryLevel` and `cameraEnabled` are retired. Both restated "a param moved" under a narrower name, and the generic announcement already carries the coerced value. `contactState` stays: it dedupes across three transports and its push carries no param at all, so retiring it would silence a door opening.

**A live stream states its coded configuration.** Consumers gain a configuration event. Any outside implementer of the consumer interface must now declare the whole event set, because a subtype cannot add one overload without restating the rest.

**A live handle is a pacing handle.** Pause, resume and keyframe-wait were always there underneath. Only the declared type withheld them, so every caller writing into its own sink had to buffer what that sink would not take.

**An attached camera's channel filter never gives up**, and media on a camera switched off is refused rather than attempted.

**A station that serves one camera at a time says so.** It used to accept a pull per camera, hand back a working handle for each, and deliver all of them in bursts.

A second camera now gets `StationBusyError` naming the channel that holds the station, and the camera being served runs at full rate. Nothing is queued and nothing is pre-empted. Work on the same camera shares one pull and is never refused, which is the shape a motion notification produces once an operator taps that camera's tile.

**Pull purposes are gone.** `PullPurpose`, `SharedLiveSource.watchedLive`, `Consumer.purpose` and the argument to `attach` are removed. See below for why.

**A media call can be abandoned.** `live`, `snapshotLive`, `openReadable` and `recordFragments` take a `signal`. Aborting settles the call with the caller's own reason and gives back what it had taken, so a pull nothing else holds is released rather than left warm for a caller that has gone. It abandons a call and never a stream: a pull another consumer holds is untouched.

**The per-member announcement opt-out is gone**, and a device publishes the station its traffic belongs to.

## Announce any property change (#95)

Thirty-four writable members across nine capabilities were readable and announced by nothing. Re-reading was the only way to learn one had moved, and re-reading cannot say when. A host that reads while its UI is open keeps showing a stale value after someone changes the setting in the vendor app.

One event now covers every readable property, derived from the same declaration the getters come from. A member is announced because it exists, not because someone remembered to wire it up. An event per member would mean a second table keyed by param id, sitting beside one that already declares that param, its type and its polarity. That is how a feature ends up disagreeing with itself.

The payload is the property name and the value, nothing else. No param id, because several ids collapse to one property and handing the id back out undoes the collapse that was the job. No previous value, because a caller that needs the delta was already told.

The value comes from live state through the same narrowing the getter uses, so an event and the getter beside it cannot disagree. Where no honest scalar exists, the name arrives alone, meaning "this moved, re-read it".

Three inbound paths reach it, each edge-triggered on the same state, so the first to see a change announces and the rest stay quiet. No dedupe table, no double delivery. Discovery is not a transition and stays silent. Echoes are announced: an inbound path cannot tell a change the SDK caused from one made in the app, and guessing loses the real external change to save one redundant re-read.

Two defects sat underneath, and both could make the SDK contradict itself. The poll never applied its diff to the live device, so a value nobody happened to read stayed stale indefinitely. And a device's own report outranked the cloud forever, with nothing ending that precedence, so a fresher cloud value could be announced and then reverted without a word. The poll now retires the report for the ids its diff moved.

The rule this leaves behind: an event earns its own name by carrying something a bare property change cannot. An inbound source the property path never reaches, a threshold crossing, or a dedupe across transports. Restating "this param moved" does not earn one.

## A live source reconfigures mid-session (#98)

The report is confirmed, and it is not one camera's quirk. Half the cameras measured change coded geometry mid-stream, oscillating rather than ramping, on both codecs, across rungs from 640x360 to 2880x1616. The ladder belongs to the camera. Nothing the SDK models caps a session.

A stream now states its configuration to each consumer, read from the parameter sets in force rather than from the frame header. Only one of those is the size a decoder produces: 1080 is not a multiple of the macroblock, so a 1080p stream codes 1088 rows and crops the difference away.

The header agreed with the sets on all but a handful of some six thousand frames, so it is not being called unreliable. It is a report rather than a definition, and it is still what gets carried before a stream's first keyframe has stated anything.

Announced per consumer, against what that consumer was given. One joining mid-session is primed with a cached keyframe it never witnessed arriving. One that overruns its queue resynchronises onto a later keyframe, having skipped the frame the change arrived on. Announcing what the source saw would leave both holding a decoder built for media they never received.

Parsing device bytes is a trust boundary, so the reader answers a geometry or nothing at all. A runaway field, an unknown chroma format, a truncated set and an out-of-range dimension all yield nothing rather than a plausible number a caller would size a decoder from.

The trace vocabulary widened while doing this: phases were added and every record now states which pull it belongs to, as an opaque per-process handle. #99 said it was not asking for a change to the trace content, so this goes beyond it. Attribution was the reason: four cameras warming off one base produced four identical records, and correlating them by time is guesswork.

## Cameras on one station showed each other's feed (#100)

The worst failure here, found while measuring #98. Switching between two cameras on one station served the first camera's video and audio on the second's stream, for that stream's whole life.

Every stream on a station sees every camera's frames, and the channel tag is the demultiplexer. It had an escape hatch: after enough frames tagged for someone else and none of its own, a stream concluded the station was not tagging per camera and took everything from then on.

That condition is what an ordinary handover produces on a station serving one camera at a time, so it cannot be told apart from the case it was written for by looking at frames. The hatch is gone. An attached camera matches its own channel unconditionally.

What the hatch protected against is still detectable. No media of its own means no keyframe, so the warm-up deadline raises a typed failure naming the stage it reached. Serving another camera's picture is silent, and on a security camera silence is the worse of the two by a wide margin.

Two things kept siblings fighting over a station that serves one at a time. A pull left lingering for a camera nobody is watching went on competing, so opening a channel now releases sibling pulls that have no viewer, while leaving watched ones untouched.

And an attached camera's keepalive was a full media start, re-asserting its channel every interval. An attached stream now stops nudging once the station delivers a frame of its own channel, and resumes only if its own media goes silent.

Two attempts are in the history rather than squashed. The first made it worse, and the second is only intelligible as its correction.

## Let a consumer pace a stream

Arming the pacing policy showed it did not hold. A sink pauses from inside its own delivery handler, and resuming drained the whole backlog regardless. That pushed a bound's worth of frames past a sink which had just said it was full, relocating the queue into whatever unbounded buffer sits behind it and defeating the drop-to-keyframe the bound exists to arm. It now stops at the re-pause and keeps the remainder queued.

Fragment recording had no bound at all. A fragment is a complete ordered unit, so its bound holds the consumer rather than discarding anything, which moves the decision down to the frame queue where dropping to a keyframe is safe.

## Why the SDK reports the constraint and does not rank its callers

An earlier round of this work had the SDK sorting pulls into viewers and thumbnails, letting a live request pre-empt a thumbnail's channel and a thumbnail decline to re-assert against a viewer. That was measured and it worked, and it was still wrong, so it was taken back out.

A priority order is a statement about what a person is looking at and about what a host can show at once. Those differ per ecosystem: one shows a still, a recording and a live view together, another is mostly tiles, a third has no continuous recording at all. An SDK serving all of them cannot carry the order of one.

The consumer already had it, which settled the argument rather than the reasoning did. A host in this account was already tracking live sessions per station and already skipping a still while a station was busy, at its own layer, keyed by station. The SDK had been reimplementing a decision its caller had made, one layer too low and shaped like one ecosystem.

What is left is the device fact, which no caller can know: a station fans several cameras over one session and serves one at a time. The refusal reports it, the `signal` is how a caller acts on it, and the ranking belongs to whoever knows what is on screen.

It is also smaller and faster. The removal took out more than it put in, and the measurement below improved.

## Also in this branch, asked for by no issue

Each of these came out of driving real hardware while fixing the four above. They are here rather than in a branch of their own because they were found on this live path and would have to be re-found on any other.

**Media on a camera switched off is refused.** Breaking. A disabled camera answers a start with audio and never a video frame, so a pull spent its whole warm-up window and then reported a timeout. It now fails immediately with a typed error naming the cause. A host that does not know that error will see a new exception type where it used to see a slow timeout.

**A still and a live view are told apart, and the live view wins.** A base cannot serve both, and the tile is off screen exactly while the live view is on it.

**A start nothing acknowledged gets a replacement session.** Roughly twenty byte-identical sends with no reply says the connection is not being heard rather than that the device is slow, so re-issuing on it spends the window to no effect. This adds a required event to the public live handle, so an outside implementer must carry it.

**A property write now waits for the level-2 negotiation to settle** rather than framing per send. Nothing bounds an unanswered gateway prompt, so the wait is bounded and charged from connect.

**A device publishes the station its traffic belongs to.** A host grouping cameras by base needs it, and it was previously derivable only from the raw record.

## Decisions worth a second opinion

**The freshness re-read announces**, against the original decision. That path fires when a caller happens to read, so its timing says nothing about the device. But it lands its fetch in the same live state the edge is computed from, and for a host that reads often it is where most fresh cloud values arrive. Silence there loses the announcement rather than deferring it.

**The per-member opt-out was built as specified, then removed.** "Nobody will care about this one" is a judgement about a host's UI, and presentation is the caller's. Its cost was fixed rather than per-use, including a field on the published schema that every consumer then has to interpret.

**Announcements need a device the caller holds.** Forced, not chosen: the value comes out of that device's live state, and one built on demand has nothing to diff against. The SDK holds them weakly and now warns once when it notices one was collected.

**The ownership guard is narrower than asked.** Grepping member identifiers out of the model floor was specified, but member keys are ordinary words like open, level, enabled and volume. A literal list would flag prose and be switched off within a week. It greps for a capability accessor being called instead, which is the reachable form of the same reach.

**A write-only member turned out to be reported**, so the flag was dropped and the read published. Its sibling stays write-only. Promoting one on the strength of its neighbour is the guess the rules forbid.

## Verified

Full gate green throughout. Exercised read-only against a fourteen-device account with nothing else driving it, and every setting touched during testing was restored and confirmed.

Property changes landed on all three inbound paths, with the getter agreeing inside the handler every time. A change made by hand in the vendor app announced once, carrying the declared payload. At rest, the fleet announced nothing across eight poll passes.

Coded configuration was measured across eight cameras and both codecs on roughly six thousand frames. A consumer joining mid-session was told the configuration in force at its join, not silence and not a replay of what it missed.

Media isolation was measured on a station carrying three attached cameras, each started in turn. Every frame arrived on the channel that started it and none on any other. A two-camera handover inside the linger window stayed on its own geometry.

Refusal on a disabled camera was measured either side of the change, on a camera switched off at the app.

Known limits, stated rather than implied:

- The leak was never reproduced on demand. It needs a race that neither an early nor a late run loses, so hermetic specs carry it, and they fail against both the original hatch and the first attempt at qualifying it.
- Consumer count is not established as a driver of reconfiguration. The comparison available was not controlled, so the guide says so rather than implying an answer.
- Not every configuration change is guaranteed to arrive on a keyframe. One measured run is unaccounted for, and a change on a delta frame would be announced when it arrived.
- The pacing paths are covered by specs rather than devices. They need a stalled encoder that only a host's own contract suite drives.
- **A caller must now choose.** A host that opened several cameras of one base and lived with the stutter has to release one before opening another. It gets the channel that holds the station and a `signal` to abandon what it no longer wants, which is enough to act, but it is work on the caller's side that did not exist before.
- **The retained still that answers a refused one is covered by specs rather than a device.** It needs a push thumbnail already in memory, which only a real motion event produces.
- The retained still that answers a yielding one is covered by specs rather than a device. It needs a push thumbnail already in memory, which only a real motion event produces.


src/transport/p2p/command-router.ts
```diff
@@ -944,15 +986,24 @@ export class P2PCommandRouter {
...
+      // that was never sent, every interval, until it timed out.
+      // Best effort means the caller can frame without the key, and `sendStartLiveOwnSession` proves it by
+      // reading `level2Key` when it sends: the framing is chosen per command, so a start issued with no key
+      // rides level 1 and the warm-up's own re-issue rides level 2 once the key has landed. Waiting first buys
+      // nothing re-issuing does not, and costs the whole grace on a camera that will never negotiate one —
+      // nothing bounds an unanswered `CMD_GATEWAYINFO`, so the caller's grace IS the bound, and it is charged
+      // from connect. Measured on three standalone cameras: 7.8 s of dead time each, then a keyframe within
+      // 400 ms of finally starting.
+      const required = opts.waitLevel2 !== "soft" || (opts.requireLevel2ForAttached === true && homeBaseAttached);
+      if (!required) return { session, parentSn, channel, accountId, homeBaseAttached };
```
martijnpoppen, line 1167: **[correctness]** This zero-wait return removed the level-2 soft grace: the command seal is now picked off `hasLevel2Key` with no wait at all, so commands issued in the first seconds after a session connects ride level-1 on stations whose key negotiation is still in flight — violating the one-seal-per-keyed-session invariant this method's JSDoc documents as the captured behavior.

**Failure scenario:** Host calls `device.setEnabled(false)` right after SDK startup; the P2P session connects and `sendBySessionLevel` resolves with `waitLevel2:'soft'` (now a zero wait, where the base waited up to 8s session-scoped via `LEVEL2_SOFT_GRACE_MS`); `hasLevel2Key` is false for the ~1–2s the negotiation takes, so the command is framed level-1 to a station that per the captured fleet only ever accepts level-2 for that family — the station ignores it and the write silently never lands.

---

src/transport/p2p/command-router.ts
```diff
@@ -655,20 +647,32 @@ export class P2PCommandRouter {
...
    */
-  async sharedLiveSourceFor(sn: string, opts: SharedLiveOpts = {}): Promise<SharedLiveSource> {
+  async sharedLiveSourceFor(
+    sn: string,
+    opts: SharedLiveOpts = {},
+    purpose: PullPurpose = "live",
+  ): Promise<SharedLiveSource> {
     const { session, parentSn, channel, accountId, homeBaseAttached } = await this.resolveSession(sn, {
       waitLevel2: "soft",
+      requireLevel2ForAttached: true,
```
martijnpoppen, line 722: **[correctness]** `MediaProvider.record` (line ~537, not in this diff) still resolves its session with a bare `waitLevel2: "soft"` (now zero wait) and without this new `requireLevel2ForAttached` flag, so on a HomeBase-attached camera `recordClip`'s first media start is silently dropped by `sendMediaPayloadLevel2` when the key hasn't landed — the exact never-sent-command failure mode the comment above says this fixed for `sharedLiveSourceFor`.

**Failure scenario:** `device.record(10)` on a HomeBase-attached camera over a fresh session: soft resolve returns before the level-2 key exists, `recordClip` opens its own `LiveStream`, the start hits `sendMediaPayloadLevel2`'s `if (!this.level2Key) return` and goes nowhere (only a debug trace); on a station whose negotiation never completes there is no reprompt and no typed key error — the clip just re-issues unsent starts until its timeout, where the shared live path now fails fast with 'level-2 key not ready'.

---

src/transport/p2p/live-stream.ts
```diff
@@ -152,6 +145,30 @@ export class LiveStream extends EventEmitter {
...
+   *
+   * A frame of this camera's own channel is the station stating it is serving THIS camera, which is the only
+   * thing the restart was trying to bring about. The SDK's own measurement agrees it is then unnecessary: both
+   * attached cameras held a 40 s stream with the nudge disabled, while the own-session camera that needs it
+   * went quiet at 13.6 s without it — so an own-session stream keeps its ping.
+   *
+   * The warm-up retry a shared source runs is untouched: it recovers a start that raced the level-2 key, and
+   * it stops at the first keyframe by its own rule.
+   */
+  private settleKeepalive(): void {
```
martijnpoppen, line 203: **[correctness]** `settleKeepalive` permanently disables the attached camera's 3s media-start re-assert after the first own-channel video unit, and nothing (no stall detector, no re-arm) recovers the stream if the station later switches to serving a sibling — consumers starve silently forever.

**Failure scenario:** Host watches attached camera A via `live()`; frames flow and `settleKeepalive` clears `kaTimer` for the stream's lifetime; a user then opens camera B in the vendor app on a one-camera-at-a-time station, so A's frames stop; the warm watch was cleared long ago, `kaTimer` is never re-armed, `acceptsMedia` now filters B's frames unconditionally, and no error or `'stop'` is ever emitted — before this PR the 3s keepalive re-asserted A's channel and resumed frames when B closed.

---

src/transport/p2p/media.ts
```diff
@@ -64,7 +64,7 @@ export async function captureSnapshotFromShared(
   const timeoutMs = opts.timeoutMs ?? 20000;
   const collectMs = opts.collectMs ?? 1500;
   const skip = opts.skipKeyframes ?? 1;
-  const consumer = source.attach();
+  const consumer = source.attach("snapshot");
```
martijnpoppen: **[correctness]** This consumer is now preemptible (a `"snapshot"`-purpose source can be dropped by `releaseLingeringSiblings` when a `live()` request lands on a sibling, and `dispose()` ends consumers with a `'stop'` emit), but the capture promise below listens only for `'video'` and `'error'`, never `'stop'` — an in-flight cold snapshot on a preempted sibling hangs its full 20s timeout and then reports the misleading no-keyframe error.

**Failure scenario:** Four snapshot tiles refresh off one HomeBase; a `live()` request for camera 2 lands while camera 3's cold snapshot is mid-warm-up; `releaseLingeringSiblings` drops camera 3's source (snapshot-only, purpose 'live' preempts), `dispose()` calls `consumer.end()` which emits `'stop'`; the capture's promise has no `'stop'` handler, so the caller waits 20s and gets `LiveSnapshotUnavailableError('no clean keyframe within 20000ms')` instead of failing fast with the real cause.

---

src/model/capabilities/camera.ts
```diff
@@ -547,16 +538,20 @@ export const CAMERA_MEMBERS = {
...
-      (opts?: Parameters<MediaProvider["live"]>[0]) =>
-        m.live({ powered: poweredOf(ctx), ...opts }),
+    (m, { ctx, read }) =>
+      async (opts?: Parameters<MediaProvider["live"]>[0]) => {
+        refuseWhenDisabled(ctx, read);
+        return m.live({ powered: poweredOf(ctx), ...opts });
+      },
     "Open a managed live stream.",
   ),
   record: provided("media", (m) => m.record, "Record N seconds → an mp4/h264 buffer."),
```
martijnpoppen: **[correctness]** The new `refuseWhenDisabled` gate covers only `live` and `snapshotLive`; `record`, `openReadable`, and `recordFragments` still start a pull against a camera whose `enabled` reading is false, hitting the 20s warm-timeout the `CameraDisabledError` was introduced to replace — the gate sits on two members instead of the shared media-pull layer.

**Failure scenario:** Camera reports `enabled=false`; `device.live()` rejects immediately with `CameraDisabledError`, but `device.record(10)` on the same device warms a stream that delivers audio and never a video frame, ties up the station channel, and fails 20s later with `LiveStreamStartError` stage 'audio-only' — two different failures for one known-off camera, and `record`/`openReadable` pulls from a disabled camera can now also preempt a sibling's lingering pull for nothing.

---

src/transport/p2p/shared-live-source.ts
```diff
@@ -417,6 +533,38 @@ export class SharedLiveSource {
...
+   * never delivered would.
+   *
+   * The start is re-issued at once and then on the retry's cadence, because a station that stopped serving a
+   * channel when its last consumer left is the very case the retry recovers: waiting the whole window to
+   * report what one re-issued start can fix is a timeout where a stream was available.
+   *
+   * Nothing is armed while a watch is already pending, so several consumers joining one reused stream share the
+   * watch the first of them started.
+   */
+  private watchReusedStream(): void {
```
martijnpoppen, line 653: **[correctness]** `watchReusedStream`'s watch is cleared only by a keyframe (`onVideo` clears the warm watch inside `if (frame.keyframe)`), not by "the first frame to arrive after the join" as its doc claims, so a reused stream serving delta frames for `warmTimeoutMs` has `onWarmTimeout` fail and tear down every consumer — including healthy ones that were decoding fine.

**Failure scenario:** A snapshot consumer attaches to a live, actively-delivering source on a camera configured with a long GOP (keyframe interval > 20s); delta frames keep flowing but the re-armed `warmDeadlineTimer` never clears, `onWarmTimeout` fires, fails ALL consumers (the healthy live viewer included) with `LiveStreamStartError`, and tears the shared source down; meanwhile the 2s retry ticker re-sent a full attached media start every tick against a stream that was never broken.

---

src/transport/p2p/shared-live-source.ts
```diff
@@ -196,22 +233,44 @@ class ConsumerImpl extends EventEmitter implements Consumer {
...
+  /**
+   * Release delivery and hand over the queued backlog, stopping the moment the sink re-pauses.
+   *
+   * A sink applies backpressure by pausing from inside its own delivery handler, so draining the whole
+   * backlog regardless would push a bound's worth of frames past a sink that already said it was full —
+   * relocating this queue into whatever unbounded buffer sits behind it and defeating the drop-to-keyframe
+   * policy the bound exists to arm. Whatever is left instead stays queued and keeps counting against the
+   * bound, so a sink that never keeps up resynchronises at an IDR rather than replaying stale media.
+   */
   resume(): void {
```
martijnpoppen, line 246: **[correctness]** `ConsumerImpl.resume()`'s drain loop checks `this.paused` but not `this.detached`, so a sink that calls `stop()`/`detach()` from inside a delivery handler mid-drain keeps receiving the rest of the drained backlog after it detached.

**Failure scenario:** A readable egress paused with 10 queued frames resumes; its `'video'` handler hits a downstream destroy and calls `consumer.stop()` while frame 3 is being flushed; `detach()` sets `detached=true` and empties `this.queue`, but `resume()` keeps iterating its local `q` and emits frames 4–10 (`'video'`/`'media'`) into handlers wired to a destroyed stream — `deliverVideo` guards `detached`, the flush does not.

---

src/transport/p2p/command-router.ts
```diff
@@ -924,7 +963,10 @@ export class P2PCommandRouter {
    * {@link P2PSession.repromptLevel2Key}, which explains why one settled negotiation is not the last word. A
    * soft caller has a level-1 path and never re-prompts.
```
martijnpoppen: **[documentation]** `resolveSession`'s JSDoc still describes `'soft'` as "best-effort short wait, don't throw", but soft now waits zero milliseconds, and with `requireLevel2ForAttached` on an attached device it takes the full hard grace and DOES throw (`'level-2 key not ready'`).

**Failure scenario:** The next contributor adds a caller relying on the documented contract that soft briefly waits and never throws; on an attached camera their call now either proceeds with no key (command silently unsent) or throws from a path the doc promises cannot throw — the doc describes the code this PR deleted.

---

src/transport/p2p/p2p-session.ts
```diff
@@ -420,9 +420,11 @@ export class P2PSession extends EventEmitter {
     const remaining = graceMs - (Date.now() - since);
```
martijnpoppen, line 480: **[dead-code]** `awaitLevel2Key`'s `graceFrom:'session'` mode is now dead code — every remaining caller passes `'call'` after this PR removed the soft-grace path — and its doc ("Best-effort media uses it") describes a caller that no longer exists.

**Failure scenario:** Maintenance cost: the `connectedAtMs`-anchored branch and its doc paragraph survive with zero production callers, inviting a future caller to use a session-scoped grace whose only rationale (`LEVEL2_SOFT_GRACE_MS` best-effort media) this PR deliberately removed; drop the parameter and the branch.

---

src/transport/p2p/command-router.ts
```diff
@@ -655,20 +647,32 @@ export class P2PCommandRouter {
...
       waitLevel2: "soft",
+      requireLevel2ForAttached: true,
     });
     const key = `${parentSn}:${channel}`;
     let source = this.liveSources.get(key);
     if (source && source.state === "stopped") {
       this.dropLiveSource(key);
       source = undefined;
     }
+    if (source) this.releaseLingeringSiblings(parentSn, channel, purpose);
```
martijnpoppen: **[simplification]** `sharedLiveSourceFor` calls `releaseLingeringSiblings` identically in both the `if (source)` and `if (!source)` branches — one unconditional call before the branch does the same job.

**Failure scenario:** Pure duplication: the two adjacent guarded calls with the same arguments read as if the branches differ when they do not, and a future edit to one site (e.g. changing the purpose argument) silently desynchronizes the reuse path from the cold-start path.

---

martijnpoppen (commented):
Code review of head 63fca5f (66 files, +3889/−348 vs `beta-0.1.0`): **10 findings — 7 correctness, 1 stale doc, 1 dead code, 1 duplication** — posted as inline comments.

The dominant theme is that the rework of the level-2 key grace and the shared-live-source lifecycle left several paths behind:

- **Level-2 key handling:** the removed soft grace means early commands can silently go out level-1 (`sendBySessionLevel`), and `record()` on attached cameras lacks `requireLevel2ForAttached`, so its media start can be silently dropped — the exact failure this PR fixed for the shared live path.
- **Stream lifecycle:** the keepalive is permanently disarmed after the first frame with no recovery if the station switches cameras; the warm watch on reused streams only clears on a keyframe (long-GOP cameras tear every consumer down); snapshot captures don't listen for the new preemption `stop` and hang 20s with a misleading error.
- **Smaller gaps:** `refuseWhenDisabled` misses `record`/`openReadable`/`recordFragments`; `resume()`'s drain loop keeps emitting to a detached consumer; a stale `resolveSession` JSDoc; a dead `graceFrom:'session'` branch; a duplicated `releaseLingeringSiblings` call.

Verified against the head: `tsc --noEmit` clean, all 3348 tests pass. H.264/H.265 SPS parsing, the crop/chroma math, poll/realtime precedence, the per-consumer `video-config` fold, and the fragment backpressure loop were checked and came up clean. Three findings (record's soft resolve, the snapshot `stop` handler, the dead `graceFrom` branch) reference code just outside the diff, so their comments are anchored to the nearest related changed line.

---

src/transport/p2p/command-router.ts
lenoxys: Right, and this was the worst of the ten. I reasoned about `sendStartLiveOwnSession`, which reads `level2Key` on every send and is re-issued by the warm-up, and never looked at a caller that picks its seal ONCE. A property write framed level-1 to a family that only accepts level-2 is ignored and nothing re-frames it.

Fixed in 9e1a197: `sendBySessionLevel` asks for a new `"settle"` mode — `awaitLevel2Key(LEVEL2_SETTLE_MS, "session")`, session-scoped so a station offering no key does not charge it to every later command, and never throwing. Media keeps its zero wait, which is where the 7.8 s of dead time measured on three standalone cameras came from.

src/transport/p2p/command-router.ts
lenoxys: Correct — the fix was incomplete. `record` opens its own pull and reaches `sendMediaPayloadLevel2` on exactly the same terms as a shared source, so the never-sent start was still reachable there.

Fixed in 9e1a197: it carries `requireLevel2ForAttached: true`. `openReadable` and `recordFragments` already resolve through `sharedLiveSourceFor`, so they were covered.

src/transport/p2p/media.ts
lenoxys: Agreed, and this one is downstream of the preemption in this same PR — the pull under a cold snapshot can now be dropped, and `dispose()` ends its consumers with `stop`, which the capture never listened for.

Fixed in 9e1a197: it listens, and rejects with `source-failed` naming the source ending before a keyframe, rather than waiting out 20 s to report a missing one.

src/model/capabilities/camera.ts
lenoxys: Agreed. The gate went on the two members I had in hand rather than on what the rule says, and `record` on a known-off camera would still warm to the `audio-only` timeout the error was introduced to replace.

Fixed in 9e1a197: `record`, `openReadable` and `recordFragments` refuse too. `snapshotStored` stays exempt — a retained push thumbnail is not a pull — and an `undefined` reading still refuses nothing.

src/transport/p2p/shared-live-source.ts
lenoxys: Verified and you are right: `clearWarmWatch()` sits inside `if (frame.keyframe)`, so my doc claim — the first frame after the join — described behaviour the code did not have. On a group of pictures longer than the window the deadline outlives an actively delivering stream, and `onWarmTimeout` fails EVERY consumer.

Fixed in 9e1a197: a reuse watch is tracked apart from a cold warm-up and any frame settles it, video or audio. A cold warm-up still requires a keyframe, because nothing can be decoded without one; a join already holds the retained keyframe, so what it is missing is only evidence the stream is still served.

src/transport/p2p/shared-live-source.ts
lenoxys: Agreed. The base did not check `paused` either, so this PR added the mid-drain guard and stopped one item short of the one you name — `flush` walks a local copy that `detach()` cannot empty.

Fixed in 9e1a197: the loop checks `detached` per item, as `deliverVideo` does.

src/transport/p2p/command-router.ts
lenoxys: Correct, and it is this repository's own rule — the doc described the code the PR deleted.

Fixed in 9e1a197: the JSDoc states the four modes as they are, including that `requireLevel2ForAttached` promotes a soft caller to a throwing one on an attached camera, and that only a caller which requires the key re-prompts.

src/transport/p2p/p2p-session.ts
lenoxys: The observation was accurate and the conclusion has been overtaken: `"session"` was dead only because I had broken its one legitimate caller.

`sendBySessionLevel` picks its seal once and needs the negotiation to have settled, which is precisely a session-scoped grace measured from connect — see 3890376203. In 9e1a197 it uses `"session"` again, under `LEVEL2_SETTLE_MS`, so the branch has a production caller and its doc names it. Not removed.

src/transport/p2p/command-router.ts
lenoxys: Agreed, pure duplication — and the two sites were identical, so nothing behavioural was hiding in it.

Fixed in 9e1a197: one call before the branch, with a comment stating why a reuse frees the station exactly as a cold start does — the measured failure was the other channels re-tasking the station, not the pull already open on this one.

src/transport/p2p/live-stream.ts
lenoxys: Verified: there is no stall detector anywhere, so once the re-assert is settled nothing recovers a stream the station gives to a sibling — no frames, no error, no `stop`. Your scenario holds.

I did not restore the 3 s re-assert, because it was measured harmful in the other direction: two attached streams each re-asserting every 3 s contend continuously, and settling it is what let both hold a 40 s stream. Silence is the condition the re-assert is actually for.

Fixed in 9e1a197: the settle holds only while own-channel media keeps arriving. Silence for `stallMs` (twice the keepalive by default, so a flowing stream never reaches it) re-arms the re-assert, and the next own frame settles it again. A spec asserting the old behaviour after ONE frame encoded the starvation and now feeds media across five intervals instead.

src/transport/p2p/media.ts
```diff
@@ -137,8 +151,8 @@ export async function captureSnapshotFromShared(
  * Recording starts at the first complete keyframe so the clip is seekable. Requires `ffmpeg`.
  *
  * Opens its OWN {@link LiveStream} over the session rather than joining the device's shared source, so it
- * costs a second pull on a camera already streaming and does not take part in the one-live-channel-per-station
- * constraint the shared path enforces. `recordFragments` is the shared-consumer path.
+ * costs a second pull on a camera already streaming, and the shared path's release of a sibling's lingering
+ * pull does not reach it. `recordFragments` is the shared-consumer path.
```
martijnpoppen, line 155: **[correctness]** `recordClip()` (body below, ~line 177 onward) can hang forever: once `capturing` flips true the timeout is cleared, resolution only happens inside the next `'video'` event (`Date.now() >= stopAt`), the `'error'` handler is an empty swallow (`stream.on("error", () => {})`), and there is no `'stop'` listener or post-capture deadline.

**Failure scenario:** `await device.camera().record(30)` on a camera that goes quiet mid-clip (this module's own history records a T8170 going silent 13.6s into a live stream with no stop/error) or whose P2P session closes: no further `'video'` events fire, so the resolve is never reached, any upstream failure is discarded, and the returned promise stays pending for the life of the process.

---

src/transport/p2p/command-router.ts
martijnpoppen, line 1: **[correctness]** `openTalkback` (line ~628, outside this diff so commented at file level): the one-talkback-per-camera exclusivity check is a TOCTOU — `talkbacks.has(key)` is checked at ~635, then `sharedLiveSourceFor()` is awaited at ~641, and `talkbacks.set(key, talk)` only happens at ~661, so two concurrent calls both pass the check and both start.

**Failure scenario:** Two callers invoke `device.camera().talkback()` concurrently for the same camera: both see `talkbacks.has(key) === false`, both await the shared source, both construct and start a `Talkback`; the second `set()` overwrites the first in the map, both pace AAC onto the session's single ordered audio sequence (interleaved, unplayable — exactly what the guard's own error message says must be refused), and the orphaned first handle is no longer reachable by `closeAll()`.

---

src/client/eufy-mega.ts
martijnpoppen, line 1: **[efficiency]** `refreshEventState()` (line ~710, outside this diff so commented at file level) polls `this.registry.getDevices()` — the unthrottled, account-wide house-scoped fetch — in a ~500ms loop for the whole convergence window, bypassing the registry's own `LIST_REUSE_MS`/`refreshedList()` coalescing (device-registry.ts:187/507) built to prevent exactly this burst.

**Failure scenario:** One observed command (or push/notify with a refresh) whose param never converges spins `beforeDeadline(this.registry.getDevices(), remaining)` roughly every 500ms until `refresh.timeoutMs`; on a multi-house account each `getDevices()` is `get_house_list` + one `get_devs_list` per house, so a single stuck write issues dozens of cloud requests, and several concurrent transitions multiply it — `refreshedList()` exists to collapse these but is not on this path.

---

martijnpoppen (commented):
Second review pass, this time reading deep into the transport layer at the PR head (now 12ad436): **6 findings, none overlapping the 10 from the first review.** Three are inline (`recordClip` hang; `openTalkback` TOCTOU and `refreshEventState` polling as file-level comments, since those bodies sit outside the diff). The remaining three live in files this PR doesn't touch, so GitHub won't take inline comments there — they're listed below because they surfaced while tracing this PR's paths and two are security-relevant:

**`src/transport/mqtt/secure-mqtt.ts:121` + `broker-discovery.ts:100` — TLS verification is fully disabled on the pinned-broker path.** `rejectUnauthorized: false` disables ALL server-certificate verification; the adjacent comment's claim that "the chain itself is still verified" because `ca` is pinned is wrong — with `rejectUnauthorized: false`, verification errors are ignored, `ca` or not. A network-position attacker answering the pinned broker IP can present any self-signed cert, complete the handshake, observe/inject the account's device command and report traffic, and exercise the per-user client cert against a hostile endpoint. Correct fix: `rejectUnauthorized: true` plus a `checkServerIdentity` override that validates against `credentials.endpoint_addr` (Node fails hostname checks for bare IPs unless overridden — that's the real reason the "proven-working recipe" needed verification off).

**`src/transport/push/push-client.ts:121` — the FCM/MCS connection to mtalk.google.com is opened with `rejectUnauthorized: false`**, sending the account's `androidId`/`securityToken` over an unverified channel. An on-path attacker can capture the LoginRequest credentials (enough to impersonate the push registration) and inject forged `DataMessageStanza` pushes — which flow into `normalizePushEvent`, semantic events, prewarm decisions, and thumbnail-candidate URLs the `StoredImageCache` then downloads. Google's endpoint serves a normal, verifiable certificate; there's no reason to disable verification here.

**`src/transport/p2p/session-manager.ts:198` — a `hold()`'s expiry timer outlives its entry.** After `close()`/`resetWhenUnused` discards the entry and a new one is created, the stale `setTimeout`'s `releaseHold` acts on the successor entry: `holds` floors at 0 but `releaseUser` still decrements `users`, stealing a live consumer's refcount. Scenario: `bumpCommand(sn)` arms a 15s hold → `resetWhenUnused` closes and discards the entry → a live stream re-acquires the station (`users=1` on the fresh entry) → the stale timer fires and drops `users` to 0, arming the idle timer; the session is later closed while the stream's consumer is still attached, and its eventual real `releaseUser` underflows to a no-op. Fix: capture the entry (or a generation token) in the timer closure and no-op when it no longer matches.

Things checked and cleared on this pass: the LZ4 block decoder, the Annex-B SPS/HEVC bit-reader, the fMP4 muxer, the ECIES/GCM crypto in `codec.ts`/`video.ts`, the `onData` sequence-wrap reassembly, and `structuralEqual`/`clamp`/`asBool` in `core/util.ts`.

---

lenoxys:
@martijnpoppen hey I'm on my phone I don't think I've touched the 3 files you mention in this review. Can you confirmed it ?

But worth to raise an issue about it. This PR is huge to add more thing into it.

martijnpoppen:
Confirmed — none of the three files is touched by this PR. The diff `beta-0.1.0...12ad436` has no hunks in `src/transport/p2p/session-manager.ts`, `src/transport/mqtt/secure-mqtt.ts` (nor `broker-discovery.ts`), or `src/transport/push/push-client.ts`; all three findings are pre-existing on the base branch. They only surfaced because the second review pass traced this PR's live-media and event paths through them.

Raised separately as requested: #101 (TLS verification disabled on the MQTT/push transports — the `rejectUnauthorized: false` sites, with the `checkServerIdentity` fix sketched) and #102 (the stale `hold()` timer stealing the successor entry's user refcount in `SessionManager`). Nothing about them blocks this PR.

The three inline comments from that pass (`recordClip` hang, `openTalkback` TOCTOU, `refreshEventState` polling) do concern code on this PR's paths, though `openTalkback` and `refreshEventState` are also pre-existing lines — happy to split those out too if you'd rather keep this PR frozen.

---

martijnpoppen (approved):

src/transport/p2p/media.ts
lenoxys: Confirmed, and worse than a hang risk: the resolve lived inside the `video` handler, so the clip's end depended on a frame arriving after it — and the empty `error` handler threw away the one signal that said none would.

Fixed in 25db428: the window is armed as a deadline when collection starts, so it closes on the clock and answers with what the camera actually delivered; the session's `close` and the stream's `error` fail the clip, since no further frame can arrive on either. The `error` listener stays attached past the settle on purpose — an unhandled one takes the host process down, and the stream is stopped only afterwards, so a late failure lands on a settled promise as a no-op.

One correction to the scenario: a `stop` listener would be dead code here. `recordClip` owns the only reference to its `LiveStream`, so nothing else can stop it — the session closing is the reachable signal, and that is what it listens for. `record-clip-bounds.spec.ts` pins all three endings.

src/transport/p2p/command-router.ts
lenoxys: Agreed — the guard read the map before a round-trip and wrote it after, so it refused nothing that arrived inside the window. Two paced streams on one ordered audio sequence, and the first handle unreachable by `closeAll`, exactly as described.

Fixed in 25db428: the talkback is constructed and recorded in the SAME synchronous step as the check, before the shared source is awaited; the consumer is wired into it once the source is warm, and a failure to warm or to open the audio path releases the claim.

The mirror case your comment implies is covered too: a `closeAll` inside that window stops the talkback holding the claim, so the claim is re-checked after the wait rather than starting the pacing tick on a stopped talkback. Both are pinned in `talkback-session.spec.ts` — one start frame on the sink for two concurrent callers.

src/client/eufy-mega.ts
lenoxys: Correct, and it is this repository's own coalescing that was bypassed — `refreshedList` exists for precisely this shape of caller and the convergence loop was not using it.

Fixed in 25db428: the loop asks through `DeviceRegistry.refreshedList`, so an unconverged param costs one account-wide list per `LIST_REUSE_MS` window instead of one per ~500 ms pass, and concurrent transitions coalesce onto the single in-flight fetch rather than multiplying. `refreshedList` is public now with its contract stating that it is the only way a caller in a loop should ask for a fresher list.

The loop keeps its own cadence deliberately: each pass re-reads what is known, so a value the device volunteers over its own session still settles the wait between two cloud reads rather than after them. `catch-all-event.spec.ts` now pins the fetch count as well as the convergence.

PR #109, opened 2026-09-04
fix(mqtt,push)!: verify the server on a pinned broker dial, and on the push channel at all

### LINKED ISSUE #101: TLS server-certificate verification is disabled on the MQTT and push transports (rejectUnauthorized: false)
Found while reviewing #97; the code is pre-existing on `beta-0.1.0` and untouched by that PR, so it's filed separately as requested there.

Three TLS connections disable server-certificate verification entirely:

- `src/transport/mqtt/secure-mqtt.ts:121` — the pinned-instance mTLS path sets `rejectUnauthorized: false`. The adjacent comment claims "The CA is still pinned above, so the chain itself is still verified" — that is not how Node's TLS works: with `rejectUnauthorized: false`, verification errors are **ignored**, `ca` option or not. Nothing about the presented certificate is checked.
- `src/transport/mqtt/broker-discovery.ts:100` — same pattern on the discovery connect.
- `src/transport/push/push-client.ts:121` — the FCM/MCS connection to `mtalk.google.com:5228` is opened with `tls.connect(PORT, HOST, { rejectUnauthorized: false })`.

**Impact**

- MQTT: a network-position attacker answering the pinned broker IP can present any self-signed certificate, complete the handshake, observe and inject the account's device command/report traffic, and exercise the per-user client certificate against a hostile endpoint.
- Push: the account's `androidId`/`securityToken` are sent in the LoginRequest over an unverified channel — enough to impersonate the push registration — and an on-path attacker can inject forged `DataMessageStanza` pushes, which flow into `normalizePushEvent`, semantic events, prewarm decisions, and thumbnail-candidate URLs that `StoredImageCache` then downloads.

**Why it was probably turned off, and the correct fix**

The brokers are addressed by bare IP (`endpoint_addr`), and Node's default `checkServerIdentity` fails hostname verification for an IP not present in the certificate's SANs — which is likely why the "proven-working recipe" needed verification off. The fix that keeps the chain check is:

```ts
{
  ca: pinnedCa,
  rejectUnauthorized: true,
  checkServerIdentity: (host, cert) => {
    // validate against credentials.endpoint_addr (and/or the pinned cert's identity)
    // return undefined on match, an Error otherwise
  },
}
```

For `mtalk.google.com` there is no such excuse — Google serves a normal, publicly verifiable certificate, so plain `rejectUnauthorized: true` (the default) should just work there.

lenoxys:
Closes #101. Design decisions recorded in https://github.com/mega-yfue/eufy-sdk/issues/101#issuecomment-5537683368.

## What changed

Three connections disabled server-certificate verification. All three now verify.

- **`transport/mqtt/bare-ip-tls.ts`** (new, not barrel-exported) — the TLS option fragment for dialling a broker instance by IP: client certificate, pinned CA, SNI on the real hostname, `rejectUnauthorized: true`, and a `checkServerIdentity` that delegates to Node's own matcher **against the hostname** instead of the dialled IP.
- **`secure-mqtt.ts`** — the pinned-instance branch uses it. The hostname branch was already `rejectUnauthorized: true` and is untouched.
- **`broker-discovery.ts`** — the probe uses it.
- **`push-client.ts`** — the options object is deleted, so `tls.connect(PORT, HOST)` gets Node's defaults.

## Why the old comment was wrong

> The CA is still pinned above, so the chain itself is still verified.

It was not. With `rejectUnauthorized: false`, verification errors are ignored whether a `ca` is supplied or not — nothing about the presented certificate was checked, so any self-signed certificate completed the handshake.

## Why the fix is this small

Both bare-IP sites already set `servername` to the real hostname, so the instance answers with the certificate it would have served a DNS-resolved connect. The only thing failing was Node matching that certificate against the *IP* it dialled. Nothing is relaxed or reimplemented — Node's `checkServerIdentity` is given the name the server actually answered for.

One place, not two: the recipe was copy-pasted, which is how the same mistake reached both call sites.

## Verified against real brokers

Confirmed after opening this PR — no credentials needed, since the server certificate arrives before any client authentication. Dialling each instance IP with SNI on the hostname and running the fix's own identity check:

- `CN=*.anker.com`, SANs `DNS:*.anker.com, DNS:anker.com`, issuer `Go Daddy Secure Certificate Authority - G2`
- `authorized: true` on three instance IPs across two regions
- verification against the hostname passes; against the dialled IP it is rejected

So the certificate does name `endpoint_addr`, and the IP mismatch really was the only thing `rejectUnauthorized: false` was hiding. Details in a comment below.

## Trust anchor, and the absent knob

The cloud-supplied CA continues to **replace** Node's store rather than extend it. It arrives per-user from `get_user_mqtt_info` over an already-verified HTTPS channel, alongside the client certificate and key it is used with, so trusting exactly that root is sound and strictly narrower than trusting every public CA. The rationale is in JSDoc above the helper, since it is the decision most likely to be "fixed" back by a future reader.

There is **no** option to disable verification, deliberately. Such a flag gets set once by whoever hits a transient error and never unset, and any name in the `allowUnverified` family would collide with `EufyMegaOptions.tuyaAllowUnverified`, which means something entirely different.

`mtalk` keeps Node's trust store rather than a pin — pinning a root we do not control is a self-inflicted outage on their next rotation.

## Tests

`bare-ip-tls.spec.ts` drives Node's real `checkServerIdentity` through the helper with hand-built certificate shapes — no socket, no fixtures, no `openssl`:

- a certificate for the hostname is accepted even though the socket dialled an IP;
- a certificate for any other name is rejected;
- a certificate naming only the instance IP is rejected, so holding one for the address is not enough;
- the hostname the TLS stack passes in is ignored — only the identity the helper was built with matters;
- and one case asserts the **default** matcher failing the legitimate certificate, which is precisely the behaviour the disabled flag was hiding.

Asserting the captured options alone would not distinguish a correct matcher from one returning `undefined` unconditionally, which is the failure mode being fixed — so the transports additionally get options-shape assertions proving they use the helper.

`broker-instance-probe.spec.ts` gives `probeBrokerInstance` its first coverage (only its pure helpers were tested before): the dial shape, a granted subscribe, a denied `0x80` grant, a TLS failure reported as unreachable rather than thrown, and that it never publishes.

## Verification

`npm run verify` green (3414 tests). `npm -C docs run build` green, including `guard:docs`.

## No disclosure process

The only published version (`0.0.5`) was an empty placeholder, so nothing carrying this code has ever shipped and real-world exposure is nil. That is also why #101 stayed a public issue.


martijnpoppen (approved):
Good fix for a real vulnerability (three sites accepting any certificate). Checked the actual mechanism: `bareIpTlsOptions` sets `servername` to the real hostname (so the instance answers with its normal certificate) and verifies against that same hostname via `tls.checkServerIdentity`, ignoring whatever hostname argument Node's TLS stack happens to pass in — which the tests confirm is exactly the behavior. `rejectUnauthorized: true` with the pinned per-user CA replacing Node's trust store (not extending it) is the right call given the CA arrives over an already-verified HTTPS channel. The `push-client.ts` fix is the simplest of the three and clearly correct — dropping `rejectUnauthorized: false` and adding explicit SNI restores Node's default verification.

One thing worth confirming before merge, which the PR description itself flags as unverified: that the broker's certificate actually names `endpoint_addr` (not just the instance IP or something else) in its SANs. If it doesn't, this breaks the pinned dial outright per the author's own note. Since there's no live broker to test against here, I can't verify that independently either — but given the author's already documented a fallback plan (verify against whatever name the cert actually carries) if that assumption turns out wrong, this seems reasonable to merge and confirm against a real broker afterward rather than block on it, since the current behavior (no verification at all) is strictly worse either way.

---

martijnpoppen (approved):
Re-reviewed after the new push — this one only trims/simplifies the test fixtures (the multi-name `certFor` helper collapsed to single-name, a redundant `secureMqtt` publish assertion dropped, etc.); the actual fix in `bare-ip-tls.ts`, `secure-mqtt.ts`, `broker-discovery.ts`, and `push-client.ts` is unchanged from what I already verified. Still good — approving again.

---

lenoxys:
**The unverified item is now verified — no account or credentials needed.** The server certificate is presented before any client authentication, so the check only needs a socket.

Dialling each instance IP directly with SNI on the hostname, and running the fix's exact identity check:

| instance | authorized | vs hostname | vs dialled IP |
|---|---|---|---|
| `35.156.167.94` (eu) | `true` | OK | **reject** |
| `3.69.251.162` (eu) | `true` | OK | **reject** |
| `3.147.137.189` (us) | `true` | OK | **reject** |

Certificate on all three, both regions:

- subject `CN=*.anker.com`, SANs `DNS:*.anker.com, DNS:anker.com`
- issuer `Go Daddy Secure Certificate Authority - G2`, valid to 2027-01-20

Three things this settles:

1. **The certificate does name `endpoint_addr`.** `aiot-mqtt-{region}.anker.com` is one label under `*.anker.com`, so the wildcard covers it and `checkServerIdentity(endpoint_addr, cert)` passes. That was the one thing I could not confirm from the code.
2. **The dialled IP is genuinely rejected** by Node's default matcher — the IP appears in no SAN. So `rejectUnauthorized: false` really was papering over a hostname mismatch and nothing else, exactly as the comment claimed, and the fix removes the cause rather than the symptom.
3. **The GoDaddy note in the code is right.** The chain really is GoDaddy, which is why `aws_root_ca1_pem` is misnamed — and it means the pinned root is a real public CA, so pinning it is narrowing the trust store rather than replacing it with something unverifiable.

Caveat worth stating: `authorized: true` above came from the system trust store, since I have no `get_user_mqtt_info` response to pin. The chain is valid either way; what the pin still assumes is that `aws_root_ca1_pem` is the *same* GoDaddy root this chain terminates in. Given the issuer above, that is the expected case, and if it ever isn't the failure is a clean handshake error rather than a silent downgrade.

Removing the pre-merge blocker from the PR body.

Issue #108, opened 2026-09-04
A station serial the account does not list is tiered wired forever, so a battery device's session never idle-closes


lenoxys:
Found while grilling #102. Distinct from the tier-staleness fixed there (which re-resolves the tier on each idle-arm): this one is a case where the tier is not stale but simply unanswerable, and the fallback is the draining one.

`stationPower` keys strictly on a device whose own serial equals the station serial, and answers `"wired"` when there is no such record:

```ts
// src/client/eufy-mega.ts:1676-1688
private stationPower(parentSn: string): PowerTier {
  const d = this.registry.list().find((x) => x.sn === parentSn);
  if (!d) return "wired";
  ...
}
```

`"wired"` means `Infinity` — no idle timer is ever armed, and the entry is never dropped by the idle path:

```ts
// src/transport/p2p/session-manager.ts:113-114
const idleMs =
  powered === "battery" ? (this.opts.batteryIdleMs ?? BATTERY_IDLE_MS) : Infinity;
```

```ts
// src/transport/p2p/session-manager.ts:217-222
private armIdle(parentSn: string, e: SessionEntry): void {
  e.idle.cancel();
  if (!Number.isFinite(idleMs)) {
    this.logger.debug(`[session ${parentSn}] idle (0 users) — staying persistent`);
    return;
  }
```

Meanwhile `openStation` explicitly tolerates a station serial with no record of its own, and builds the session from a **child's** `p2pDid`:

```ts
// src/transport/p2p/command-router.ts:312-318
private async openStation(parentSn: string): Promise<P2PSession> {
  return this.manager.acquire(parentSn, async (register) => {
    const devs = this.deps.listDevices();
    const stationDev = devs.find((d) => d.sn === parentSn) ?? devs.find((d) => this.stationKeyFor(d) === parentSn);
```

That second `find` is proof-by-code that the case is expected. The station key comes from `stationKeyFor` (`:279-283`), which returns `raw.parent_sn` or `dev.stationSn` whenever either is set — neither of which need appear as a device record.

## Call path

```
EufyMega.commandSinkFor(sn).dispatch          eufy-mega.ts:996-1001
  → routeCommand → p2p.dispatchCommand        eufy-mega.ts:1104 / 1116
    → P2PCommandRouter.resolveSession         command-router.ts:1132
      → deviceFor                             command-router.ts:443-446
          list non-empty → no reload; dev found
          openStation(stationKeyFor(dev))      → a station serial with no record
[...]

Issue #107, opened 2026-09-04
resetWhenUnused() awaits forever while a stream stays attached, hanging the next write to the same member


lenoxys:
Found while grilling #102; pre-existing on `beta-0.1.0`, filed separately.

`resetWhenUnused` parks an unbounded promise when active consumers remain:

```ts
// src/transport/p2p/session-manager.ts:253-262
async resetWhenUnused(parentSn: string): Promise<void> {
  const entry = this.entries.get(parentSn);
  if (!entry) return;
  if (entry.users <= entry.holds) {
    await this.close(parentSn);
    return;
  }
  entry.resetPending = true;
  return new Promise<void>((resolve, reject) => entry.resetWaiters.push({ resolve, reject }));
}
```

There is no timer on that promise. It settles only via `settleResetWaiters` (`:265-271`), reached from `remove()` (`:243`) or `closeEntry()` (`:277`, `:279`). With `resetPending` set and `users > holds`, no idle timer is armed either — `releaseUser` calls `armIdle` only at `users === 0` (`:186`). So the promise settles only when the last consumer releases, or on `closeAll()`.

That is defensible as a contract on its own. The problem is where it is awaited.

## The await is inside the serialized write queue

```ts
// src/client/eufy-mega.ts:1026-1034
const transaction = this.enqueueStateTransition(key, async () => {
  try {
    if (epoch !== this.realtimeEpoch) throw new Error("observed command superseded by disconnect");
    await this.routeCommand(sn, cmd);
    acknowledge();
    const refreshed = await this.refreshEventState(sn, observation);
    if (!refreshed || epoch !== this.realtimeEpoch) return;
    this.emitSemantic(observation.event, { deviceSn: sn });
    if (observation.resetStandaloneSession) await this.p2p.resetStandaloneSession(sn);
```

`enqueueStateTransition` serializes on `${sn}:${param}` (`eufy-mega.ts:773-788`), and a subsequent dispatch for the same param awaits its predecessor **before it routes** (`:776-779`). So while a viewer stays attached to a standalone device's live stream, the next write to that member never even reaches the wire.

The chain: `resetStandaloneSession` (`command-router.ts:296-301`) awaits `manager.resetWhenUnused(station)` with no timeout, and the flag is set by exactly one member — `src/model/capabilities/arming.ts:290`.

## Reproduction

1. Start a live stream on a standalone battery camera (a session consumer via `onActive` → `addUser`, `command-router.ts:757`).
[...]

Issue #106, opened 2026-09-04
A manager-initiated session close tears down nothing: the live source, the talkback and the p2pClose event are all skipped


lenoxys:
Found while grilling #102; pre-existing on `beta-0.1.0` and unrelated to that issue's refcount defect, so it is filed separately.

`P2PCommandRouter`'s session `close` handler guards on session identity:

```ts
// src/transport/p2p/command-router.ts:377-390
session.on("close", () => {
  if (this.manager.get(stationSn) !== session) return;
  this.manager.remove(stationSn);
  for (const key of [...this.liveSources.keys()]) {
    if (key.startsWith(`${stationSn}:`)) this.dropLiveSource(key);
  }
  for (const [key, talk] of this.talkbacks) {
    if (key.startsWith(`${stationSn}:`)) {
      this.talkbacks.delete(key);
      void talk.stop().catch(() => {});
    }
  }
  this.deps.onClose(stationSn);
});
```

`SessionManager.close()` discards the entry **before** awaiting the session close:

```ts
// src/transport/p2p/session-manager.ts:247-250
async close(parentSn: string): Promise<void> {
  const entry = this.discard(parentSn);
  if (entry) await this.closeEntry(entry);
}
```

and `P2PSession.close()` emits `"close"` as its last statement (`p2p-session.ts:1838`). By then the entry is gone, so `manager.get(stationSn)` is `undefined`, the guard is satisfied, and the handler **returns at line 378** — skipping all four of `manager.remove`, the `dropLiveSource` loop, the talkback teardown, and `deps.onClose`.

That is every manager-initiated close: `onIdle` → `close` (`session-manager.ts:237`), `resetWhenUnused` → `close` (`:257`, `:181`), `onLiveStartFailed` (`command-router.ts:894`), `replaceUnreachableSession` (`:915`), and `closeAll`. The only close that satisfies the guard is a session closing itself while still registered — the connect timeout at `p2p-session.ts:666-671`, which is effectively never.

The handler's own JSDoc states the contract the guard prevents:

```
command-router.ts:342-343
 * The `close` handler drops the session from the SessionManager and disposes any shared live
 * source riding this station (consumers get `stop`; a later attach rebuilds via the factory).
```

## Impact

**1. A cached `SharedLiveSource` survives the close and is reused over the dead session.**

`SharedLiveSource` never holds a session — the router holds it in a `HeldSession` cell captured **once** at source construction:

```ts
// command-router.ts:738-741
[...]

PR #110, opened 2026-09-04
fix(p2p)!: give an expiring hold to the entry that took it, and re-read a station's power per idle-arm

### LINKED ISSUE #102: SessionManager: a hold()'s expiry timer outlives its entry and steals the successor entry's user refcount
Found while reviewing #97; the code is pre-existing on `beta-0.1.0` and untouched by that PR, so it's filed separately as requested there.

In `src/transport/p2p/session-manager.ts`, `hold()` arms a raw `setTimeout` keyed only by `parentSn`:

```ts
hold(parentSn: string, ms: number): void {
  const entry = this.entry(parentSn);
  this.addUser(parentSn);
  entry.holds += 1;
  setTimeout(() => this.releaseHold(parentSn), ms).unref?.();
}

private releaseHold(parentSn: string): void {
  const entry = this.entries.get(parentSn);
  if (!entry) return;
  entry.holds = Math.max(0, entry.holds - 1);
  this.releaseUser(parentSn);
}
```

The timer survives its entry: if the entry is discarded (via `close()` / `resetWhenUnused`) and a **new** entry is later created for the same station, the stale timer's `releaseHold` acts on the successor. `holds` floors at 0, but `releaseUser` is called unconditionally — so the stale timer decrements a user count it never incremented on this entry.

**Failure scenario**

1. `bumpCommand(sn)` arms a 15s hold (`users=1`, `holds=1`).
2. `resetWhenUnused(sn)` sees `users <= holds` and closes, discarding the entry.
3. A live stream re-acquires the station; `addUser` puts `users=1` on the fresh entry.
4. 15s after the original command, the stale timeout fires `releaseHold` on the fresh entry: `holds` floors at 0 but `releaseUser` drops `users` 1 → 0, arming the battery idle timer.
5. The session is idle-closed while the stream's consumer is still attached; the consumer's eventual real `releaseUser` underflows to a no-op.

**Fix**

Capture the entry (or a generation token) in the timer closure and no-op when it no longer matches, e.g.:

```ts
hold(parentSn: string, ms: number): void {
  const entry = this.entry(parentSn);
  this.addUser(parentSn);
  entry.holds += 1;
  setTimeout(() => {
    if (this.entries.get(parentSn) !== entry) return; // entry was replaced; this hold died with it
    this.releaseHold(parentSn);
  }, ms).unref?.();
}
```

(Alternatively track the pending hold timers on the entry itself and cancel them when the entry is discarded.)

lenoxys:
Closes #102. Design decisions recorded in https://github.com/mega-yfue/eufy-sdk/issues/102#issuecomment-5537689144.

## The reported defect

A hold armed a bare `setTimeout` keyed only by serial, and `discard()` cancelled only `entry.idle`. An expiring hold therefore outlived its entry and released a consumer counted by whatever entry next occupied that serial — arming an idle-detach underneath a live viewer, whose own release then underflowed to nothing.

## What the issue missed

**`prewarm` had the same shape and a worse version of it.** 28 s instead of 15 s, and it incremented no hold at all, so its stale release was a bare consumer release. Meanwhile `hold()`'s own JSDoc claimed to be *"the primitive behind command-keepalive + event pre-warm"* while having exactly one caller.

So the fix is not a guard on one closure — it is one primitive with one invalidation rule.

## Changes

**`prewarm` routes through `hold()`, twice.** Once before the open so a slow connect cannot idle-close mid-flight, and again once the session lands. The second hold is not redundant: `command-router.ts` documents the warm window as running from a session being *ready*, not from the attempt, so a single hold at call time would have silently shortened it by the connect duration. Two expiring holds express that with no new API — exactly as a command burst already re-holds.

This also makes a pre-warm **ignorable by `resetWhenUnused`**. That is the intent — `holds` means "reasons that need not be waited for", and a speculative pre-warm with nothing attached qualifies just as a command keepalive does. It previously blocked a reset for up to 28 s. Behaviour change beyond the reported bug, called out rather than smuggled in.

**`holds: number` → `holdTimers: Set<Timeout>` on the entry, cleared by `discard()`.** The size *is* the count, so the second representation that drifted is gone. The invariant is now stated in `discard`'s JSDoc: **a discarded entry owns no live timer.** Chosen over capturing the entry in the closure, which would leave a timer running for up to 28 s to wake and do nothing.

**A release with nothing counted is refused, not clamped.** `Math.max(0, n - 1)` looked harmless but fell through to `armIdle`, so every unmatched release restarted a battery station's **full 300 s window** — and could complete a deferred reset a real consumer had not yet earned. Now it warns and returns.

**The power tier is re-resolved per idle-arm.** It was snapshotted at entry creation and never re-read, while `poweredFor` is a live lookup. A station whose battery evidence was not yet in the cached device list at first contact was pinned `wired` — infinite window, never sleeps — for the client's life, even after the parameter arrived. `armIdle` is the only consumer of the window, so that is where it is resolved.

**`wiredIdleMs` removed.** Declared, read once, never set by any caller, no public option, no test. The wired branch was always `Infinity`.

**`users`/`addUser`/`releaseUser` → `consumers`/`addConsumer`/`releaseConsumer`.** `SessionManager` is not exported (the p2p barrel exports only `type PowerTier`), so this is internal with no public surface. In a codebase where "user id" means account PII, a counter named `users` meaning "reasons to stay connected" reads as the wrong thing.

**`SessionSupersededError`** replaces the bare `Error` thrown by `acquire`, and `prewarm` swallows only that. Without it, a reset closing a station mid-pre-warm emits an error to the host for entirely correct behaviour — and matching on the message string would silently stop working the first time someone rewords it.

## Tests

Six new cases, all behavioural with fake timers. Each one was verified to **fail with its specific defect re-introduced** and pass without it:

| Re-introduced defect | Cases that fail |
|---|---|
| `discard()` not clearing `holdTimers` | the stale-hold scenario, its pre-warm twin, and the concurrent-holds case (3) |
| clamp-and-continue in the release path | the extra-release-does-not-restart-the-window case (1) |

Plus: a release with nothing counted resurrects no timer, and a late tier change lets the device sleep on the next idle-arm.

No counter accessor and no `as unknown as` cast — asserting the numbers would pin an implementation detail this change deletes, whereas asserting *the live viewer's station stays open* states the actual harm.

## Verification

`npm run verify` green (3406 tests). `npm -C docs run build` green including `guard:docs`. No prose in `docs/`, `examples/` or `README.md` referenced the renamed or removed symbols.

## Split out, not fixed here

#106, #107 and #108 came out of grilling this issue and are deliberately separate — #106 and #107 are host-visible in their own right, and #108 needs a device-model decision.


martijnpoppen (approved):
Reviewed the diff and the full resulting `session-manager.ts`. The core invariant — a discarded entry owns no live timer — is enforced correctly: `discard()` clears both the idle timer and every `holdTimers` entry before the map delete, so a hold that was mid-flight when its entry closed can never land on a successor entry for the same serial. The reset-vs-hold distinction (`consumers <= holdTimers.size`) and the refusal (not clamp) of an unmatched `releaseConsumer` both check out against the described failure modes, and the six new tests map cleanly onto each one (verified by reading the assertions against the logic, not just trusting the table in the description).

Traced through the two-hold `prewarm()` sequencing (hold-before-open, hold-after-ready) for the case where a connect takes longer than `ms`: the first hold's timer can expire mid-connect, but that only arms the idle window (5 min default for battery, infinite for wired) rather than closing anything, and the second `hold()` after connect completes cancels it via `addConsumer`. So the described protection holds in practice even though the first hold is no longer literally unbounded like the old `addUser`-before/`releaseUser`-after-ms shape — a reasonable, disclosed tradeoff, not a regression.

`wiredIdleMs` removal and the `users`→`consumers` rename are both clean, no leftover references. Looks good to merge.

---

martijnpoppen (approved):
Re-reviewed after the new push. The `prewarm()` simplification from two holds to one is a real behavior change from the earlier revision I approved, so I traced it rather than taking the JSDoc's word for it: `openStation()` → `SessionManager.acquire()`'s factory calls `await session.connect()`, and `P2PSession.connect()` (p2p-session.ts) resolves right after binding the UDP socket and firing the initial lookup packets — it does not wait for the peer to answer (that's tracked separately via `connected`/`level2Ready`). So a hold taken again immediately after `openStation()` returns would arm at essentially the same wall-clock moment as the pre-open hold, confirming "both would expire together" and that dropping the second hold isn't a reintroduction of the connect-duration-shortening problem the original two-hold design was written to avoid. Good catch simplifying that.

Everything else (hold-timer ownership via `discard()`, the refuse-not-clamp release, per-idle-arm power resolution) is unchanged from the version I already verified. Still good.

---

lenoxys:
Renamed the session refcount to `retain`/`release` (was `addConsumer`/`releaseConsumer`).

Reason: `command-router.ts` imports `type Consumer` from `shared-live-source` for a viewer attached to a stream, and then called `manager.addConsumer` for a command keepalive. One file, one word, two meanings — `addUser` didn't have that problem, so the rename had traded one collision for another.

`retain`/`release` is the standard refcount vocabulary, shorter than what it replaced, and reads correctly next to `hold()`: `retain()` is a reason to stay open, `hold(ms)` is a reason that expires on its own. The field follows, so the reset condition now reads `retained <= holdTimers.size`.

Also dropped the last of the rationale left over from the earlier design — the refusal was documented as guarding "a deferred release whose own entry was already discarded", which `discard()` clearing its hold timers had already made impossible. The guard stands; the stated cause didn't.

PR #111, opened 2026-09-04
fix(client): stop a write waiting on a viewer for the session recycle it asked for

### LINKED ISSUE #107: resetWhenUnused() awaits forever while a stream stays attached, hanging the next write to the same member

lenoxys:
Closes #107.

## The defect

An arming-mode write on a standalone device asks for its P2P session to be recycled, and awaits that **inside** the keyed state transition:

```ts
// eufy-mega.ts:1034 (before)
if (observation.resetStandaloneSession) await this.p2p.resetStandaloneSession(sn);
```

`resetWhenUnused` deliberately waits for every viewer to detach, so a write cannot drop a live stream. That wait has no upper bound — a viewer may watch indefinitely — and `enqueueStateTransition` serialises on `${sn}:${param}`, with a following dispatch awaiting its predecessor *before it routes*. So while someone is watching, the next write to that member never reaches the wire.

The write that asked is unaffected: it is answered at `acknowledge()`, before the recycle. The hang lands on the successor.

## The fix

The wait is bounded; the recycle is not cancelled. It stays pending and still runs when the station falls idle — it just stops gating an unrelated write. A failure still propagates; only a wait that has outlived its usefulness is abandoned.

A recycle nothing is holding completes in the time one session teardown takes, so the bound is never reached in the ordinary case. That is why the existing serialization test (`catch-all-event.spec.ts`, which gates the recycle on a manual promise and asserts the successor waits for it) **passes unchanged** — the intentional serialization is preserved.

## Why not the alternatives

- **Timing out inside `resetWhenUnused` and closing anyway** would drop the viewer's stream to satisfy a write that had already succeeded. That is precisely the outcome the deferred reset exists to avoid.
- **Floating the await** would let a following write race the recycle the flag was added to guarantee. The guarantee is kept for the common case; only the pathological wait is capped.

## Test

One new case, verified to fail with the unbounded await restored (it hangs and times out at 5005 ms) and pass with the bound: a recycle that never settles must not stop the queued write from reaching the wire.

## Verification

`npm run verify` green (3401 tests).

## Note

`SESSION_RECYCLE_WAIT_MS` is a fixed internal constant, not an option — it is a bound on an internal wait, not a policy a caller should tune. Easy to expose later if a real case wants it.


src/client/eufy-mega.ts
```diff
@@ -787,6 +797,41 @@ export class EufyMega extends EventEmitter {
...
+      }, SESSION_RECYCLE_WAIT_MS);
+      timer.unref?.();
+      recycle.then(
+        () => {
+          clearTimeout(timer);
+          resolve();
+        },
+        (error) => {
+          clearTimeout(timer);
+          reject(error);
```
martijnpoppen: The PR description says "A failure still propagates; only a wait that has outlived its usefulness is abandoned" — but that's only true when the recycle rejects *before* the 5s bound. If the timeout fires first (`resolve()` at line 819), the outer promise is already settled, so this `reject(error)` on an already-settled promise is a no-op and the error is dropped entirely: nothing logs it, nothing reaches `reportUnacknowledged`.

In the common case this is where it matters most too — the bound is only reached when a viewer is attached and the recycle is genuinely pending, so a later failure there (e.g. the eventual `close()` erroring) currently vanishes silently instead of surfacing anywhere. Worth at least logging it in the `(error) => { ... }` branch when the timer has already fired, e.g.:

```ts
recycle.then(
  () => { clearTimeout(timer); resolve(); },
  (error) => {
    clearTimeout(timer);
    if (settled) this.opts.logger?.error(`[session ${sn}] recycle failed after the write moved on`, error);
    else reject(error);
  },
);
```
(with a `settled` flag set in the timeout callback, or equivalent).

Not a blocker for the core fix — the bounded-wait behavior itself is correct and well tested — but it's a real gap versus the stated invariant.

---

martijnpoppen (commented):
Solid fix for the underlying hang, and the bounding logic itself (race timeout vs. actual completion) is correct — the existing serialization test passing unchanged is a good sign the intentional behavior is preserved. One gap flagged inline: a recycle failure that lands *after* the bound expires is silently dropped rather than propagating, which contradicts the PR's own stated invariant. Non-blocking for the hang fix, but worth addressing for observability.

---

martijnpoppen (approved):
The `Promise.race` rewrite addresses the earlier comment: a failure before the bound still propagates (Node attaches a handler to each raced promise internally, so no unhandled-rejection risk either), and a failure arriving after the bound is now at least logged as debug rather than vanishing with zero trace. The JSDoc is honest about the remaining shape ("survives only as the session manager's own log") — worth noting that's only strictly true for the deferred-reset path (`releaseConsumer`'s own `.catch` in session-manager.ts logs it there); the immediate-close branch of `resetWhenUnused` has no separate log if it happens to take >5s to fail. That's a narrow enough corner (it'd require the close itself to hang) that I don't think it's worth blocking on. Approving.

---

lenoxys:
Fixed the waiter accumulation I'd flagged and then decided to leave. On reflection the decision was wrong, for a reason worth stating: I'd priced it as *adding* code for a bounded leak. It's a net deletion.

**Why it belongs in this PR.** Before the bound, the keyed transaction queue meant write N+1 couldn't start until write N's reset had settled — so at most one caller was ever parked. Releasing the write is what lets the next one call `resetWhenUnused` again while the first reset is still pending. This PR made it reachable, so this PR fixes it.

**Why it deletes rather than adds.** `settleResetWaiters` gave every entry in the array the *same* outcome, so N waiters were N copies of one signal. The outcome is a property of the station's teardown, not of who asked for it. One deferred per entry says exactly that and cannot grow with the number of callers — `Promise.withResolvers` (ES2024, already the target) replaces the array, the splice-and-loop drain, and `resetPending`, whose only question was whether a reset was pending. `reset !== undefined` answers it.

**One thing that fell out.** `resetWhenUnused` was `async`, so it returned a fresh wrapper on every call — which hid the sharing from the outside and made it untestable. Dropping `async` is shorter *and* makes the invariant observable, which is what the new case asserts: restore a deferred per caller and it fails.

Pre-existing reset behaviour is untouched — the per-entry isolation case (`closeAll` rejects A with its own failure while B resolves) passes unchanged. Verify green, 3402.

PR #112, opened 2026-09-04
fix(p2p): announce the closes nobody asked for, so nothing is handed a dead session

### LINKED ISSUE #106: A manager-initiated session close tears down nothing: the live source, the talkback and the p2pClose event are all skipped

lenoxys:
Closes #106. **Stacked on #110** (`fix/session-hold-lifetime`) — it touches the same regions of `session-manager.ts`. Base is that branch; retarget to `beta-0.1.0` once #110 merges.

Corrected design in https://github.com/mega-yfue/eufy-sdk/issues/106#issuecomment-5537883379.

## The defect

A `SharedLiveSource` never holds its own session — the router captures one in a `HeldSession` cell when the source is built (`command-router.ts:738`) and never re-resolves it. A source left cached past that session is handed to the next viewer over a dead connection: it answers the retained keyframe, then fails on the warm-up deadline 20 s later. Indistinguishable from a camera that has stopped working. Talkbacks are the same shape.

The `close` handler drops both, but it is guarded on the session still being the station's registered one — and `SessionManager.close()` discards the entry **before** awaiting the teardown. So on every close the manager initiated, the guard is already false and the handler does nothing. The only close that ever reached it is a session dying while still registered, in practice just the connect timeout.

Consequences: a stale source reused over a dead session, a talkback never stopped, and `p2pClose` — a declared public event — never firing for an idle-detach while `p2pConnect` fires on every open.

## Why the obvious fix is wrong

The issue originally proposed relaxing the guard to `get(sn) !== session && has(sn)`. I disproved it before writing it:

```ts
// replaceUnreachableSession, command-router.ts:923-937
await this.manager.close(station);              // emits "close" during this await
const { session } = await this.resolveSession(sn);
if (this.liveSources.get(key) !== undefined) {  // ← now undefined
  held.session = session;                        // ← never runs
  this.liveSources.get(key)?.rewarm();
```

`P2PSession.close()` emits `"close"` before `manager.close()` resolves, so the relaxed guard fires mid-teardown and disposes the very source that method is about to rewarm. The whole "replace an unreachable session and warm again on it" recovery dies, and **nothing would have caught it** — `live-source-unreachable-session.spec.ts` tests `SharedLiveSource` in isolation and never touches the router.

## The actual asymmetry

Every close a **caller** makes already cleans up, and they differ in what they want kept:

| Path | Live source |
|---|---|
| `closeAll` | disposes all sources first, then closes |
| `onLiveStartFailed` | drops in its `.finally` |
| `replaceUnreachableSession` | **keeps it on purpose**, to rewarm |
| `acquire` supersede | belongs to the successor |

The two with no cleanup are the ones with no caller: an elapsed idle window, and a deferred reset falling due. The manager now announces **those and only those**, via `SessionManagerOpts.onAutoClose`. The guard is untouched.

## Ordering

The announcement fires the moment the entry is discarded — before the socket teardown and before any deferred reset settles. Both deliberate: from that instant a fresh acquisition resolves a new session while a cached source still points at the old one, so announcing later leaves a window in which a viewer can attach to a stale source; and a caller awaiting a reset should find the riders already dropped when it resumes.

I had this wrong first (announcing in a `finally` after the close), and the deferred-reset test caught it — the reset promise resolved before the teardown.

## Behaviour change

`p2pClose` now fires for an idle-detach, restoring the pair a host needs. It stays quiet for internal recoveries that immediately reopen, which is the event churn a blanket fix would have produced. `types.ts:300-301` already establishes that transport lifecycle never touches availability, so nothing downstream reinterprets it.

## Test

`autonomous-session-close.spec.ts` is the first spec to drive a station close through a **real** `SessionManager` and `P2PCommandRouter`. The handler previously had zero coverage. It pins both directions:

- **3 cases fail** if the announcement is not wired (idle-detach drops the source, the station is reported closed, a deferred reset drops it too);
- **1 case fails** if a caller's own close announces as well — i.e. it catches exactly the `replaceUnreachableSession` regression the rejected fix would have introduced.

## Verification

`npm run verify` green (3411 tests).


martijnpoppen (approved):
Reviewed against the full `session-manager.ts` on this branch (stacked on #110). The `autoClose()` → `tearDownStation()` wiring is idempotent by construction: `discard()` removes the map entry before `onAutoClose` fires, so the router's `remove()` call inside `tearDownStation` is a guaranteed no-op on that path, and the later `session.close()` → `"close"` event is blocked by the `manager.get(stationSn) !== session` guard since the entry is already gone. Traced both directions — a caller-initiated `close()`/`closeAll()` never touches `onAutoClose`, and the four "leaves it alone" test cases match that. The ordering claim (announce before socket teardown, before reset waiters settle) also checks out: `onAutoClose` runs synchronously inside `autoClose()` before the `await closeEntry(entry)` that eventually calls `settleResetWaiters`.

The disproof of the originally-proposed fix (relaxing the guard) is correct too — `replaceUnreachableSession` really would race the guard mid-teardown since `P2PSession.close()` emits `"close"` before `manager.close()` resolves, and this PR's approach sidesteps that entirely by not touching the guard.

Good test coverage for a previously-uncovered handler. Approving.

---

martijnpoppen (approved):
Re-checked after the rebase onto #110's latest commit. Confirmed with `git merge-base --is-ancestor` and a direct diff that this branch is cleanly stacked on #110's current tip (no divergence, e.g. `prewarm()` matches #110's one-hold version exactly) — no merge-conflict risk from the stacking. The `onAutoClose`/`tearDownStation` logic itself is unchanged from what I already verified (idempotent discard-then-announce ordering, guard correctly skips the caller-close paths). Still good.

---

lenoxys:
Merged #110 in — the resolution keeps both sides: `retained` from #110 and `autoClose` from here. Verify green (3408), and the three defects are still pinned individually: unwiring the announcement fails 3 cases, routing the immediate reset back through `close()` fails 1, dropping the `entry.session` guard fails 1.

PR #114, opened 2026-09-04
feat(rtsp): read the camera-reported RTSP URL as inbound state

martijnpoppen:
## What changed

A camera's **own RTSP URL** — host, path, and the credentials it is enforcing right now — becomes a
readable property, so a host can adopt the running stream instead of assembling an address or imposing
its own credentials.

- **`rtsp.decodeState`** lifts the URL out of the station's push. The station answers the publish
  switch (1145) by pushing the same id BACK as a data frame whose string payload is the full
  `rtsp://user:pass@host/path` — a bare string, not the `params` array the transport unwraps
  generically, so without this the URL is announced on the wire and never reaches a getter.
- **`url` member** (flat property `rtspUrl`, read as `dev.rtsp()?.url`) — read-only, P2P-notify only,
  absent until a push lands, then it reads back like any other state. It lands under a **synthetic id**
  (`STREAM_URL`, 11450) because the wire supplies no second id: 1145 already belongs to the `published`
  bool. `property-id-integrity` quarantines it like 1612.
- The push **also sets `published` true**. It is the device stating its own state — the best evidence
  of publication the wire offers, and a poll sooner than the cloud. `published` is device state, not a
  record of who turned it on; a caller needing the latter tracks its own `publish()` call.
- **`docs/rtsp.md`** documents the arrival contract, which is the part a host gets wrong: absent until
  the push lands (after the write resolves, not with it), announced as `rtspUrl` only when the value
  MOVES, never retracted by `withdraw()`, and carrying a live password that therefore rides
  `deviceState` and `propertyChanged`.
- Specs share one router harness in `session-fixtures.ts` instead of two near-copies.

Nothing in `transport/` changes. No new public facade method, no provider, no event.

## Why

The device does not report its RTSP credentials anywhere durable — they regenerate whenever
publication is toggled, and the cloud record trails a cycle behind. The only source of the current
pair is the push the station sends on the publish switch's own wire. That push already reaches
`onP2PFrame` with its bytes intact, so per `AGENTS.md` ("if the data already arrives on an inbound
path, decode it there and let it flow in as state") the URL is a capability decode, not a read.

Earlier iterations of this PR did it as a facade `reportedRtspUrl(sn)` over a P2P read plus a cloud
record miner. Both are gone: the read was byte-identical to `publish()` when the switch was off, so a
caller who believed they were reading silently turned on a persistent device setting — with nobody
owning `withdraw()`. Moving it inbound dissolved that instead of documenting it.

## Evidence

**Frame shape — observed.** Confirmed end to end on a standalone camera: publication on, the station
pushed `rtsp://user:pass@host/path` back on 1145 as a NUL-terminated string, an authenticated
`DESCRIBE` against that URL returned a picture. That capture is what `decodeState` is written from.

**The provoke — inferred, and marked so in the JSDoc.** The capture that observed the frame also sent
`CMD_NAS_TEST` (1146), which this PR no longer sends. So "`publish()` alone elicits the push" has not
been seen in isolation. It fails safe — no push means the property never appears, nothing misreports —
and if a camera turns out not to push on the switch alone, the answer is a `MediaProvider` pull rather
than anything here. Confirming it needs a device this environment does not have.

HomeBase-3 attached cameras did not serve on current firmware. Upstream, not addressed here.

## Verified how

- [x] `npm run verify` is green locally (182 files, 3429 tests)
- [x] Ran against a real device — a standalone camera streamed over its device-reported URL, on the
      earlier wire (1145 + 1146). The current wire's provoke is unverified; see Evidence.
- [x] Added or updated specs (offline, synthetic fixtures — the pushed URL frame, the typed read-back,
      the shared router harness)
- [x] Docs updated — `docs/rtsp.md` gains "The device-reported URL"

## Unverified or deferred

- **That `publish()` alone provokes the push** — see Evidence. Read path only; nothing writes on a
  guess, and a URL that never arrives is a getter that never installs.
- **HomeBase-3 attached-camera RTSP** is a firmware-side gap; re-test after a HomeBase update rather
  than working around it in the SDK.
- **11450 is a free number, not a reserved band.** It is the only synthetic id today. If a second is
  ever needed, reserve a stated band rather than copying this per-site choice.


src/transport/p2p/lan-ip.ts
```diff
@@ -47,3 +47,55 @@ export function freshestLanIp(raw: unknown): string | undefined {
...
+ * The vendor app keeps referring to "the device-reported URL" as the thing NAS storage is verified
+ * against, which implies the device publishes one — the open question (see the publish switch,
+ * `CMD_NAS_STREAM_SWITHC` 1145, and its read-only neighbour `CMD_NAS_TEST_STREAM` 1146) is where.
+ * Params the device volunteers land in the cloud record with timestamps, the same channel
+ * {@link freshestLanIp} mines for addresses — so until a live read of 1146 is captured, the record
+ * is the one place a reported URL could already be sitting. This scans every string param for an
+ * `rtsp://` URL and returns the newest; a consumer should treat it as a HINT to verify (a
+ * `DESCRIBE` costs one round trip), not an address to trust blindly, exactly like the record's IPs.
+ */
+export function reportedRtspUrl(raw: unknown): string | undefined {
```
lenoxys: **This function is the previous iteration of this PR, kept beside the new one.**

The PR body says the live P2P read is *"the only source of the freshly-generated credentials"* and that the cloud copy *"lags a cycle"*. So this returns a URL whose embedded credentials don't authenticate — which isn't a fallback, it's a wrong answer that looks like a right one.

The JSDoc still says *"until a live read of 1146 is captured, the record is the one place a reported URL could already be sitting"*. That live read **is** captured, in this same PR.

`AGENTS.md`: *"Remove speculative and dead code rather than keeping old shapes beside new ones."* Deleting it takes the spec (−51), the import in `eufy-mega.ts`, and the try/catch in the facade with it.

One more signal that this path rests on a guess: `1146` is a **command** id in `command-router.ts` and a **param** id in the fixtures here. `AGENTS.md` says model and transport use disjoint id subsets — a wire id being both is usually a sign it was inferred rather than observed.

src/transport/p2p/lan-ip.ts
```diff
@@ -47,3 +47,55 @@ export function freshestLanIp(raw: unknown): string | undefined {
...
+    | {
+        params?: Array<{ param_value?: unknown; update_time?: unknown }> | Record<number | string, string>;
+        paramUpdatedAt?: Record<number | string, number>;
+        dpParams?: Record<number | string, string>;
+      }
+    | undefined;
+  // The device's own realtime report outranks every cloud snapshot: `dpParams` is what the station
+  // volunteered over its live wire, which is where a just-regenerated URL lands first — the cloud
+  // copy follows on the device's own schedule.
+  for (const value of Object.values(rec?.dpParams ?? {})) {
```
lenoxys: Three record shapes handled (cloud `{param_type, param_value, update_time}` array, merged `params` + `paramUpdatedAt`, and `dpParams`) for one caller, which passes exactly one of them (`registry.record(sn)`).

Also the same `rtsp://` regex is compiled twice inside the one function, L76 and L96.

Both moot if the function goes.

src/client/eufy-mega.ts
```diff
@@ -1235,6 +1236,32 @@ export class EufyMega extends EventEmitter {
...
+    // existing — a viewer opening the tile wants the stream anyway, and adoption at connect time
+    // races session warm-up. Any failure (no route, level-2 not ready, no push in the window)
+    // falls back to mining the record.
+    try {
+      const live = await this.p2p.readReportedRtspUrl(sn);
+      if (live) return live;
+    } catch {
+      /* fall back to the record */
+    }
+    return reportedRtspUrl(await this.registry.record(sn));
```
lenoxys: With the record miner gone this whole method is `return this.p2p.readReportedRtspUrl(sn);`.

Worth keeping as the public door — transport is internal — but it doesn't need the try/catch: `readReportedRtspUrl` already swallows everything and resolves `undefined`, so the catch here is unreachable.

src/client/eufy-mega.ts
```diff
@@ -1235,6 +1236,32 @@ export class EufyMega extends EventEmitter {
...
+   * none.
+   *
+   * Read-through, not roster: this goes via the registry's per-device record fetch (the live
+   * `get_device_param_list` merged over the account list), because the URL's embedded credentials
+   * are REGENERATED every time the vendor app toggles the publish switch — a cached roster keeps
+   * returning the previous pair, which authenticates nothing. Still a hint to verify, not a
+   * guarantee: the device pushes the new value to the cloud on its own schedule.
+   */
+  async reportedRtspUrl(sn: string): Promise<string | undefined> {
+    // The LIVE P2P read first — it is the only source of the freshly-generated credentials (the
```
lenoxys: Six lines of `//` narration inside the body, and the JSDoc directly above already carries most of it.

`AGENTS.md`, hard rule: *"JSDoc above the declaration ... do not narrate inside the body with `//` lines. A body comment is a smell that the JSDoc is incomplete — move it up, or delete it if the code already says it."*

Same pattern at `command-router.ts` L1243 and L1257, `lan-ip.ts` L68 and L76, `p2p-session.ts` L1796, and in the new spec. ~25 lines total.

src/transport/p2p/command-router.ts
```diff
@@ -1226,6 +1231,51 @@ export class P2PCommandRouter {
   }
 
   /** Send a level-1 int-plus-string frame with authenticated account identity injected by the transport. */
+  /**
```
lenoxys: **Botched insertion — the method above lost its JSDoc.**

L1233 is `/** Send a level-1 int-plus-string frame ... */`, which documented `sendIntString` at L1279. The new method was inserted between them, so `sendIntString` is now undocumented and `readReportedRtspUrl` has two doc blocks.

One line to move back.

src/transport/p2p/command-router.ts
```diff
@@ -1226,6 +1231,51 @@ export class P2PCommandRouter {
...
+   * Read the camera's LIVE authoritative RTSP URL — host, path, and the credentials it enforces
+   * RIGHT NOW — by writing CMD_NAS_TEST (1146) for the channel and awaiting the CMD_NAS_SWITCH
+   * (1145) DATA frame the station pushes back with the `rtsp://…` string. This is the only source
+   * of the freshly-generated credentials: the vendor app regenerates them on every publish toggle
+   * and the cloud record lags a cycle. `undefined` when no URL is pushed within the window.
+   */
+  async readReportedRtspUrl(sn: string, timeoutMs = 12000): Promise<string | undefined> {
```
lenoxys: Two things here.

**`timeoutMs = 12000` has no caller that passes it.** Inline the constant until one exists.

**The listener leaks when the deadline wins.** `onUrl` is only detached on the resolve path and in the catch — if `deadline` wins the race, it stays attached to a station session that outlives the call, so every read that times out adds one. `events.once` with `AbortSignal.timeout` detaches on abort for free, which would also collapse the four nesting levels here (`Promise.race` → async IIFE → `.catch` → `new Promise` → `try/catch`) down to roughly a third. The one thing it can't express is the per-channel filter — worth saying whether that filter is necessary or defensive, because if it's defensive this whole block is a couple of lines.

**No test.** The two specs cover the frame decode and the record miner. The logic that actually carries this feature — provoke, filter by channel, bound the wait — is untested.

src/transport/p2p/command-router.ts
```diff
@@ -197,6 +197,11 @@ export interface P2PRouterDeps {
   localAddresses?: Record<string, string>;
 }
 
+/** CMD_NAS_SWITCH (1145): the RTSP publish switch — writing 1 makes the station push its URL. */
+const CMD_NAS_SWITCH = 1145;
```
lenoxys: `CMD_NAS_SWITCH = 1145` is declared here and again at `p2p-session.ts` L222, each with its own JSDoc explaining the same wire. Two owners for one id in the same transport folder — pick one and import it.

src/__tests__/freshest_ip.spec.ts
```diff
@@ -55,3 +55,54 @@ describe("freshestLanIp — newest heartbeat wins over stale top-level fields",
     expect(freshestLanIp(undefined)).toBeUndefined();
   });
 });
+
+describe("reportedRtspUrl — a device-volunteered rtsp:// URL mined from the record's params", () => {
```
lenoxys: 51 lines of specs for the function I'd delete — they go with it.

Independent of that: `await import("../index.js")` is repeated in all five cases and can be a single top-level import. Pulling it from the barrel rather than the module also quietly asserts this is public API, which is worth being deliberate about.

src/transport/p2p/__tests__/rtsp-url-frame.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+  target.send = vi.fn();
+  const urls: { channel: number; url: string }[] = [];
+  session.on("rtspUrl", (ev) => urls.push(ev as { channel: number; url: string }));
+  return { urls, feed: (p: Buffer) => target.onData(p, ADDRESS), close: () => session.close() };
+}
+
+describe("the RTSP URL the station pushes on CMD_NAS_SWITCH", () => {
+  it("is emitted as rtspUrl with the channel and the full authoritative link", () => {
+    const { urls, feed, close } = harness();
+    // A NUL-terminated rtsp:// string — host, path, and the credentials the device enforces now.
```
lenoxys: No other spec in this folder has inline `//` comments — the sentence reads better as part of the `it` title.

The two cases here are the right minimum for the frame decode. It's `readReportedRtspUrl` that has nothing on it.

lenoxys (commented):
Read the whole thing. The frame decode in `p2p-session.ts` and the `rtspUrl` event are the right shape, and the evidence note (standalone works, HomeBase-3 doesn't, not worked around) is exactly what it should say.

The concern is size: +260/−0 for one read, and roughly 150 of those lines are the *previous* approach kept alongside the new one. `reportedRtspUrl` in `lan-ip.ts` mines the cloud record for a URL whose credentials this PR's own description says don't authenticate — and its JSDoc still reads as though the live P2P read hadn't been found yet. Deleting it, its spec, and the facade fallback takes the diff down to roughly the frame decode plus the read, which is the feature.

Two real defects flagged inline: the JSDoc of `sendIntString` was orphaned by the insertion, and the `rtspUrl` listener leaks on every read that times out.

Nothing here needs a redesign — it's mostly deletion.

martijnpoppen (commented):
removed

src/transport/p2p/lan-ip.ts
martijnpoppen: Agreed — deleted `reportedRtspUrl` from `lan-ip.ts` along with its 51-line spec, and the fallback call site in `eufy-mega.ts`. Fixed in 3916e7e.

---

src/transport/p2p/lan-ip.ts
martijnpoppen: Moot — gone with the function in 3916e7e.

---

src/client/eufy-mega.ts
martijnpoppen: Done — it's now `return this.p2p.readReportedRtspUrl(sn);`, no try/catch. Fixed in 3916e7e.

---

src/client/eufy-mega.ts
martijnpoppen: Folded all of these into JSDoc (or deleted where the JSDoc already covered it): `eufy-mega.ts`, `command-router.ts` L1243/L1257, `lan-ip.ts` (gone with the function), `p2p-session.ts` L1796. Fixed in 3916e7e.

---

src/transport/p2p/command-router.ts
martijnpoppen: Moved back — `sendIntString` has its comment again, `readReportedRtspUrl` keeps its own. Fixed in 3916e7e.

---

src/transport/p2p/command-router.ts
martijnpoppen: All three addressed in 3916e7e:
- `timeoutMs` is no longer a per-call parameter — moved to `P2PRouterDeps.rtspUrlReadTimeoutMs`, a construction-time test hook matching the existing `sessionIdle` pattern (the default still applies for every real caller).
- Rewrote around `events.once(session, "rtspUrl", { signal })` with one `AbortSignal` shared by `resolveSession` and the wait, which collapses the nesting the way you described and fixes the leak — `once`'s own abort handling deregisters the listener when the signal fires, so a timed-out read no longer leaves anything behind. The per-channel filter is necessary, not defensive: a HomeBase session multiplexes every attached camera's channel, so an unfiltered wait would resolve on another camera's push. Kept it, documented why in the JSDoc.
- Added `read-reported-rtsp-url.spec.ts`: the provoke-command shape, the channel filter (a wrong-channel push doesn't resolve the read), and — empirically confirmed vitest's fake timers don't intercept `AbortSignal.timeout`, so this one uses a short real timeout via the new test hook — that a timed-out read leaves `session.listenerCount("rtspUrl")` at 0.

---

src/transport/p2p/command-router.ts
martijnpoppen: Picked `p2p-session.ts` as the owner (exported it, it already carried the fuller JSDoc) and import it into `command-router.ts`. Fixed in 3916e7e.

---

src/__tests__/freshest_ip.spec.ts
martijnpoppen: The 51 lines are gone with the function. Nothing left importing `../index.js` repeatedly in this file since those were the only cases doing it.

---

src/transport/p2p/__tests__/rtsp-url-frame.spec.ts
martijnpoppen: Moved the sentence into the `it` title. `readReportedRtspUrl` now has its own spec (`read-reported-rtsp-url.spec.ts`) covering the provoke commands, the channel filter, and the listener-leak fix. Fixed in 3916e7e.

---

src/transport/p2p/command-router.ts
```diff
@@ -1225,6 +1233,44 @@ export class P2PCommandRouter {
...
+   *
+   * `undefined` on any failure: no route, level-2 not ready, or no matching push before the deadline.
+   */
+  async readReportedRtspUrl(sn: string): Promise<string | undefined> {
+    const log = this.deps.logger ?? noopLogger;
+    const signal = AbortSignal.timeout(this.deps.rtspUrlReadTimeoutMs ?? RTSP_URL_READ_TIMEOUT_MS);
+    try {
+      const { session, accountId, channel } = await this.resolveSession(sn, { waitLevel2: "soft", signal });
+      if (!accountId) return undefined;
+      session.sendIntStringCommand(CMD_NAS_SWITCH, 1, channel, accountId, channel);
```
lenoxys: This pins level 1. The existing publish path for 1145 goes through the level decision, which picks level 2 whenever the session holds a key — so on a keyed HomeBase this sends the seal the app never uses for that peer. Likely the actual cause of the attached-camera failure the description attributes to firmware. Route it through the same decision point and retest before calling it upstream.

Also skips the router's own int-string send, so no repeat — one datagram instead of five.

src/transport/p2p/command-router.ts
```diff
@@ -1225,6 +1233,44 @@ export class P2PCommandRouter {
...
+      session.sendIntStringCommand(CMD_NAS_TEST, 1, channel, accountId, channel);
+      for (;;) {
```
lenoxys: Re-arming after each non-matching event leaves a window with no listener attached. A matching push landing in that window is lost and the read times out — and the multiplexed-station case in the rationale above is precisely what produces back-to-back pushes.

One persistent listener with a channel filter, attached before the two writes, fixes this and the arm-after-send ordering at once. The two comparable readers in this file both arm first for that reason.

src/transport/p2p/command-router.ts
```diff
@@ -1225,6 +1233,44 @@ export class P2PCommandRouter {
...
+   *
+   * `session` is shared by every channel on the station (a HomeBase multiplexes its attached cameras
+   * over one session), so a push for another camera must not resolve this read — the loop re-arms the
+   * wait until either a matching push arrives or the deadline aborts it.
+   *
+   * `undefined` on any failure: no route, level-2 not ready, or no matching push before the deadline.
+   */
+  async readReportedRtspUrl(sn: string): Promise<string | undefined> {
+    const log = this.deps.logger ?? noopLogger;
+    const signal = AbortSignal.timeout(this.deps.rtspUrlReadTimeoutMs ?? RTSP_URL_READ_TIMEOUT_MS);
```
lenoxys: The signal reaches the connect poll and the event wait, but not the device resolution ahead of it — device load, station open, key fetch. Those carry their own HTTP timeouts so nothing hangs, but the worst case is well past the 12 s. Either thread the signal in or drop "end to end, `resolveSession` included" from the doc.

src/transport/p2p/command-router.ts
```diff
@@ -195,8 +196,15 @@ export interface P2PRouterDeps {
   sessionIdle?: Pick<SessionManagerOpts, "batteryIdleMs" | "wiredIdleMs" | "commandKeepAliveMs">;
   /** LAN address overrides for direct P2P, keyed by parent-station serial (host or host:port). */
   localAddresses?: Record<string, string>;
+  /** Override for {@link RTSP_URL_READ_TIMEOUT_MS} — a construction-time test hook, not a policy a caller tunes. */
+  rtspUrlReadTimeoutMs?: number;
```
lenoxys: Only caller is a test. Fake timers cover the same case without a permanent dep field.

src/transport/p2p/command-router.ts
```diff
@@ -195,8 +196,15 @@ export interface P2PRouterDeps {
...
 }
 
+/** CMD_NAS_TEST (1146): starts the RTSP livestream; also elicits the URL push on some firmwares. */
+const CMD_NAS_TEST = 1146;
```
lenoxys: Both ids already live in `commands.ts`. This makes a third declaration of the pair — use the existing one.

src/transport/p2p/p2p-session.ts
```diff
@@ -211,6 +211,16 @@ interface RetainedDatagram {
...
+/**
+ * CMD_NAS_SWITCH (1145) — the RTSP publish switch. The station ALSO pushes it back as a DATA frame
+ * whose NUL-terminated string payload is the camera's full authoritative `rtsp://user:pass@ip/path`
+ * URL, with the credentials it is enforcing RIGHT NOW. That live push is the only source of the
+ * regenerated pair: the vendor app rewrites the credentials on every publish toggle and the cloud
+ * record lags a cycle behind, so a host adopting a running stream must read the URL from here. This
+ * session emits that push as `rtspUrl`, `{ channel, url }`. Provoked by writing CMD_NAS_TEST (1146)
+ * for the channel (see the command router).
+ */
+export const CMD_NAS_SWITCH = 1145;
```
lenoxys: Only exported id among the ~18 locals in this file, and `export *` carries it to the root barrel, so 1145 becomes public API. Keep it `const`.

src/transport/p2p/__tests__/read-reported-rtsp-url.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+const DEVICE_SN = "T8114P0000000000";
+const STATION_SN = "T8010P0000000000";
+const ACCOUNT_ID = "0000000000000000000000000000000000000000";
+const CMD_NAS_SWITCH = 1145;
+const CMD_NAS_TEST = 1146;
+
+interface FakeSession extends FakeP2PSession {
+  sendIntStringCommand: ReturnType<typeof vi.fn>;
+}
+
```
lenoxys: `setup` here is a near-copy of the sibling int-string spec's. `session-fixtures.ts` is where the shared part belongs.

src/transport/p2p/__tests__/read-reported-rtsp-url.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+    await expect(read).resolves.toBe("rtsp://u:p@host/live0");
+  });
+
+  it("ignores a push for another channel on the same station session", async () => {
+    const { router, session } = setup();
+    const read = router.readReportedRtspUrl(DEVICE_SN);
+    await tick();
+
+    session.emit("rtspUrl", { channel: 2, url: "rtsp://wrong-camera/live0" });
+    await tick(); // let the wait loop re-arm its listener before the next push arrives
```
lenoxys: This `tick` is the listener gap, made explicit — the emits have to be spaced or the second is dropped. Once the wait holds a single listener, both pushes can land in the same turn and the tick goes away.

src/transport/p2p/__tests__/read-reported-rtsp-url.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+    const { router, session } = setup({ rtspUrlReadTimeoutMs: 20 });
+
+    await expect(router.readReportedRtspUrl(DEVICE_SN)).resolves.toBeUndefined();
+
+    expect(session.listenerCount("rtspUrl"), "a timed-out read must not leave a listener on the station session").toBe(
+      0,
+    );
+  });
+
+  it("leaves no listener behind on the success path either", async () => {
```
lenoxys: Asserts that the stdlib event helper deregisters its own listener on resolve, not our behaviour. The on-demand station open and the empty-account path are the uncovered ones — both blocked by the pre-registered session in `setup`.

lenoxys (commented):
Second pass. The record miner is gone, which was the right call. Two real changes left: the level choice on the provoke write (which I think explains the HomeBase gap the description hands to firmware) and the listener window in the wait. Rest is id ownership and test hygiene. Details inline.

martijnpoppen:
Second-pass review addressed in 9c2a0af:

**Behaviour**
- **Provoke level (`command-router.ts`).** Both provokes now go through `resolveScalarParam(sn, …, "auto")` — the same level decision the normal 1145 publish uses — instead of a pinned level-1 `session.sendIntStringCommand`. On a keyed HomeBase they ride the level-2 seal, and the shared path repeats the datagram for RF resilience. They're fired without blocking the read on the retransmits, so the URL still returns as soon as the push lands. 
- **Listener window.** One persistent, channel-filtered listener (`awaitRtspUrl`), armed before the provokes and torn down on resolve or abort — no re-arm gap, and a wrong-channel push no longer needs a spacing tick.

**Hygiene**
- **Signal scope.** JSDoc softened: the deadline covers the connect wait and the URL wait; device/station resolution relies on its own HTTP timeouts.
- **Test-only dep.** Timeout is now a `setTimeout`-driven `AbortController` (fake timers cover it), so `rtspUrlReadTimeoutMs` is gone.
- **Id ownership.** `CMD_NAS_SWITCH` / `CMD_NAS_TEST` come from `CommandType` in `commands.ts`; dropped the local `CMD_NAS_TEST` const and the `p2p-session.ts` `export const CMD_NAS_SWITCH` that was leaking 1145 to the root barrel.
- **Specs.** Shared the router harness in `session-fixtures.ts` (both the RTSP and int-string specs use it), dropped the tick, and added coverage for the level-2 wire, the empty-account path, and the unresolved-session path.

`npm run verify` green (3408 tests).

---

src/transport/p2p/command-router.ts
```diff
@@ -1267,6 +1271,84 @@ export class P2PCommandRouter {
...
+    const { signal } = controller;
+    try {
+      const { session, channel, accountId } = await this.resolveSession(sn, { waitLevel2: "soft", signal });
+      if (!accountId) return undefined;
+      // Arm the wait BEFORE provoking, so a push that arrives between the two writes is not lost.
+      const url = this.awaitRtspUrl(session, channel, signal);
+      // Provoke through the ONE level decision (and its RF-resilience repeat), but do not block the read
+      // on the retransmits finishing: the URL push can land after the first datagram, so the read
+      // returns as soon as it arrives (or the deadline aborts), while the repeats run to completion.
+      void this.resolveScalarParam(sn, CommandType.CMD_NAS_SWITCH, 1, "auto").catch((e) =>
```
lenoxys: **A method named like a read turns on a persistent device setting.**

This is byte-identical to what `rtsp.publish()` sends — `publishCommand(true)` → `setScalar(RTSP_PARAM.STREAM_SWITCH, 1)`, `model/capabilities/rtsp.ts:95`. So every constraint this repo already documents for `publish()` now applies to a caller who believes they are reading:

- "A station publishes for ONE attached camera at a time. Enabling a second withdraws the first, silently" (`rtsp.ts:75`). A NAS recording camera A stops recording because a host read camera B's URL. No error, no event, and that same doc says the SDK can neither detect nor prevent it.
- "It will drain a battery camera" (`docs/rtsp.md:67`).
- "The caller that enables publication therefore owns its lifetime" (`docs/rtsp.md:43`) — here nobody owns it, and the switch stays on after the read has forgotten it happened.

The JSDoc above says the read "never touch[es] the credentials themselves, so a NAS/NVR consuming the stream elsewhere is undisturbed". True of the credentials command; not true of the station's single-camera arbitration. And "idempotent when already on" is load-bearing: when it is off, this call **is** `publish()`.

I don't think the fix is code — you can't withdraw afterwards, because the toggle regenerates the credentials you just returned. The URL is only valid while published, which means this read *is* a publish. So either name and document it as one and hand the caller the `withdraw()` obligation, or put it next to `publish`/`withdraw` where that obligation is already written down.

This has been here since 14488b5 and I missed it twice; it isn't a regression from the last commit.

src/transport/p2p/command-router.ts
```diff
@@ -203,6 +204,9 @@ export interface P2PRouterDeps {
   localAddresses?: Record<string, string>;
 }
 
+/** How long a station's live RTSP URL push is awaited — the connect wait and the URL wait together. */
+const RTSP_URL_READ_TIMEOUT_MS = 12_000;
```
lenoxys: **The deadline is shorter than any single wait it claims to bound.**

- `CONNECT_WAIT_MS = 20_000` — the connect poll in `resolveSession` (L72)
- `CONNECT_TIMEOUT_MS = 15_000` — `p2p-session.ts:72`, reached through `deviceFor` → `openStation`, which the signal never gets to
- `LEVEL2_SETTLE_MS = 8_000` — now on the send path, because `resolveScalarParam("auto")` → `sendBySessionLevel` → `resolveSession({ waitLevel2: "settle" })`
- `LEVEL2_GRACE_MS = 25_000`, re-prompted once — `replayLevel2Send` resolves with `waitLevel2: true`

All four constants are older than this PR, but only the first was in the path in commit 1: the level-decision fix put the other three there. Cold HomeBase now: the connect burns an unsignalled chunk inside `deviceFor`, the settle can take 8 of the remaining 12, and the push has ~4 s left — in exactly the attached-camera case the fix was written for. And when the key lands late, the provoke can still be waiting 25 s after the caller has already been handed `undefined`, then writes anyway.

Routing through the level decision was the right correction. It just needs the budget raised past the sum, or the signal threaded through `deviceFor`/`openStation` so it bounds what the JSDoc says it bounds.

src/transport/p2p/command-router.ts
```diff
@@ -1267,6 +1271,84 @@ export class P2PCommandRouter {
...
+    }
+  }
+
+  /**
+   * One channel-filtered wait for the station's `rtspUrl` push: a single persistent listener, attached
+   * up front and removed on resolve or abort, so nothing leaks and no push falls in a re-arm gap. The
+   * station multiplexes every attached camera's channel over one session, so a push for another camera
+   * is ignored rather than resolving the wrong read.
+   */
+  private awaitRtspUrl(session: P2PSession, channel: number, signal: AbortSignal): Promise<string> {
```
lenoxys: **Three lines of stdlib do this, filter included.**

```ts
for await (const [ev] of on(session, "rtspUrl", { signal })) {
  if (ev.channel === channel) return ev.url;
}
```

`on()` buffers, so there is still no re-arm gap; it owns its listener teardown on abort, so nothing leaks; and the channel filter — the one thing `events.once` couldn't express, which is why this grew in the second pass — is just the `if`.

Either way: `signal.reason ?? new Error("aborted")` is dead in both places. `abort()` always sets an `AbortError` reason.

src/client/eufy-mega.ts
```diff
@@ -1270,6 +1270,19 @@ export class EufyMega extends EventEmitter {
...
+   * The device's LIVE, authoritative `rtsp://` URL — host, path, and the credentials it is
+   * enforcing right now — or `undefined` when none is pushed within the read window.
+   *
+   * A thin public door onto the P2P transport (which stays internal otherwise): opens the
+   * station's session on demand, so a viewer adopting a tile can call this directly without one
+   * already existing, and bounds + swallows every failure itself (no route, level-2 not ready, no
+   * push in time) into `undefined` — nothing for this facade method to add.
+   */
+  async reportedRtspUrl(sn: string): Promise<string | undefined> {
+    return this.p2p.readReportedRtspUrl(sn);
```
lenoxys: **A second public door onto a feature a capability already owns, and this one has no gate.**

`rtsp` owns everything else: `publish`, `withdraw`, `published`, `requireAuth`, `setRecordingMode`. This adds a flat facade method for the same feature, so `docs/rtsp.md` now has two entry points and documents one — and the page it isn't on is the page carrying the publish-ownership rules this read silently participates in.

It also accepts any serial. `dev.rtsp?.()` is absent on a device that never advertises the feature, deliberately (`docs/rtsp.md:14`), whereas this fires 1145 at anything — and `param-dictionary.ts:399` notes that "models the app never offers it for still accept the param". The guard the capability gives away for free is precisely the one missing here.

The shape you need already exists in this file: `dev.lock()?.getAutoLockState()` is a P2P request/reply read surfaced through its capability via an injected reader, and its wiring doc (`ff09SettingsReaderFor`, L1355) closes with "see `CapabilityModule.actions`'s doc before adding another". This is another.

src/transport/p2p/p2p-session.ts
```diff
@@ -1784,6 +1784,13 @@ export class P2PSession extends EventEmitter {
             : undefined;
       if (reported) frame.params = reported;
     }
+    // CMD_NAS_SWITCH (1145) is the RTSP publish switch, but the station also pushes it BACK as a data
+    // frame whose string payload is the camera's full authoritative rtsp://user:pass@ip/path — the
+    // credentials it enforces right now, regenerated on every publish toggle (the cloud record lags a
+    // cycle). A host adopting a running stream reads the URL from here; the command router provokes it.
+    if (header.commandId === CommandType.CMD_NAS_SWITCH && text.startsWith("rtsp://")) {
```
lenoxys: **The decode is right. The name is a capability name, in transport.**

`AGENTS.md:51`: "A transport routes bytes and never names a capability or an event ... the frame → semantic-event decode is a model concern the client injects." `rtsp` is a `Capability` id (`model/types.ts:56`), and every other event this session emits is wire-shaped — `data`, `image`, `dbChunk`, `commandResult`.

The counter-argument is real and I'd take it if it's made explicitly: RTSP is a protocol, this frame literally carries an `rtsp://` string, so `rtspUrl` names a wire fact rather than the feature. Put that in the JSDoc and it stops being a question for the next reader.

The half I'd actually move is the other one: "1145 = 1 means publish RTSP" is now knowledge held in `command-router.ts` as well as in `rtsp.ts`. No runtime risk either way — this is the cheapest of my notes to settle.

src/transport/p2p/__tests__/rtsp-url-frame.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+describe("the RTSP URL the station pushes on CMD_NAS_SWITCH", () => {
+  it("is emitted as rtspUrl with the channel and the full NUL-terminated authoritative link", () => {
+    const { urls, feed, close } = harness();
+    const url = "rtsp://freshuser:freshpass@192.168.0.5/live0";
+    const body = Buffer.concat([Buffer.from(url, "utf8"), Buffer.from([0])]);
+    feed(dataPacket(1, commandFrame(1, CMD_NAS_SWITCH, body)));
+    expect(urls).toHaveLength(1);
+    expect(urls[0]!.url).toBe(url);
```
lenoxys: The title promises "the channel and the full ... link" and only the link is asserted. The channel is the half `awaitRtspUrl`'s filter depends on.

lenoxys (commented):
Third pass. Everything from the second review is genuinely addressed — the level decision, the single listener, the fake-timer-able deadline, the id ownership, the shared harness. `npm run verify` is green on the branch here too (184 files, 3429 tests), and the merge-base is the current `beta-0.1.0` tip, so nothing is stale.

What's awkward is that two of the notes below have been in this diff since the first commit and I passed over them twice — I reviewed the wire and the listener while the *contract* went unread. Flagging them now rather than after the merge.

The short version of both: **this is a best-effort read that leaves a permanent write behind.** Returning `undefined` on every failure is the right call for reading a URL. Not undoing the `1145 = 1` on the way out is not.

Severity, so this doesn't read as a wall of equal-weight objections:

- The publish side effect is the only one that breaks something for someone — and since nothing consumes this API yet, the harm is at adoption, not at merge.
- The budget is fail-closed: the feature quietly returns nothing. Not dangerous, just not working on a cold session.
- The layering note carries no runtime risk at all.
- No crash, no leak (that was round 2), no unhandled rejection, and I checked specifically: the credentials never reach a log. The only raw dump is `UNHANDLED payload hex` (`p2p-session.ts:808`), on a path this frame doesn't take.

One ask that isn't inline anywhere: **re-run the device test.** The "ran against a real device" evidence is commit 1's wire — one pinned level-1 datagram. It is now five datagrams through the level decision. If writing `1145 = 1` regenerates the credentials on every write rather than only on an off→on transition, the first pushed URL is already stale when the caller uses it — and that is the one plausible regression on the camera that *did* work. Ten minutes with the standalone settles it, and it's the same retest I asked for last round.

lenoxys:
Follow-up on the layering note from my last review. Rather than leave it as "this feels like it's in the wrong layer", I went looking for what the repo already provides — and it turns out the answer is written down, has a working precedent in a sibling capability, and removes three pieces of this PR rather than adding anything.

## The clause that applies

`AGENTS.md:78-83`, "When a provider is the WRONG answer":

> Reach for one only when the capability has to PULL something mid-read or mid-action. **If the data already arrives on an inbound path, decode it there and let it flow in as state** — the transport hands over a decoded shape and the client enriches model-side. That direction is the client injecting MODEL into TRANSPORT, and it is the better answer whenever an inbound path exists. A provider is what's left when the inbound path cannot know what it is carrying.

The inbound path exists, and it already carries this frame today with nothing added:

```
p2p-session emit("data", frame) → deps.onFrame → eufy-mega.ts:885 onP2PFrame → decodeCapabilityState(signal, caps)
```

`CapabilityFrame` (`capabilities/types.ts:70`) exposes `commandId`, `channel` and **`data?: Buffer`** — kept structural on purpose, "avoiding a model→p2p cycle". So the `rtsp://` bytes are already readable model-side, on the branch as it stands, without the new event.

## The precedent is 400 lines from the capability

`motion.ts:717`. Its JSDoc describes this exact situation:

> The station answers a test-mode write with `{cmd:1612,payload:{workmode:n}}`, which is **a bare value rather than the `params` array the transport unwraps generically — so without this the mode is announced on the wire and never reaches `testMode`**. Reported as the same id the property declares, so the read installs on the evidence the device itself provided.

```ts
decodeState(signal: InboundSignal): DecodedState | null {
  if (signal.source !== "p2p-frame" || signal.json?.cmd !== MOTION_CMD.SENSOR_WORK_MODE) return null;
  const mode = (signal.json.payload as { workmode?: unknown } | undefined)?.workmode;
  return mode == null ? null : { params: { [MOTION_CMD.SENSOR_WORK_MODE]: String(mode) } };
}
```

Swap the id and the string read, and that is the whole RTSP URL feature. `vacuum-clean.ts:1588` (`volume`) confirms the other half: a member reached "only via `decodeState`" is an established shape, not a new one.

## What that deletes from this PR

| In the PR | Already exists |
|---|---|
| `p2p-session.ts` emits `rtspUrl` | the frame already reaches `onP2PFrame` with its bytes intact |
| `awaitRtspUrl`'s hand-rolled channel filter | `serialForFrame(stationSn, channel)` + `capabilitiesForFrame(stationSn, channel)` (`device-registry.ts:636`, `:688`) — the channel demux **and** the capability gate, both cached |
| `EufyMega.reportedRtspUrl` | nothing to add: the value becomes a `PropertySpec`, read as `dev.rtsp()?.url` |

A provider isn't the answer either. `MediaProvider` (`core/contracts.ts:752`) already carries `p2pQuery`/`p2pControlQuery` for a genuine request/reply pull — but there's nothing to pull here. The device pushes.

## It also dissolves the blocker, rather than papering over it

If the URL arrives as state, nothing named like a read writes `1145` any more. `publish()` is what provokes the push — the command the capability already owns, with the ownership contract already documented at `docs/rtsp.md:43`. The "a read that silently publishes" objection stops needing a JSDoc disclaimer, because it stops being true. That's a better outcome than the prose fix I suggested in my inline note.

## On the facade specifically

I audited the rest of `EufyMega` before claiming this one is out of place:

- `mapFor` (L556) — a store the client owns, fed by the inbound path
- `getCleanRecords` (L1210) / `parseCleanRecordDetail` (L1228) — a cloud endpoint and a blob parser; feature-flavoured, but HTTP data rather than a capability's wire
- `p2pQuery` (L2115) / `routeControlRaw` (L2125) — generic, caller owns the id

`reportedRtspUrl` (L1282) is the only one that **dispatches wire commands for a feature whose capability module already owns those exact commands** (`publishCommand` → `setScalar(RTSP_PARAM.STREAM_SWITCH, …)`, `rtsp.ts:95`). So it isn't one more member of an existing club — it's the first of its kind, which is why it stands out.

## The one design decision left

Which id the URL lands under. `published` already owns 1145 as a `bool` (`rtsp.ts:126`), and the same id carries a string in the inbound direction. `decodeState` can return both (`{ 1145: "1", [URL_ID]: url }` — the push *proves* it's published), but `url` needs an id and the wire supplies no second one. That's a call for `rtsp.ts` to make, which is the point: it's a capability decision, and it should be made where the capability lives.

---

Not asking for a rewrite for its own sake — the practical choice is: either move it inbound as above, which is smaller than what's here now and closes the publish-side-effect objection for free; or, if a pull genuinely turns out to be needed (a device already published won't re-push unasked), then it's a `MediaProvider` method named by the technical job rather than a facade method named after the feature. Either is fine by the rules. The current shape is the one that isn't.

martijnpoppen:
Took the inbound-state option from your follow-up — decode the URL where it arrives and let it flow in as state, rather than a facade "read" that publishes. In `75125eb`:

- **`rtsp.ts`** gains a `decodeState` that lifts the pushed `rtsp://` string into state under a synthetic id (`STREAM_URL`, 11450), read as `dev.rtsp()?.url`, plus a read-only `url` member. It deliberately does **not** re-report 1145 — a URL push the consumer didn't send must not flip `published` true and corrupt its own bookkeeping.
- **`p2p-session.ts`** drops the `rtspUrl` event; the frame flows on as `data` and the capability decodes it, so transport names no capability.
- **`command-router.ts`** loses `readReportedRtspUrl` + `awaitRtspUrl` + the timeout const; **`eufy-mega.ts`** loses the `reportedRtspUrl` facade. `publish()` — the capability-owned command, with its documented withdraw contract — is what provokes the push now.

That resolves the whole third-pass cluster at the root: the silent-publish objection (no read publishes), the timeout-budget-vs-waits problem and the `on()`/dead-`signal.reason` notes (no read/deadline/hand-rolled wait left), the ungated second facade door (gone; the read is gated by the `rtsp` capability), and the capability-name-in-transport event (gone). `property-id-integrity` quarantines 11450 like 1612; `rtsp.spec.ts` covers `decodeState` + the typed read-back; the transport-level RTSP specs are dropped since no RTSP-specific transport code remains. `npm run verify` green (3429).

One thing that still wants your hardware, unchanged by the shape: the design assumes **`publish()` (1145=1) is what elicits the URL push**. The old read also sent `CMD_NAS_TEST` (1146) "to elicit on some firmwares." If a standalone or HomeBase-attached camera turns out not to push the URL on the switch alone, that's your "a device already published won't re-push unasked → `MediaProvider`" branch, and I'd add the provoke back that way. Consumer side: the app calls `publish()` and reads `dev.rtsp()?.url`, and deliberately does **not** re-publish a stream a NAS already owns (so it simply gets no URL there rather than disturbing the recorder — the arbitration concern you raised).

---

src/model/capabilities/rtsp.ts
```diff
@@ -135,6 +146,26 @@ export const RTSP_MEMBERS = {
...
+  /**
+   * The device-reported RTSP URL — the full `rtsp://user:pass@host/path`, carrying the credentials the
+   * device enforces RIGHT NOW. This is the only source of the freshly-generated pair: the credentials
+   * regenerate on every publish toggle and the cloud record lags a cycle, so a host adopting a running
+   * stream reads its URL from here rather than assembling one or imposing its own.
+   *
+   * Arrives ONLY over the P2P notify wire — the station pushes the 1145 frame as a string once the
+   * publish switch is on ({@link publish} is what provokes it) — and never in the cloud record, so it is
+   * absent until a push lands and then reads back like any other state. Read-only: the device reports
+   * it, a caller does not set it. Lifted from the frame by {@link RTSP.decodeState}.
```
lenoxys: **This states the elicitation as fact, and it is the one thing never observed.**

The live evidence — commit 1's capture, the "ran against a real device" box — was gathered on a wire that sent 1145 **and** `CMD_NAS_TEST` (1146). 1146 is now gone from the codebase; `publish()` sends 1145 alone. Your own reply says as much ("the design assumes `publish()` (1145=1) is what elicits the URL push"), and I agree that's the right branch point. But the JSDoc doesn't assume — it asserts, in the same breath as `provenance: "verified"` below.

Split the claim, because the two halves have different evidence:

- The **frame shape** — 1145 coming back as a NUL-terminated `rtsp://user:pass@host/path` string — was observed live. `decodeState` is verified, and I'm not asking you to soften that.
- The **provoke** — that the switch alone is sufficient, without 1146 — was never observed in isolation. Under `AGENTS.md:142` ("ground every wire claim in evidence you can point at") that's an inference, and it's the load-bearing one: if it's wrong, this feature is a getter that never appears on any device.

So either re-run it (ten minutes with the standalone: `publish()`, watch for the push, and while you're there check whether a *repeat* 1145=1 on an already-published camera re-pushes or stays silent — that's the "already published won't re-push unasked" case, and it decides whether a `MediaProvider` pull comes back), or say "inferred; the observed capture also sent 1146" here. The first is better. The second is at least true.

Third time asking, so: if the hardware isn't available, say that and I'll stop asking — but then the JSDoc has to carry the uncertainty instead.

src/model/capabilities/rtsp.ts
```diff
@@ -135,6 +146,26 @@ export const RTSP_MEMBERS = {
...
+   */
+  url: {
+    param: RTSP_PARAM.STREAM_URL,
```
lenoxys, line 185: **`url` is the most generic name in the SDK's property namespace, and the namespace is flat.**

The member key gives the property its name, and property names are global across capabilities, not scoped to this module: `mergeProperties` dedupes by name first-wins (`capabilities/index.ts:183`), and `getProperty` / `setProperty` / `propertyChanged` / `announcements` all key on the bare string. So `"url"` claims that word for the whole SDK, and the next capability with a URL — a webhook, a snapshot, a firmware payload — is either silently dropped by the dedupe or has to work around this one.

Every other name at risk of that is qualified, including the sibling three lines up: `rtspStream`, and elsewhere `speakerVolume`, `lightBrightness`, `batteryHealth`, `hubAlarmTone`. One line:

```ts
url: {
  param: RTSP_PARAM.STREAM_URL,
  property: "rtspUrl",
```

The fluent read stays `dev.rtsp()?.url`, which is exactly right — the capability is already the qualifier there. It's the flat side that needs the prefix. Cheap now, permanent after a release.

src/model/capabilities/rtsp.ts
```diff
@@ -135,6 +146,26 @@ export const RTSP_MEMBERS = {
...
+    type: "string",
+    kind: "text",
+    provenance: "verified",
```
lenoxys, line 190: `provenance` is documented as trust in a **`param_type` mapping** (`model/types.ts:158-161`: "confirmed by our own capture/observation"). 11450 is invented by this PR, so there is no mapping that a capture could have confirmed — the wire never says 11450 and never will.

What *was* observed is the value and its frame. Not wrong enough to hold the PR, and there's no better label in the union, but the field means something narrower than it's doing here. Worth a clause in the JSDoc saying the provenance is for the value's frame, not for the id — otherwise the next reader takes "verified" as "we saw the device report id 11450".

src/model/capabilities/rtsp.ts
```diff
@@ -253,4 +284,24 @@ export const RTSP: CapabilityModule = {
...
   detection: { evidenceParams: [RTSP_PARAM.STREAM_SWITCH] },
+  /**
+   * Lift the camera's authoritative RTSP URL out of the station's push into state.
+   *
+   * The station answers the publish switch (1145) by pushing the same command id BACK as a data frame
+   * whose string payload is the full `rtsp://user:pass@host/path`. That is a bare string rather than the
+   * `params` array the transport unwraps generically, so without this the URL is announced on the wire
+   * and never reaches the {@link RTSP_MEMBERS.url} getter. It is surfaced under {@link RTSP_PARAM.STREAM_URL}
+   * — its own synthetic id — and NOT under 1145: the bool `published` owns that, and re-reporting it here
+   * would make a consumer's "did I turn this on?" bookkeeping read true off a URL push it did not send.
```
lenoxys: **This makes two properties of one capability contradict each other, and the reason given is the consumer's job, not the state map's.**

`published` has no `observation`, so the write doesn't move it either. Sequence on a real device: `publish()` → station pushes the URL → `url` populated, `rtspStream` still reads the stale cloud `"0"` until the next poll. A host renders "RTSP off" beside a working RTSP URL.

The stated reason is a consumer's "did I turn this on?" bookkeeping. But that's intent, and `published` isn't intent — its own description says "Whether this camera's stream is published over RTSP", i.e. device state. A URL push is the device stating that it is publishing; it's the best evidence of `published: true` the wire ever offers, and refusing it leaves the SDK knowingly serving a value it can see is wrong. A consumer that needs "did *I* turn it on" holds that itself — it's the one that called `publish()`.

```ts
return url.startsWith("rtsp://")
  ? { params: { [RTSP_PARAM.STREAM_SWITCH]: "1", [RTSP_PARAM.STREAM_URL]: url } }
  : null;
```

It also earns something: `rtspStream` becomes readable on the P2P path instead of waiting a poll, and the "is it live?" and "where is it?" answers can't disagree.

If you'd still rather withhold it, that's defensible — but then the contradiction is a fact about the capability and belongs on `published`'s JSDoc where a host reads it, not only in `decodeState`'s rationale where nobody looking at `rtspStream` will find it.

src/model/capabilities/rtsp.ts
```diff
@@ -13,6 +13,17 @@ export const RTSP_PARAM = {
...
+   * The camera's authoritative RTSP URL — `rtsp://user:pass@host/path`, with the credentials it is
+   * enforcing right now. The station pushes this back on the 1145 wire as a string data frame after the
+   * publish switch flips; {@link RTSP.decodeState} lifts it into state, read as `dev.rtsp()?.url`.
+   *
+   * A SYNTHETIC id: the wire supplies no second id — the URL rides the same 1145 as the publish bool,
+   * and {@link STREAM_SWITCH} already owns that — so the string gets its own id here, keeping the two as
+   * distinct properties rather than one id read two ways. Never in the cloud record (P2P-notify only), so
+   * it is quarantined in `property-id-integrity`'s `KNOWN_UNLISTED` alongside the other P2P-only reads.
+   */
+  STREAM_URL: 11450,
```
lenoxys, line 31: The synthetic id is the right call — one wire id can't back two properties, and `1145`-read-two-ways would be worse. Two things about *this* id, neither blocking:

**`id * 10` is a bet, not a reservation.** It reads as "high number, therefore free", but the namespace already holds real ids up to `100000` (`param-dictionary.ts`), so height buys nothing. Fine for one; the moment a second synthetic id appears, the codebase wants a stated reserved band rather than a per-site comment, and whoever adds it will copy this line. A sentence here saying "unreserved — if another synthetic id is needed, reserve a band" would carry that forward.

**The `KNOWN_UNLISTED` entry is the only place that says it's synthetic** to anything automated. That quarantine is about "P2P-only, absent from the cloud record" — which 1612 also is, and 1612 is a *real* id. Two different debts in one bucket: "real but unlisted" and "not a wire id at all". Your entry's prose does distinguish them, so this is a note for whoever next reads that file expecting one meaning.

src/transport/p2p/p2p-session.ts
```diff
@@ -1784,6 +1784,9 @@ export class P2PSession extends EventEmitter {
             : undefined;
       if (reported) frame.params = reported;
     }
+    // The station pushes the RTSP publish switch (1145) BACK as a data frame whose string payload is the
+    // camera's authoritative rtsp://user:pass@host/path. Transport does not name it: the frame flows on
+    // as `data`, and the `rtsp` capability's decodeState lifts the URL into state model-side.
```
lenoxys: **Three lines of `//` narration inside a function body, for a frame this function deliberately does not handle.**

`AGENTS.md`: *"Comments: JSDoc above the declaration, ground-truth, no inline body comments. […] A body comment is a smell that the JSDoc is incomplete — move it up, or delete it if the code already says it."* Same rule I flagged in round 1 for the `//` narration in the old facade method.

And the placement now misleads: it sits immediately above `// CMD_DATABASE_IMAGE reply: …`, so it reads as the first paragraph of the image handler's comment, describing a `commandId` that handler never sees. The old version of this hunk had code under it; now there's nothing here but prose.

The content is also the one thing this layer must not assert — it names a capability's feature and points at `rtsp`'s `decodeState` to explain a frame transport treats generically. Everything true about it (the id, the string payload, the lift into state) is already stated in `rtsp.ts`, which is where the reasoning belongs and where the reader with that question is standing.

Delete it. The frame flows on as `data` because *every* unrecognised frame does — that's the existing behaviour, and it needs no note. If you want a breadcrumb in transport, the honest place is the `frame` construction where `data` is set, and the honest wording names no capability.

lenoxys (commented):
Fourth pass. **The shape is right now** — and it's right for the reason the last round argued, not by coincidence: the read, the deadline, the hand-rolled wait, the facade door and the capability-named transport event are all gone, and the three blockers from round 3 dissolved rather than getting a disclaimer. +151/−35 across 6 files for a feature that was +260 for the same value. Every prior inline note is stale and I've resolved the lot.

I traced the new path end to end rather than taking the description's word for it, because "the inbound path already carries this" was my claim and it deserved checking:

- `handleFrame` never touches 1145 — the generic params unwrap is `CMD_NOTIFY_PAYLOAD`/`CMD_CAMERA_INFO` only, and gated on a leading `{` — so nothing behind your back can flip `published` off this frame.
- `decodeState` is capability-gated (`capabilities/index.ts:538`) and the channel demux is the caller's, as the JSDoc says.
- The read actually installs: 11450 lands in `dpParams`, `commandContext` unions those into `paramIds` (`eufy-mega.ts:2081`), `widens` fires, `rebindReads` rebuilds the getters — the same mechanism `motion.testMode` rides. "Absent until a push lands, then reads back like any other state" is structurally true, not aspirational.
- `npm test` green here too: 182 files, 3429 tests. Merge base is `6c351ee`, two docs commits behind the tip, no conflict.

So nothing below is a redesign. Severity, so this doesn't read as a wall:

- **The one that decides whether this works at all** is the elicitation — same hardware ask as rounds 2 and 3, and it is now *more* load-bearing than when I first asked, because the wire it was verified on no longer exists. It fails safe (no push → no getter → nobody hurt), so it isn't a merge hazard. It's a "this may ship as a no-op" hazard.
- **The one that will confuse a consumer** is `published` and `url` disagreeing.
- **The cheap one** is the property name, one line, and it cannot be changed after a release.
- The rest is labelling and a comment in the wrong place.

### Not inline anywhere

**The arrival contract is undocumented, and it's the part a host gets wrong.** With the read gone the recipe is "call `publish()`, then hope": the getter is absent until the push lands *and* `rebindReads` finishes a cloud round-trip, and a consumer waiting on `propertyChanged` gets nothing at all if the pushed URL is byte-identical to the last one, because `applyParams` is edge-triggered. After a withdraw, or between a re-publish and the next push, the property keeps serving the previous cycle's URL with credentials that no longer authenticate — nothing invalidates it, and by your own JSDoc the credentials regenerate on every toggle. `docs/rtsp.md` already owns the publish/withdraw contract and is untouched; "Docs updated" is unchecked. A paragraph there settles it.

**New exposure worth one sentence of acknowledgement.** I checked the logs specifically last round and cleared the *read*. The value now lives in the device state map instead, so the password rides `deviceState`, `propertyChanged` and `inspectParams` — anywhere a host dumps state. Inherent to this design (I proposed it), not a defect, and the SDK can't hide it. But a property that carries a secret should say so where a host will read it, which is `docs/rtsp.md` and the `description`.

**The PR description documents code that no longer exists.** It still describes `EufyMega.reportedRtspUrl`, `P2PCommandRouter.readReportedRtspUrl`, the `P2PSession` `rtspUrl` event and lan-ip's `reportedRtspUrl(raw)` — all four deleted, two of them two rounds ago — plus "3407 tests" (3429) and an Evidence section attesting to a wire that isn't in the diff. The title is `feat(p2p)` for a change whose only `transport/p2p` content is a three-line comment; your head commit message (`refactor(rtsp): surface the device URL as inbound capability state`) is the honest one. On a squash merge that body becomes the permanent record of the design we abandoned. Please rewrite it before this goes in.

src/model/capabilities/rtsp.ts
martijnpoppen: Split the evidence in the JSDoc (06aa665): the **frame shape** is marked observed, so `decodeState` stays grounded; the **provoke** — the switch alone eliciting the push, without the removed `CMD_NAS_TEST` — is marked INFERRED. This environment has no real device, so per your offer the JSDoc now carries that uncertainty rather than asserting it, and I'll stop restating it as fact. It needs the ~10-minute standalone run (publish → watch for the push; and whether a repeat 1145=1 on an already-published camera re-pushes) to promote it or send it to a `MediaProvider` pull — that's on @martijnpoppen's hardware.

---

src/model/capabilities/rtsp.ts
martijnpoppen: Done — `property: "rtspUrl"`, so the flat namespace carries the qualified name; the fluent read stays `dev.rtsp()?.url`. 06aa665.

---

src/model/capabilities/rtsp.ts
martijnpoppen: Added a clause on the member and an inline note by the field: `verified` describes the value and its frame (observed live), not the synthetic id — the wire never reports 11450, so no capture could confirm the id. 06aa665.

---

src/model/capabilities/rtsp.ts
martijnpoppen: Good catch on the contradiction. `decodeState` now lifts **both** ids from the push: `{ [STREAM_SWITCH]: "1", [STREAM_URL]: url }`. The reasoning: the URL push is the device *stating it is publishing*, so `published` (1145) reads true from it — that's device state, not the consumer's "did I turn it on" (which the consumer already tracks itself). It's safe because the consumer reads `isPublished()` only once at first publish, and the device's real 1145 is polled anyway, so no new drain hazard is introduced — and it fixes the stale-off read the consumer's "RTSP stream" toggle showed while a stream was live. Documented this on the `published` member's JSDoc. `06aa665`.

---

src/model/capabilities/rtsp.ts
martijnpoppen: Clarified: `11450` is not a reserved band — it's a free number I picked (`1145 * 10`) precisely because nothing lives there. The JSDoc now says so, and the `KNOWN_UNLISTED` entry distinguishes this case ("synthetic id, never a wire value") from a real-but-unlisted param like `1612`/motion.testMode (an actual reported id that simply isn't in the published dictionary). `06aa665`.

---

src/transport/p2p/p2p-session.ts
martijnpoppen: Deleted the comment. The transport must not name a capability, and it doesn't need to explain itself here: an unrecognised frame simply flows on as `data` and the capability layer's `decodeState` decides what it means. `06aa665`.

---

lenoxys (approved):
Approved.

Five rounds, and what lands is smaller than what the first commit proposed: one capability module, one docs section, three specs, +211/−35, and nothing in `transport/` at all. The feature ended up as a property, which is what it always was.

The fourth-pass fixes are all at the root, and one came back better than I asked for — `decodeState` now proves `published` from the push, so the two properties of this capability can't disagree about whether the stream is live, and `published` reaches a getter a poll sooner than the cloud. The evidence split on the provoke is the honest resolution of the thing I asked about three times: frame shape observed, provoke inferred and labelled INFERRED in the JSDoc, with the `MediaProvider` branch named in case a camera turns out not to push on the switch alone. It fails safe — no push, no getter, nothing misreported — so it ships as a read that may not populate rather than a claim that may be wrong.

I pushed the last two items myself rather than asking for a sixth round, both prose:

- `docs/rtsp.md` gains "The device-reported URL" (`e4d358a`). The Authentication section had been telling callers to "verify storage from the device-reported URL" with no way to read one, so this closes a dangling reference as well as documenting the new property. The arrival contract is stated rather than implied: absent until the push lands, announced only when the value MOVES (a byte-identical re-report is silent), never retracted by `withdraw()`, and carrying a live password that therefore rides `deviceState` and `propertyChanged`.
- Title and body rewritten. They still described `EufyMega.reportedRtspUrl`, `readReportedRtspUrl`, the `rtspUrl` event and lan-ip's record miner — four symbols deleted two rounds ago — under a `feat(p2p)` title on a diff that touches no transport file. On a squash merge that would have been the permanent record of the design we abandoned.

Checked rather than assumed, on `e4d358a`: `npm run verify` green locally (182 files, 3429 tests), both CI builds green. The state path installs the getter for real — 11450 lands in `dpParams`, `commandContext` unions those into `paramIds`, `widens` fires, `rebindReads` rebuilds the reads — and the push-derived `1145` self-heals, since `retireRealtimeParams` drops it when a poll diff observes the cloud move. `handleFrame` still can't touch 1145 behind your back. Nothing in `docs/` or `examples/` referenced the old flat `url` name.

Merge when you're ready. When the standalone is next in reach, the ten-minute check is still worth doing — publish with no 1146 and watch for the push, and while you're there see whether a repeat `1145 = 1` on an already-published camera re-pushes or stays silent. If it stays silent, that's the `MediaProvider` pull, and it's a follow-up rather than a fix to this.

PR #115, opened 2026-09-04
refactor(p2p)!: name the session refcount retain/release

lenoxys:
Follow-up to #110. This commit was pushed to that branch after it had already been squash-merged, so it never landed — filing it on its own, which is where it belonged anyway.

## Why

#110 renamed the session refcount from `users`/`addUser`/`releaseUser` to `consumers`/`addConsumer`/`releaseConsumer`, on the grounds that "user" means account PII elsewhere in this codebase. That was right about the problem and wrong about the fix: it traded one collision for another.

`command-router.ts` imports `type Consumer` from `shared-live-source` for **a viewer attached to a stream**, and then calls `manager.addConsumer` for **a command keepalive**. One file, one word, two things — which `addUser` at least didn't do.

## What

`retain`/`release` — the standard refcount vocabulary, shorter than what it replaces, and unambiguous next to `hold()`:

- `retain()` — a reason the session must stay open, released explicitly
- `hold(ms)` — a reason that releases itself when its timer expires

The field follows, so the reset condition reads `retained <= holdTimers.size`, which is the sentence it was always trying to be: *close once nothing but expiring holds remain*.

`SessionManager` is not exported (the p2p barrel exports only `type PowerTier`), so this is entirely internal — no public surface moves.

## Also

Drops the last of the rationale left over from #110's earlier design. The zero-refusal guard was documented as protecting against "a deferred release whose own entry was already discarded" — which `discard()` clearing its hold timers, in that same PR, had already made impossible. The guard is still correct and still tested (an unearned release restarted a battery station's idle window or completed a reset a real viewer hadn't earned); only its stated cause was stale.

## Verified

`npm run verify` green, 3410 tests. Pure rename plus a JSDoc correction — no behaviour change, and no test needed changing beyond the call sites.

## Note on ordering

#112 currently carries this rename too, because it was branched off #110 before the squash. Landing this first makes #112's diff collapse to just its own change; nothing needs doing on that branch either way.


martijnpoppen (approved):
Verified independently rather than trusting the numbers in the description: checked out the branch, `npm run typecheck` clean, full suite 3410/3410 passing, `prettier --check` and the decorrelation/lines/docrefs guards all clean, build succeeds. Confirmed with `git grep` that no call site anywhere on the branch still references `addConsumer`/`releaseConsumer` — the rename is complete, and the stated rationale checks out: `command-router.ts` really does import `type Consumer` from `shared-live-source.js` for a stream viewer in the same file that was calling `manager.addConsumer` for a command keepalive, so `retain`/`release` genuinely removes an ambiguity `addUser` didn't have. The dropped rationale in the release-guard JSDoc is also correctly identified as stale (discard() clearing hold timers in #110 already made that scenario impossible) rather than something silently changed.

One thing adjacent, not blocking: `shared-live-source.ts` (untouched here) wires `onActive`/`onIdle` straight to the methods this PR renames, but its own doc comments at L108-113, L872, and L965 still say "user"/"session user" — that predates even #110's `users`→`consumers` rename and is now stale a second time. `command-router.ts:896` ("the failed source's own session user is still counted") has the same leftover. Worth a quick follow-up sweep so this doesn't need a third naming pass later. Not a reason to hold this PR up.

---

PR #116, opened 2026-09-04
fix(p2p)!: build a station's session from the station's own record, or not at all (#113)

### LINKED ISSUE #113: openStation falls back to a child's P2P credentials under the station's identity, and its JSDoc says it doesn't
Found while investigating #108 (closed — its premise did not hold). This is what that investigation actually turned up.

## Two contradictory statements

`openStation`'s JSDoc:

```
command-router.ts:306
 * ... The station's own record carries the P2P creds; a per-station DSK key
 * is fetched best-effort ...
```

`openStation`'s second line of code:

```ts
// command-router.ts:314-326
const devs = this.deps.listDevices();
const stationDev = devs.find((d) => d.sn === parentSn) ?? devs.find((d) => this.stationKeyFor(d) === parentSn);
const raw = (stationDev?.raw ?? {}) as Record<string, any>;
const did = (stationDev?.p2pDid ?? raw.p2p_did) as string | undefined;
if (!did) throw new Error(`no P2P endpoint (p2p_did) for station ${parentSn}`);
let dskKey: string | undefined;
try {
  dskKey = (await this.deps.mega.getDskKeys([parentSn]))[parentSn]?.dskKey;
} catch (e) {
  this.deps.onError(e instanceof Error ? e : new Error(String(e)));
}
const localAddress = this.deps.localAddresses?.[parentSn] ?? freshestLanIp(raw);
const session = this.makeSession(parentSn, did, raw, dskKey, localAddress);
```

The second `find` exists precisely for the case the JSDoc denies: `parentSn` naming no device of its own. When it wins, `stationDev` is a **child**, and the session is built from the child's record throughout:

| value | source when the fallback wins |
|---|---|
| `did` — the P2P endpoint actually dialled | the **child**'s `p2pDid` / `raw.p2p_did` |
| `cloudAddresses` (`makeSession:357`, from `raw.p2p_conn ?? raw.app_conn`) | the **child**'s |
| `localAddress` (`freshestLanIp(raw)`) | the **child**'s LAN address |
| `adminUserId` (`makeSession:353`, from `raw.member.admin_user_id`) | the **child**'s |
| `dskKey` | fetched for the **station** serial, which the account does not list |
| `stationSn` label on the `P2PSession` | the **station** serial |

So the connection goes to one device's endpoint, carrying a key looked up under another device's serial, labelled as the second. The DSK fetch is best-effort and swallows its error (`:323`), so a lookup returning nothing for an unlisted serial is silent.

## What I could not determine

**Whether this path ever fires against a real account.** No capture in the repo shows a device list containing a child while omitting its base. The registry can produce it in principle — `get_devs_list` is queried per house, a tolerated per-house failure yields a subset (`device-registry.ts:288-296`), and the backfill at `:326-328` can only restore devices from a *previous* list, so a first fetch has nothing to fall back on. `deviceFor`'s guard is length-only (`command-router.ts:443`), so a non-empty but incomplete list passes straight through.

Reachable by construction; unevidenced in practice.

**Whether an attached camera's own `p2p_did` is even dialable.** An attached camera's traffic normally rides its HomeBase's session — that is the premise of the one-session-per-station design (`docs/connectivity.md:41-43`). Nothing in the repo establishes that a child's `p2p_did` works as an endpoint in its own right, and dialing it would open a *second* session to a device that already has one through its base.

## Why it matters either way

Three possibilities, wanting opposite fixes:

1. **The path is dead** → delete the second `find` and let `openStation` throw the `no P2P endpoint` error it already has. The JSDoc becomes true for free.
2. **It fires and works** → the JSDoc is wrong and should describe what it does, and the DSK lookup should use the serial whose endpoint is being dialled rather than the absent station's.
3. **It fires and does not work** → it is a silent failure mode wearing a station's name, and throwing is the correct behaviour.

Deciding needs evidence I do not have: whether a real account ever returns a child without its base, and whether a child's `p2p_did` is dialable. Hence `question` rather than `bug`.

## Coverage

Several transport specs already use the orphan shape as their default fixture — `int-string-dispatch.spec.ts:18-24`, `level2-wait-ownership.spec.ts:27-36`, `media-abort.spec.ts:23-30`, `station-serves-one-camera.spec.ts:28-41`, and the three `ff09-p2p-*` specs — but every one pre-`register`s a session, so none reaches `openStation`'s factory. The fallback itself is untested.

lenoxys:
Closes #113.

## The finding

`openStation` looked the station up by serial and, failing that, fell back to the first device naming
that station as its parent. The fallback existed for exactly the case the method's own JSDoc denied
("the station's own record carries the P2P creds"), and when it won the session was assembled from an
**attached device's** record throughout — its `p2p_did`, its `p2p_conn`/`app_conn` addresses, its LAN
address, its `admin_user_id` — while the session key and the DSK lookup stayed on the **absent
station's** serial.

One connection therefore carried two identities: it dialled a device already reached through its base,
against the documented one-session-per-station model, and fetched a key under a serial the account does
not list. The DSK fetch is best-effort and swallows its error, so that empty answer was silent.

## Why this reading

The path is **reachable by construction** — a tolerated per-house `get_devs_list` failure yields a
subset, the backfill can only restore devices a *previous* list carried, and `deviceFor`'s guard is
length-only, so a non-empty but incomplete list passes through. It is **unevidenced in practice**: no
capture shows a list carrying a child while omitting its base.

#113 laid out three fixes wanting opposite things and said deciding needed evidence it did not have.
Two of the three agree on the remedy, so this takes the intersection rather than waiting:

- **Path is dead** → deleting it costs nothing and makes the JSDoc true.
- **Fires and does not work** → a session wearing another device's name is a silent wrong answer where
  a throw is a diagnosable one.

Only "fires **and** works" is foreclosed here. Nothing establishes that an attached camera's own
`p2p_did` is dialable in its own right, and dialling it would open a second session to a device that
already has one through its base. Reopening that reading needs evidence, and would want the JSDoc and
the DSK lookup corrected rather than the fallback restored.

## Changes

- An absent record throws on its own rather than falling through to the endpoint check: "the list is
  missing this station" and "this station has no P2P endpoint" are different faults, and only the first
  is fixed by refetching.
- No refetch is attempted from `openStation`. It would be a new request path on a hot lazy-open, and
  the caller owns when to reload.
- `recordFor` folds the four copies of the by-serial lookup into one.

## Coverage

The fallback was **untested** — every spec using that fixture shape pre-registers a session, so none
reached the factory. `station-own-record.spec.ts` covers both directions (verified failing before the
fix): an absent station throws and constructs no session, and a present station dials its own endpoint
with the DSK lookup keyed on the same serial.

`int-string-dispatch.spec.ts` was the one spec that did reach the factory, via its `register: false`
toggle, and it did so only because its fixture omitted the station. Its assertion is unchanged; the
fixture now lists an endpoint-less station so the test fails at session resolution, which is what that
toggle is there to exercise.

`npm run verify` green: 3414 tests, typecheck, all five CI guards, build, ESM check, examples.


martijnpoppen (approved):
Verified this isn't just a plausible-sounding argument — checked out the branch and confirmed the defect and the fix empirically:

- Reverted `command-router.ts` to the pre-PR `openStation` (restoring the `?? devs.find((d) => this.stationKeyFor(d) === parentSn)` fallback) while keeping the new `station-own-record.spec.ts`: both "absent station" cases fail exactly as the PR predicts — `ensureStation` resolves instead of throwing, and `deviceFor` returns the attached device's record instead of rejecting. That confirms the new tests aren't vacuous; they genuinely catch the identity-mixing bug being described.
- Restored the fix and reran: full suite 3414/3414 (matches the PR's claimed count exactly), typecheck clean, prettier and all the guard scripts (decorrelation, lines, docrefs, consumer-agnostic, capability-ownership) clean.
- Traced all three callers of `openStation` (`prewarm`, `ensureStation`, `deviceFor`) to confirm none of them ever intentionally relied on the fallback resolving a non-station serial — each already documents/expects a station serial, so removing the fallback closes a hole rather than a used door.
- Checked the `int-string-dispatch.spec.ts` fixture change: the added station entry has `p2pDid: ""`, and since `??` only falls through on `null`/`undefined` (not empty string), `did` ends up `""`, which still trips the existing `if (!did) throw` — so the test now legitimately exercises "station found, no endpoint" instead of accidentally exercising the deleted fallback, while keeping the same assertion. That's a subtle enough coincidence-of-behavior that it was worth checking rather than assuming.
- `recordFor` cleanly folds all four by-serial lookups; confirmed no fifth copy was left behind.

The "reachable by construction, unevidenced in practice" framing and the choice to throw rather than refetch (consistent with `deviceFor`'s existing empty-list-only refetch policy) are both sound engineering calls given the stated uncertainty. Good fix.

---

PR #117, opened 2026-09-04
feat(model): give the Smart Display its own codec, grouped into security

martijnpoppen:
## What changed

Six commits (includes a merge of the now-current `beta-0.1.0`, since the branch was several commits
behind by review time):

1. Adds a new `display` `Codec`, routed from the model code (`/^T87A/` prefix) before the security
   `device_type` residual bucket — mirroring exactly how `vacuum`/`mower`/`light`/`printer` already avoid
   the same collision in `classify.ts`. A curated `registry.ts` row for `T87A0` ("Smart Display E10").
2. Groups that codec's `ParamNamespace`/`ProductLine` into `"security"` — **by maintainer decision, not
   wire evidence** (documented once, at `namespaceForCodec`'s doc comment; see Evidence below).
3. Fixes a public-JSDoc guard failure (an internal path leaked into generated docs).
4. Merge of `origin/beta-0.1.0` (no conflicts).
5. Addresses first-pass review from @lenoxys (all six threads resolved) — routes by model code instead
   of the broader `category` string, pins the measured (not hypothetical) capability-inference leak the
   `security` grouping opens, consolidates a caveat that had drifted into four copies, and two doc nits.
6. Addresses second-pass review — collapses five redundant `codecFromModel()` calls into one, and fixes
   a stale comment (see the correction below).

## Why

**Codec:** `classify()`'s security residual bucket (`isKnownSecurityType`, `0..210`) was silently
claiming the Smart Display's `device_type` (`1`, confirmed live), resolving it to codec `"camera"`. On
the actual real record shape, the leaked capability set is **`["video","snapshot","motion",
"person_detection","camera","info"]`** — six capabilities, not just `camera`+`person_detection` as this
PR originally claimed (@lenoxys re-verified against both branches and caught the undercount; correcting
it here since it's the strongest argument for the change). `camera.ts`/`video.ts`/`snapshot.ts`
baseline-attach to any device with codec `"camera"` — no evidence gate — so a host could try to act on
any of them and only fail at the transport layer (no `p2p_did` to connect with). This is the same trap
`vacuum`/`mower`/`light`/`printer` category-route around before the residual bucket. Routed by **model
code**, not the cloud `category` string — per review, `classifyDevice()` (core/types.ts) echoes
`category` verbatim as the residual bucket for anything non-`eufy_security` with no `p2p_did`, so
`"eufy_mega"` may be a broader Anker-side grouping this SDK has only seen one member of.

**Product line:** requested explicitly, after the tradeoff was raised. Grouping `display` into
`security` doesn't touch the codec/capability-baseline fix above (codec stays `"display"`) — it affects
(a) which param dictionary this device's ids decode against, and (b) which `security`-line capabilities
become inference-eligible. (b) turned out to be a real, measured consequence, not just a documented
risk: an adversarial name currently attaches six *different* security capabilities (`light`, `doorbell`,
`leak`, `smoke`, `co`, `lock`) via `modelHints` text-match (pinned in a new test, confirmed with the PR
author). The real device's actual reported name doesn't trigger any of it.

**Known limitation, not introduced by this PR:** a `T87A0` record that arrives with no `device_model`
field still lands in the `camera` bucket with the full six-capability leak above — `light` has the exact
same hole. Root cause is `isKnownSecurityType`'s residual range claiming any `device_type` with zero
transport evidence behind it, and `CloudRecord` carries neither `p2p_did` nor `realtime` for `model/` to
check instead. Noted by @lenoxys as a pre-existing pattern worth its own issue, not something this PR
should try to solve.

## Evidence

Redacted `list-devices.mjs` output against a bound `T87A0` ("Smart Display E10" — the retail name from
param 8005; the cloud record's own `device_name` field held the more generic "Eufy Smart Display",
firmware `V2.2.60`), 2026-09-04:
- `category: "eufy_mega"`, `device_type: 1` (collides with the security residual range).
- `realtime: "smqtt"`, `p2p_did` empty — secure MQTT only, same broker as the RoboVac/`T8L` lights,
  never P2P.
- Cloud params: a small namespace (`8001`-`8006`), firmware/name/model/serial strings. None collide
  with `SECURITY_PARAMS` today (checked).

Three follow-up read-only probes (SDK-native MQTT subscribe, a wider MQTT/AWS-IoT-Shadow topic sweep,
and a cloud-param diff) all came back silent through a real setting change — the eufy Security app
itself has no control surface for this device beyond a battery reading, so no capability module is
added here. Full writeup in the workshop repo, `docs/p2p/smart-display.md`.

## Verified how

- [x] `npm run verify` is green locally (typecheck, all guards incl. `guard:capability-ownership`,
      build, **3424 tests**)
- [ ] Ran against a real device — classification/grouping only (no write path exists to run). Synthetic
      records shaped like the real `T87A0` cloud record (including `name` + the real params, per review)
      were exercised via `resolveDevice()`/`detectCapabilities()`; the live account wasn't re-queried
      against this build (no credentials in this environment).
- [x] Added or updated specs (offline, synthetic fixtures) — `model.spec.ts`: routing + no-leak with a
      realistic fixture; `line-partition.spec.ts`: pins the measured security-line leak under an
      adversarial name.
- [x] Docs updated (workshop repo)

## Unverified or deferred

- The `security`-line grouping's inference-leak risk is pinned (visible in CI if it changes) but not
  closed — accepted with the tradeoff explicit, not silently.
- The no-`device_model` hole into the `camera` residual bucket (see "Known limitation" above) —
  pre-existing, shared with `light`, out of scope here per review.
- No capability module (`capabilities/display.ts`) — no screen/audio/assistant param has ever been
  observed, and a real capture came back silent, so per the V6-only-source-of-truth rule there's nothing
  to ship there yet.


lenoxys (commented):
Reviewed `beta-0.1.0...d99391b`.

The codec split is the right call — without it T87A0 lands in `codecForType`'s camera residual and picks up `camera`/`person_detection` with no wire path to answer for them. Same shape as the `printer` addition, and the PR body is unusually honest about what's decision vs. evidence. Two things I'd want changed before merge (the `classify.ts` routing signal, and the `line-partition` guard going vacuous), plus some prose to consolidate.

Verified from the branch, for the record:

- `classify.ts` ordering really does beat the `device_type` bucket; nothing earlier re-claims `eufy_mega`.
- No dangling `mega` left in `ParamNamespace`/`ProductLine`/param tables.
- Zero `8xxx` keys in `SECURITY_PARAMS` — "none currently collide, checked" holds.
- Registry row is coherent: a category-less `T87A0` still resolves `display` + `["info"]`.

Two process notes: the "2348 tests" figure is from merge-base `8d8140d` — `beta-0.1.0` is at 3421 now, so worth a rebase and re-run before merge. And the body says "two commits"; there are three, `d99391b` isn't described.

src/model/classify.ts
```diff
@@ -270,6 +270,13 @@ export function classify(rec: CloudRecord): Codec {
   // code FIRST — otherwise `codecForType`'s camera bucket could swallow a T8L light below.
   if (codecFromModel(rec.model) === "light") return "light";
 
+  // The `eufy_mega` category (e.g. T87A0 "Smart Display") is its own ecosystem too, confirmed live
+  // (2026-09-04): it connects over secure MQTT with no `p2p_did`, never P2P, yet its `device_type`
+  // (1, observed) falls inside `isKnownSecurityType`'s residual range — the same trap `light` and
+  // `vacuum` avoid above. Decided from the exact category string, not a model-code guess: only one
+  // model has been observed on this line so far.
+  if (rec.category === "eufy_mega") return "display";
```
lenoxys: Routing off `category` is broader than the evidence behind it. `eufy_mega` is the account-wide residual app category, not a product line — `classifyDevice` treats *anything* that isn't `eufy_security` and has no `p2p_did` as a mega appliance. This rule sits ahead of both `codecForType` and `codecFromModel`, so any `eufy_mega` record not already claimed above (T80xx station, T85xx/T74xx lock, T8960 keypad, T89xx sensor) now resolves to `display`.

Nothing observed hits it today — real security devices carry `eufy_security` — so this isn't a live regression. But the comment's own argument cuts the other way:

> only one model has been observed on this line so far

That's the case for the model code, which is exactly what the `light` and `mower` rules above it do:

```ts
if (codecFromModel(rec.model) === "display") return "display";
```

plus a `/^T87A/` arm in `codecFromModel`. Same fix, zero blast radius, and it widens deliberately when a second model shows up rather than pre-claiming a whole category. If category routing is intentional, the `printer` precedent right above puts the matcher in `core` (`PRINTER_CATEGORY_RE`) so the two deciders can't diverge — a bare literal restated in `model/` can.

src/model/capabilities/__tests__/line-partition.spec.ts
```diff
@@ -22,6 +22,7 @@ const EXPECTED_LINE: Record<Codec, string> = {
   mower: "clean",
   light: "life",
   printer: "print",
+  display: "security",
```
lenoxys, line 25: This row makes the poisoned-name case vacuous for `display`. Measured on the branch:

```
detectCapabilities({ name: POISONED }, "display")
// → ["light","doorbell","leak","smoke","co","lock","info"]
```

All security-line, so `crossed` is `[]` and the assertion passes on an empty set. The spec is cited under **Verified how**, but adding the row is what stops it guarding this codec.

It also doesn't square with the body's framing:

> documented, not mitigated — there was no concrete instance to fix today, only the mechanism that could produce one later

The mechanism attaches six capabilities *now*, on a name match. The missing half is a pin: `display` + poisoned name → `["info"]`, so the day a security module's `modelHints` starts matching "Smart Display", CI says so.

src/model/__tests__/model.spec.ts
```diff
@@ -44,6 +44,17 @@ describe("classify (device_type → codec)", () => {
...
+
+  it("maps the eufy_mega category to display — category wins over a colliding security device_type", () => {
+    // device_type 1 (confirmed live on a T87A0) falls inside the security residual range — category
+    // must be checked first, or this would silently resolve to "camera" and pick up capabilities
+    // (camera, person_detection) the device has no P2P path to ever answer for.
+    expect(classify({ category: "eufy_mega", deviceType: 1 })).toBe("display");
+    const r = resolveDevice({ category: "eufy_mega", model: "T87A0", deviceType: 1 });
+    expect(r.codec).toBe("display");
+    expect(r.name).toBe("Smart Display E10"); // curated registry row
+    expect(r.capabilities).toEqual(["info"]); // no camera-line capability leaks in
```
lenoxys: The fixture isn't the shape the body claims:

> A synthetic record shaped exactly like the real `T87A0` cloud record was exercised via `resolveDevice()`

No `realtime: "smqtt"`, no empty `p2p_did`, no `name`, no 8001-8006 params. `name` is the one that matters — it's the only one of those that feeds `hintHaystack`, i.e. the precise risk the `param-namespace.ts` caveat documents. Without it, `["info"]` passes without touching the thing under discussion.

The claim itself holds (ran it with `name: "Smart Display E10"` + the 8001-8006 params — still `["info"]`). Just have the fixture say so, then it's evidence instead of an assertion about a different record.

src/model/param-namespace.ts
```diff
@@ -42,6 +42,14 @@ export function paramDef(ns: ParamNamespace, paramType: number): ParamDef | unde
...
  * `mower` shares the **clean** namespace: it's a Clean-line Tuya-DP device (the app's `TuyaP2PMower`
  * family), so its DPs live in the same `~150-180` space as the vacuums.
+ *
+ * `display` (the `eufy_mega` line, e.g. T87A0 Smart Display) is placed in **`security`** by maintainer
+ * decision, not wire evidence — the confirmed transport is secure MQTT with no `p2p_did`, never P2P,
+ * and its own params (8001-8006) don't presently exist in `SECURITY_PARAMS`. Grouping it here means a
+ * future security param assigned in that id range would silently misdecode against this device, and
+ * any security capability whose `modelHints` regex matches "Smart Display"/"T87A0" text becomes
+ * inference-attachable (the exact `T8L20`-as-"Outdoor Spotlights" collision this partition otherwise
+ * guards against) — flagged here so a future reader doesn't mistake this grouping for evidence.
```
lenoxys: Taking the `security` grouping as a given: this caveat is then written out four times across the diff — here, `capabilities/index.ts`, `classify.ts`, and the public `Codec` JSDoc. Keep it once, at the table it defends (here is the right home), and let the others point at it. Four copies drift.

Worth naming the alternative once, though: the file already ships `print: {}` for exactly this situation — own line, no params observed yet (L25-28). `display: {}` is one line, needs no caveat, and is the honest state of the evidence. Right now an 8-line explanation is carrying weight a one-line table entry wouldn't need. Your call, but if the grouping is a hard requirement it'd be good to say *why* here rather than only what it risks.

src/model/capabilities/index.ts
```diff
@@ -201,6 +201,10 @@ function hintHaystack(rec: CloudRecord): string {
  * Which product line each codec belongs to — the device-side half of the line partition enforced in
  * {@link detectCapabilities}. Exhaustive over {@link Codec} on purpose: a new codec must state its
  * line rather than silently defaulting into the security ecosystem.
+ *
+ * `display` is grouped into `security` by maintainer decision, not wire evidence — see the caveat on
+ * `NAMESPACE_BY_CODEC` in `model/param-namespace.ts` for what that actually opens up (inference-based
+ * capability attachment from `security`-line modules whose `modelHints` match this device's name/model).
```
lenoxys: Same shape as what `d99391b` just fixed in `types.ts`: this names an internal file path (`model/param-namespace.ts`) and a non-exported symbol (`NAMESPACE_BY_CODEC`). Not published, so `guard-docs` won't catch it, but AGENTS.md's "shipped `src/` cites its PEERS only" still applies and the `{@link namespaceForCodec}` remedy from that commit works here too.

src/client/device-registry.ts
```diff
@@ -122,6 +122,7 @@ const CLASS_BY_CODEC: Record<Codec, DeviceClass> = {
   keypad: "other",
   light: "light",
   printer: "printer",
+  display: "other",
```
lenoxys, line 154: Nit: the table's JSDoc enumerates what collapses to `"other"` ("`lock`, `keypad` and `light`") — already stale on `light`, and now missing `display`. Cheap to fix while you're in here.

src/model/classify.ts
martijnpoppen: Agreed and confirmed — `classifyDevice()` in `core/types.ts` echoes `category` verbatim as the residual bucket for anything non-`eufy_security` with no `p2p_did`, so nothing ties `"eufy_mega"` to this product specifically. Switched to your proposed shape in 4908e84: a `/^T87A/` arm in `codecFromModel`, plus `codecFromModel(rec.model) === "display"` in `classify()` ahead of the `device_type` bucket, mirroring `light`/`mower` exactly.

---

src/model/capabilities/__tests__/line-partition.spec.ts
martijnpoppen: Added the pin in 4908e84 — but checked with the PR author first on which value to assert, since your measured `[...,"info"]` set is the real, current, live behavior, not a hypothetical: the security-line grouping genuinely does open this door. Author confirmed pinning the actual leaked set (`["light","doorbell","leak","smoke","co","lock","info"]`) rather than asserting the leak away, so a future change to it is caught rather than silently passing OR silently asserting something untrue. The real device's actual name doesn't trigger any of it (confirmed via the model.spec.ts fixture fix too), so it's an open door, not a live problem.

---

src/model/__tests__/model.spec.ts
martijnpoppen: Fixed in 4908e84 — fixture now includes `name: "Eufy Smart Display"` and the real 8001-8006 params, matching your independent check. Confirmed still `["info"]`, so it's now evidence rather than an assertion about a different record.

---

src/model/param-namespace.ts
martijnpoppen: Consolidated to this one spot in 4908e84 — `capabilities/index.ts` and the `Codec` JSDoc now just point at `{@link namespaceForCodec}` instead of restating it. Kept the `security` grouping itself (per the PR author, after seeing the concrete leak your other comment measured) rather than switching to `display: {}` — but took your "say why, not just what it risks" point: this comment now states plainly that the measured leak is real but the real device's own name doesn't trigger it, rather than only listing hypothetical risk.

---

src/model/capabilities/index.ts
martijnpoppen: Fixed in 4908e84 — same `{@link namespaceForCodec}` remedy.

---

src/client/device-registry.ts
martijnpoppen: Fixed in 4908e84 — `light` dropped, `display` added.

---

martijnpoppen:
@lenoxys all six threads addressed and resolved in 4908e84 (details in each thread's reply) — model-code routing instead of the broader `category` string, the poisoned-name leak pinned to its actual measured value (confirmed with the PR author before pinning, rather than asserting it away), the caveat consolidated to one home, and the two doc nits. Also rebased onto current `beta-0.1.0` (was several commits behind) and fixed the PR body's stale commit/test-count claims. Re-review whenever you get a chance.

---

src/model/registry.ts
```diff
@@ -170,6 +170,11 @@ export const MODEL_REGISTRY: Readonly<Record<string, RegistryEntry>> = {
   T8L40: { codec: "light", name: "Indoor Floor Lamp E10" },
   // 3D printers (eufyMake/AnkerMake) classify from `category` alone — no curated model-code rows, since
   // no printer has been observed on a code yet (rows would guess; the category path already resolves one).
+
+  // eufy_mega category — its own `display` codec (classify.ts). Confirmed live (2026-09-04): the
+  // cloud `device_name` self-reports "Smart Display E10" (param 8005 echoes the same string). No
```
lenoxys: Two claims about the same captured field, in the same PR: this says the cloud `device_name` self-reports `"Smart Display E10"`, while `model.spec.ts:62` — the fixture cited as the evidence — has `name: "Eufy Smart Display"`.

The conclusion survives either way; I ran all three shapes:

```
name: "Eufy Smart Display" → ["info"]
name: "Smart Display E10"  → ["info"]
name absent               → ["info"]
```

But one of the two lines is wrong about a live capture, and the fixture is the one being cited as ground truth. Worth reconciling while the capture is still to hand — `param 8005` echoing the marketing string doesn't tell us what `device_name` held.

src/model/classify.ts
```diff
@@ -270,6 +277,12 @@ export function classify(rec: CloudRecord): Codec {
   // code FIRST — otherwise `codecForType`'s camera bucket could swallow a T8L light below.
   if (codecFromModel(rec.model) === "light") return "light";
 
+  // Smart Display (T87A0) is its own ecosystem too, confirmed live (2026-09-04): it connects over
+  // secure MQTT with no `p2p_did`, never P2P, yet its `device_type` (1, observed) falls inside
+  // `isKnownSecurityType`'s residual range — the same trap `light`/`mower` avoid above. Decided from
+  // the model code (see `codecFromModel`'s `T87A` arm for why), not the cloud `category`.
+  if (codecFromModel(rec.model) === "display") return "display";
```
lenoxys: `codecFromModel(rec.model)` is now called four times in this function (L271, L273, L278, L284) plus once more at L290, on a pure function of one argument. The two adjacent arms collapse with no behaviour change at all:

```ts
const byModel = codecFromModel(rec.model);
if (byModel === "light" || byModel === "display") return byModel;
```

and the same `const` then serves L271/L273 and the L290 fallback. Five calls → one, six arms → five.

(I checked before suggesting it: `mower` and `vacuum` can't join the same set, because the `clean|vacuum|robovac` category rule has to stay between them. `light` and `display` both sit after it, so hoisting just those two is order-preserving.)

lenoxys (commented):
Second pass. All six threads are closed properly, and the model-code routing is the right shape — I checked it rather than taking it on trust:

```
classify: T87A0 | t87a0 | T87A1 | T87A0X | " T87A0 "  → display
classify: {category:"eufy_mega", deviceType:1}         → camera   (unchanged from base)
classify: {category:"eufy_mega", model:"T87B0", …}     → camera   (unchanged from base)
```

So the category is no longer pre-claimed and the rule still widens on a suffix or a second `T87Ax`. `claimsDevice` keys off `p2pDid`, not the codec, so transport routing is untouched — the "nothing user-visible changed" claim holds. `npm run verify` green here too (182 files, 3424 tests).

**The fix is bigger than the PR body says.** Resolving the real record on both branches:

```
beta-0.1.0 : codec "camera"  → ["video","snapshot","motion","person_detection","camera","info"]
this branch: codec "display" → ["info"]
```

The body (and `e03e03a`'s message) says the leaked set was `camera` and `person_detection`. It was six capabilities, including `video` and `snapshot`, on a device with no `p2p_did` to serve either. Worth correcting, since it's the strongest argument for the change.

Two minor notes inline. Neither blocks.

One observation, not a change request: the fix is model-code-only, so a `T87A0` record that arrived without `device_model` still lands in the camera bucket with those six capabilities. `light` has exactly the same hole (`{category:"eufy_life", deviceType:9}` → `camera`), so this is consistent with the established precedent rather than a gap this PR introduces. The root cause is `isKnownSecurityType`'s `0..210` residual claiming a `device_type` with no transport evidence behind it — and `CloudRecord` carries neither `p2p_did` nor `realtime`, so `model/` can't see the field that would settle it. This PR is the fifth arm planted in front of that bucket; the sixth should probably be an issue about the bucket instead. Not something to solve here.

src/model/registry.ts
martijnpoppen: Good catch, and you're right about which one was wrong — the real captured `device_name` was the generic "Eufy Smart Display"; param 8005 separately held the retail name "Smart Display E10". Fixed in 8d251e0: the comment now describes them as two distinct fields (curated from the more specific one), not an echo, and `model.spec.ts`'s fixture stands as the accurate one.

---

src/model/classify.ts
martijnpoppen: Done in 8d251e0, exactly as proposed. Re-verified against your test matrix after the refactor (T87A0/t87a0/T87A1/T87A0X/" T87A0 " → display; `{category:"eufy_mega"}` with no model and a `T87B0` model both still → camera) — unchanged.

---

martijnpoppen:
Second-pass findings addressed in 8d251e0 (both inline threads replied to and resolved). Also fixed the PR body's undercount you caught — the original leaked set really was six capabilities (`video`, `snapshot`, `motion`, `person_detection`, `camera`, `info`), not just two; that's now corrected in the description along with a note on the pre-existing no-`device_model` hole (agreed it's out of scope for this PR, worth its own issue against the residual bucket rather than another arm here).

---

lenoxys (approved):
LGTM.

Both notes addressed, and the collapse is order-preserving — re-ran the routing matrix against `8d251e0` rather than reading it:

```
T87A0 | t87a0 | T87A1 | T87A0X | " T87A0 "  → display
{category:"eufy_mega", no model} | T87B0    → camera
T8L20 → light   T2880B → mower   T2103 → vacuum
ankermake → printer   T8410 → camera   T8010 → station   T8520 → lock   {} → camera
real T87A0 record → display, ["info"], "Smart Display E10"
```

Identical to the pre-refactor branch on every case, and hoisting `byModel` above the printer check is safe since `codecFromModel` is pure. `npm run verify` green (3424 tests).

The `device_name` correction is better than what I asked for — saying the two fields *disagreed* and that 8005 was curated for being the more specific one is the useful version of that note.

PR #119, opened 2026-09-06
docs: add a script that converts the device renders to lossless WebP

lenoxys:
Adds `scripts/device-artwork-to-webp.sh`. **It is not run in this PR** — only the tool lands, so the format switch and the matching extension change on the consumer side can go together in one step.

```
bash scripts/device-artwork-to-webp.sh            # convert, then rewrite the gallery references
bash scripts/device-artwork-to-webp.sh --dry-run  # report, write nothing
```

## Measured, on the current set

| encoding | size | visible difference |
|---|---|---|
| PNG, as committed | 21.8 MB | — |
| oxipng `-o max` | 19.6 MB | none, bit-identical |
| **lossless WebP** | **11.8 MB** | **none, peak delta 0** |
| lossy WebP q90 | 3.7 MB | yes, worst 1.50 MAE / peak 82 |

Lossy is rejected on purpose: several renders carry soft translucent glows that every lossy encoder tested visibly desaturates, and palette-reduced PNG fails the same way. Under 20 MB of assets is not worth degrading artwork a caller displays.

## Verified

Ran against a throwaway copy of `docs/`, then compared all 153 outputs to their originals:

- dimensions preserved on every file
- composited over white and black: 0.0000 mean absolute error, peak delta 0
- gallery rewritten to 142 `.webp` references, 0 `.png` left
- re-run is a no-op that exits 0
- an undecodable input aborts with exit 1 and every PNG still present

One trap worth recording: libwebp rewrites the colour channels beneath fully transparent pixels, since nothing can observe them and zeroing them compresses better. A raw byte comparison therefore reports a large difference for invisible data. Any re-check has to composite first, or it will conclude the encoder is broken. The script header says so.

## Dependency

None added. `npx --yes sharp-cli@6` fetches the encoder for the length of the run — this is needed once per artwork refresh, not on any build, so it does not earn a place in `devDependencies`.

## Note for whoever runs it

Callers that build an artwork path from a model code need the same `.png` -> `.webp` change; the script prints that reminder. Independent of #118, which only downscales two oversized renders.


lenoxys:
Folded into #118, which now carries the script and the converted assets together.

PR #123, opened 2026-09-07
refactor: carry only what is read, derive what was stated twice

lenoxys:
## What changed

−4,760 lines and one runtime dependency. Three deps now, not four; one less CI guard.

Param dictionary entries carry name, type, provenance and encoding. The five other fields per entry serve no reader: two have none anywhere, one restates the key its entry is filed under, one is only ever read off a capability's own property spec, and one is true on all 329 — membership in the table already carries that, and the module doc now says so. Per-entry sweep provenance lives in git history, next to the entry it justifies.

The P2P command id→name map is the inversion of the id enum. The separate hand-maintained copy held 536 ids against the enum's 540, five of them disagreeing.

The WebRTC transport is gone and its engine dependency with it. No layer above the transport referenced it, its signaling was unfinished by its own barrel doc, and a CI guard existed solely to keep the engine's types out of the published declarations. The README, the guides and the ffmpeg notes no longer offer a container path that had no caller.

The capability binder and every `actions()` take the dependency bag that already existed. The seven positional parameters carried the same four optional providers in two different orders; the `ff09` reader now sits in the bag beside the other three, so a provider member indexes it by name. One module no longer declares two placeholder parameters to reach the fifth, and the spec shim over it is a passthrough.

Storage and person-detection contribute no state properties. No owned station reports a capacity or SD-card id and no owned camera reports a detection id, so those getters never installed — except one, which pointed at an id the dictionary names `deviceStatus`, reported by ten devices with an unrelated meaning, and answered a status code as free megabytes. Push events were already person-detection's whole read surface.

Property name and paramType shape is asserted once across every capability module, which also covers the modules that never had it.

A dozen symbols with no reader are gone, along with the facade forwarders the raw client accessor already reaches and the map, bulk-inspect, rename and raw-control entry points no example, guide, spec or caller named. Five hand-rolled sleeps, a hand-rolled base64url and a re-inlined clamp use the standard library.

## Why

Two tables each stated one thing twice and both had drifted. A helper for the first already existed. The dictionary shipped ~1,900 lines of sweep bookkeeping to every consumer; git holds it better. The WebRTC stack was a dependency and a CI guard maintained for a path nothing could reach.

## Evidence

n/a for the wire — no parameter id, polarity or frame shape moved.

The one substantive claim is the storage getter: the id it read is reported by ten devices on this account under a different meaning, so removing it stops a wrong answer rather than tidying an unused one.

## Verified how

- [x] `npm run verify` green — 179 files, 3,363 tests, public surface 277 → 264 exports
- [x] **Ran against the fleet**, service stopped for the duration. 14 devices resolve with their capability sets and named properties intact.

  Livestream through the ffmpeg-free readable, **every camera except the one disabled on this account**:

  | model | chunks | bytes |
  |---|---:|---:|
  | T8442 | 179 | 1,275,739 |
  | T8170 | 156 | 778,845 |
  | T8114 | 166 | 732,469 |
  | T8400 | 171 | 531,592 |
  | T8114 | 164 | 479,151 |
  | T8210 | 147 | 407,344 |
  | T8171 | 146 | 406,078 |

  Writes, each read back off the wire and then restored:

  | write | observed |
  |---|---|
  | arming mode | `1 → 0 → 1` |
  | camera spotlight | `on()`, `off()` |
  | camera enable | `true → false → true` |

  A final pass re-read all three: arming `1`, spotlight off, both toggled cameras back where they started.  No siren.
- [x] Specs updated
- [x] Docs — README, getting-started, troubleshooting and CONTRIBUTING no longer name the removed transport or the dropped dependency

**The fleet is security-only** — no vacuum, lock or light. Those paths carry spec cover alone, which is why their tables are untouched here.

## Kept

Wire catalogs stay complete transcriptions. A partial transcription of someone else's protocol costs the next reader more than an unused constant does.

The 24 `unverified` writes on vacuum, dock and lock keep their frame builders. An absent setter and a throwing intent path is what `unverified` is specified to do, so an unreachable builder there is a derived frame awaiting a capture, not dead code.

The HTTP face roster stays, and `securityAppPost` stays public with it. A probe script in the reverse-engineering repo calls `getFaces()` and carries a TODO naming `securityAppPost` as the door it needs, so the same argument that protects the P2P cluster protects this — a different path, which is why a grep for one found nothing about the other.

The P2P database cluster stays — the face roster and history-record reads on the session object. Nothing in this repository calls them, which is exactly why they looked dead: the caller is the Home Assistant bridge in this org, which drives `queryDatabase` and `requestFaces` and listens on the `dbChunk` event to resolve a `personDetected` push to a name. A repo-local grep is not evidence of disuse for a private SDK with sibling consumers.

The CO, smoke and leak detected-state ids stay. This account owns no such sensor, so their absence from the param sweep says nothing about the ids.

Five timing and buffering options no caller sets today tune paths against real hardware. The convergence poll keeps its inline timer, which is what its specs' fake clock patches.


src/model/capabilities/light.ts
```diff
@@ -3,7 +3,7 @@ import { DeviceType } from "../device-types.js";
 import { WALL_LIGHT_TYPES } from "../device-family.js";
 import { setScalar, setJson, setPayload, describeDevice } from "./access.js";
 import type { CapabilityModule, CapabilityActions, CommandContext } from "./types.js";
-import { propertiesOf, type Members, type Surface } from "./members.js";
+import { propertiesOf, type Members, type Surface, type MemberDeps } from "./members.js";
 import type { Command, CommandSink } from "../../core/contracts.js";
```
martijnpoppen: `CommandSink` is now unused here — `actions()` takes it off `MemberDeps`, and this import is its only remaining occurrence in the file. `noUnusedLocals` isn't on, so `tsc` won't catch it.

```suggestion
import type { Command } from "../../core/contracts.js";
```

---

src/model/capabilities/camera.ts
```diff
@@ -3,7 +3,7 @@ import { DeviceType } from "../device-types.js";
 import { isIndoorCamera, isIndoorCamMini, isIndoorPanTiltS350, isOutdoorPanTilt } from "../device-family.js";
 import { setScalar, setPayload, hasCapability } from "./access.js";
 import { AUDIO_CMD } from "./audio.js";
-import { accepts, propertiesOf, provided, type Members, type Surface } from "./members.js";
+import { accepts, propertiesOf, provided, type Members, type Surface, type MemberDeps } from "./members.js";
 import type { CapabilityModule, CapabilityActions, CommandContext } from "./types.js";
 import { CameraDisabledError, type Command, type CommandSink, type MediaProvider } from "../../core/contracts.js";
```
martijnpoppen: Same here: `CommandSink` no longer appears anywhere in this file except this import. (`MediaProvider` and `Command` are both still used, so only the one name drops.)

```suggestion
import { CameraDisabledError, type Command, type MediaProvider } from "../../core/contracts.js";
```

---

src/model/capabilities/smart-light.ts
```diff
@@ -7,7 +7,7 @@ import type {
   CapabilityActions,
 } from "./types.js";
 import type { Command, CommandSink } from "../../core/contracts.js";
```
martijnpoppen: `CommandSink` is unused after the signature change — this import is its only occurrence in the file.

```suggestion
import type { Command } from "../../core/contracts.js";
```

Same file, just above the hunk: `CommandContext` on line 5 also has no remaining use (the old `actions(_ctx: CommandContext, sink)` was the only one), so it can come out of the `./types.js` import too.

---

src/model/capabilities/suction.ts
```diff
@@ -1,7 +1,7 @@
 import type { AvailabilityContext, CapabilityModule, CapabilityActions, CommandContext } from "./types.js";
```
martijnpoppen: `CommandContext` is unused now that `actions()` destructures `ctx` off `MemberDeps` — this line is its only occurrence in the file.

```suggestion
import type { AvailabilityContext, CapabilityModule, CapabilityActions } from "./types.js";
```

---

src/client/eufy-mega.ts
martijnpoppen, line 1: **Dangling `{@link}` left by the `mapFor` removal.** Line 305, on the `mapStores` field doc, still reads:

> a store for a robot that has never sent a map would answer `undefined` to everything, which `{@link EufyMega.mapFor}` already does without allocating anything.

`mapFor` is gone as of this commit, so that link resolves to nothing — TypeDoc will warn on it when `docs.yml` next builds the API reference. `mapFor` is also the only removed symbol in this PR that leaves a dangling reference behind; I grepped the other 20-odd (`renameDevice`, `routeControlRaw`, `inspectAllDevices`, `formatInspection`, `decryptJson`, `COMMAND_NAMES`, `STORAGE_MEMBERS`, `WebRTCPeerHandle`, …) and they're all clean.

Worth rewording the field doc while you're there, since its whole justification was pointing at the accessor that no longer exists.

Separately, on the removal itself: `mapFor` was the only *pull* path to `mapStores`, so the `map` event is now the sole way to reach an assembled map. That's a defensible narrowing — the store is in-memory and empty after a restart anyway — but it does mean a listener attached after a robot's map has already converged sees nothing until the next frame that actually changes the map (`store.apply(piece)` returning true is what gates the emit). If that's the intent, fine; flagging it because `VacuumMapStore` stays exported from `model/index.ts`, which reads like there's still a way to get one.

---

martijnpoppen (commented):
Reviewed the whole diff and re-ran the verification independently rather than taking the PR body's word for it. The substance holds up — the claims I could check all reproduce exactly. Not approving only because of a handful of leftovers from the refactor itself, all of them small and mechanical; see below and the inline notes.

## What I verified

`npm run verify` green on `a1106a5`: **179 files, 3363 tests, 264 exports** — the numbers in the description, to the digit.

**The command-map inversion.** Parsed both tables out of the base revision and diffed them against `enumLabels(CommandType)`. The description's arithmetic is exact: enum has 545 declarations over **540** unique ids, the hand map had **536** entries, and **five** disagree —

| id | hand map | derived (later declaration) |
|---|---|---|
| 1040 | `CMD_DEVS_UNBIND` | `P2P_RESET` |
| 1238 | `CMD_SET_UPGRADE` | `P2P_ENTER_OTA` |
| 2107 | `CMD_SMART_DROP_IS_PIN_REQUIRED` | `SUB1G_REP_RUNTIME_STATE` |
| 2109 | `CMD_SMART_DROP_HAS_MASTER_PIN` | `SUB1G_REP_BIND_FAILED` |
| 1946165248 | `CMD_SMARTSAFE_IS_ENABLE_BEEP` | `CMD_SMARTSAFE_PARAM_BASE` |

Critically, **no id in the hand map is absent from the enum**, so nothing falls back to `CMD_<id>` that didn't before — the four new ids are pure gain. The five collisions are all genuine id reuse in the vendor's own space, the new doc comment names every one of them, and the only consumer is `p2p-session.ts:1766` populating a log/telemetry field. Deriving this is plainly right.

**The storage getter.** This is the one behavioural claim and it checks out. Base `param-dictionary.ts:289` has 1131 as `name: "deviceStatus"`, `type: "bool"`, `note: "app:DEVICE_STATUS"`, observed on 13 models. `STORAGE_MEMBERS.free` declared it `type: "number"`, `unit: "MB"`, `kind: "megabytes"`. So the getter that did install was reading a device-status code and answering it as free megabytes. Removing it stops a wrong answer, as the description says — that's the strongest change in the PR.

**Signature collapse to the dependency bag.** `deps[m.needs]` in `bindMembers` is sound: `Providers` and `MemberDeps` agree on both key names (`media`, `ff09Settings`), so the indexed access is type-checked rather than positional — which is the actual win over `m.needs === "media" ? media : ff09Settings`. `read` moving from an optional positional to a required `MemberDeps` field is handled at both call sites (`device.ts` passes the real reader, `__tests__/bind.ts` defaults to `() => undefined`).

**Test coverage isn't lost.** The eight per-module `"every property has a string name + numeric paramType"` tests are replaced by one cross-module assertion in `property-id-integrity.spec.ts` that walks `CAPABILITY_MODULES` — strictly wider, as claimed.

**Spot checks on the small rewrites.** `generateFid`'s `base64url` swap is byte-identical for the first 22 chars (the manual `+/` → `-_` replace is exactly what `base64url` does; the stripped `=` is past the slice). The `McsParser.waiting` removal is a true no-op — both branches of the old `handleData` called `waitForData()` unconditionally, and `waiting` was read nowhere else. `buildIntCommandPayload` really is a dead legacy variant; the live 1003 path goes through `sendMediaPayloadLevel2`, and `buildIntStringCommandPayload` stays.

**Docs.** README, AGENTS.md, getting-started, troubleshooting, CONTRIBUTING, `transport/README.md` and `copilot-instructions.md` all agree with `package.json` on three runtime deps, and the transport folder lists are corrected to `tuya` rather than just having `webrtc` deleted. `gen_param_dictionary.py` doesn't exist in the repo, so dropping the "GENERATED by" header is a correction, not a loss.

## Leftovers to fix

1. **A dangling `{@link EufyMega.mapFor}`** at `eufy-mega.ts:305`, pointing at the method this commit removes. Inline note on the file. It's the only such reference — I grepped every removed symbol and the rest are clean.
2. **Four dead type imports** the `actions()` signature change left behind — `CommandSink` in `light.ts`, `camera.ts` and `smart-light.ts`, `CommandContext` in `smart-light.ts` and `suction.ts`. `noUnusedLocals` isn't enabled so `tsc` stays quiet. Inline suggestions on each.
3. **`param-namespace.ts:8-10`** still carries the claim this PR removes from `param-dictionary.ts`'s own header:

   > This lives outside `param-dictionary.ts` because that file is generated (`scripts/gen/gen_param_dictionary.py`) — keeping the namespace union and the dispatch here means a regeneration can't drop a namespace.

   That path doesn't exist (and never did at that spelling — the old header said `scripts/gen_param_dictionary.py`). Since the dictionary is now hand-maintained, the module's stated reason for existing separately is no longer true; it's still the right place for the namespace union, but the *why* needs rewriting. `guard-doc-refs.sh` won't catch this — it only scans for `docs/` and `.md`.

## One question, not a blocker

`getFaces` / `getFacesByIds` are removed from `MegaHttpClient` and `securityAppPost` narrowed to `private`, leaving `getCiphers` as its only caller. That runs against the argument the "Kept" section makes for the P2P database cluster — *"a repo-local grep is not evidence of disuse for a private SDK with sibling consumers."* The HTTP roster is a different path from the `queryDatabase`/`requestFaces`/`dbChunk` one the HA bridge drives, and the removed doc admitted it "returns an empty roster for accounts whose faces live on the HomeBase", so I suspect this is fine — but it's the same class of symbol the PR deliberately protects elsewhere, so worth a positive check against the bridge rather than a grep. The `private` on `securityAppPost` also closes the generic escape hatch to the `eufylife.com` host specifically (`request()` only reaches mega hosts), which is a separate small narrowing.

Fix the three leftovers and I'd sign off — the reasoning throughout is unusually well evidenced, and the storage getter alone justifies the change.

---

lenoxys:
Thanks — I re-checked all four independently rather than taking them as read. **All three leftovers are real, and the question turned out to be a real finding too.** Fixed in `e677b20`.

## 1. Dangling `{@link EufyMega.mapFor}` — real

`eufy-mega.ts:305`. The sentence leaned on `mapFor` to explain lazy allocation, so it now states the reason directly rather than pointing at a method that no longer exists. I also swept every removed symbol for `{@link}` references — that was the only one, matching your grep.

## 2. Dead type imports — real, and there were **14**, not 5

Rather than fix the five by hand I ran the compiler with the check on, against this branch **and** against the base, and kept only the difference. The base has five of its own (`isOutdoorPanTilt`, `ParamValue`, `types.ts` `RawDpCodec`, `AnyDeviceEvent`, `QUERY_COMBINATION_WITH_AI`) — untouched here.

Attributable to the signature change:

| file | dead specifiers |
|---|---|
| `capabilities/index.ts` | `CapabilityStateReader`, `CommandSink`, `MediaProvider`, `Ff09SettingsReader`, `RawDpCodec` |
| `capabilities/types.ts` | `CommandSink`, `MediaProvider` |
| `smart-light.ts` | `CommandContext`, `CommandSink` |
| `camera.ts` / `light.ts` / `motion.ts` | `CommandSink` each |
| `suction.ts` | `CommandContext` |

`motion.ts` and the two barrels weren't in your list. Two more the same class, from other removals in this PR:

- `command-router.ts` imported `buildDeviceNameBody` for the `renameDevice` this PR drops.
- `McsParser.handshakeComplete` was left assigned and never read once `isHandshakeComplete` went — the getter was its only reader.

Both are the same defect as the imports: `noUnusedLocals` is off, so nothing complained.

## 3. `param-namespace.ts:8-10` — real

Worse than stale: `scripts/gen/gen_param_dictionary.py` never existed at that spelling *or* the header's. It now states what the module is — one small table a reader can hold at once, beside the guard that a codec resolves to exactly one namespace — instead of a reason that was never true. You're right that `guard-doc-refs.sh` can't see this class; it only scans `docs/` and `.md`.

## 4. `getFaces` — you were right to push, and a grep was not enough

I did the positive check across the whole org, and it found a caller: **`mega_preview/scripts/control/media_dump.mjs:112`** calls `eufy.api.getFaces()`, with

```
// TODO(eufylife-keyident): unblock getFaces — see MegaHttpClient.securityAppPost
```

It's inside a `try/catch`, so removal wouldn't crash it — it would degrade silently and make that TODO unreachable, which is the worst of the options. `securityAppPost` going `private` would have closed the very door the TODO names.

So `getFaces`, `getFacesByIds` and a public `securityAppPost` are restored. Your instinct was exactly right: this is the same class of symbol the PR protects elsewhere, and the argument has to hold in both directions or it isn't an argument. Your reasoning on *why* it looked safe was also correct — the HTTP roster and the `queryDatabase`/`requestFaces`/`dbChunk` path the HA bridge drives are genuinely different paths, which is why a grep for one says nothing about the other.

For the record, the positive check on the bridge specifically: `getFaces`/`getFacesByIds`/`securityAppPost`/`aiassis`/`get_faces` → zero occurrences; its only `.api` use is `eufy.api?.auth?.userId`; it resolves a `person_id` through `person_basic_info` over P2P. Its suite is 14/14 against this branch.

## State

`npm run verify` green — 179 files, 3363 tests, 264 exports. Net now **−4,760** (was −4,764; the roster comes back). Compiler-with-the-check-on reports only the base's own five.

Also re-verified downstream against this branch: the HA bridge 14/14, and the Homebridge plugin compiles and runs — 13 accessories, `runtime-ready`, no errors, every HomeKit service still attaching, which is what would break first if the dependency bag had dropped a provider.

One correction to the review body, no consequence: the per-module assertion was in **20** specs, not eight — `arming, battery, co, contact, doorbell, keypad, leak, light, locate, lock, motion, person-detection, siren, smoke, snapshot, storage, suction, vacuum-clean, vacuum-dock, video`. Your conclusion that the replacement is strictly wider is unchanged.

martijnpoppen:
Following up on the one question I left open in my review — I've now checked it against the sibling repos, and it resolves against the PR. **Removing `getFaces` breaks `mega-yfue/homey-eufy-app` at compile time.**

I'd asked for a positive check rather than a grep, on the grounds that the description's own rule — *"a repo-local grep is not evidence of disuse for a private SDK with sibling consumers"* — applies to `getFaces` as much as to the P2P database cluster. It does, and the consumer is the Homey app, not the HA bridge I'd guessed at.

`homey-eufy-app` depends on this SDK by path — `"@mega-yfue/eufy-sdk": "file:../eufy-sdk"` — so it compiles against this working tree, and it calls the removed method in two places:

`app.ts:199-208`, the account-level face roster:

```ts
get faces(): FaceRoster {
  return (this.faceRoster ??= new FaceRoster({
    fetch: async () => {
      const eufy = this.client;
      if (!eufy) throw new Error('not signed in');
      return await eufy.api.getFaces();
    },
    …
  }));
}
```

and `lib/faces.ts:113`, where it is the documented contract of the injected dependency:

```ts
/** `eufy.api.getFaces()`. Answers the account's enrolled faces, in a shape this does not assume. */
fetch(): Promise<unknown>;
```

Note it reaches the method through the public `api` accessor, so making `securityAppPost` private doesn't save it — `getFaces` itself is the surface being consumed.

Worth reading `lib/faces.ts:129` before deciding this is dead code, because that app has already absorbed the caveat the removed doc carried:

> which is what happens on a HomeBase-3 account, where `getFaces` *"returns an empty roster … the faces live on the HomeBase"* — costs one request every few minutes rather than one per detection

So the empty-roster behaviour on HB3 accounts isn't evidence the endpoint is useless; it's a known case that consumer already rate-limits around. The account this PR's fleet testing ran on is HB3-based, which is exactly why a local probe would make the method look dead.

Two ways forward, both fine by me:

1. **Keep `getFaces` / `getFacesByIds`** (and `securityAppPost` public) and drop them from this PR's removal set — the smallest change, and consistent with the "Kept" section's own reasoning.
2. **Remove them deliberately**, and land the replacement in `homey-eufy-app` in the same cycle — the P2P `requestFaces` / `dbChunk` path this PR keeps is the better roster on HB3 anyway, so this could be a genuine upgrade rather than a break. But it's a coordinated two-repo change, not a cleanup.

What it shouldn't be is silent. Everything else in my review stands — the three doc/import leftovers are still the only other items, and the substance of the PR still checks out.

For completeness, I swept `homey-eufy-app` for every other symbol this PR removes (`mapFor`, `renameDevice`, `routeControlRaw`, `inspectAllDevices`, `formatInspection`, `getProductDataPoint`, `getDeviceParamList`, `megaRequest`, `decryptJson`, `splitAdtsFrames`, `COMMAND_NAMES`, `ServerPushEvent`, `GarageDoorPushEvent`, `SmartSafeEvent`, `HB3PairedDeviceMessageType`, `STORAGE_MEMBERS`, `PERSON_DETECTION_MEMBERS`, the `webrtc` barrel, …) — **`getFaces` is the only hit.** Its `personDetected` usages are all the push-event name, which this PR keeps, so those are safe. I haven't checked `ha-eufy-sdk-bridge` or `ha-eufy-sdk`; worth the same sweep before merge.

---

lenoxys:
Good catch on the consumer, and thanks for doing the sweep. `getFaces` / `getFacesByIds` / public `securityAppPost` were already restored in `e677b20`, which crossed with this — so we agree on the outcome. But I want to correct **why**, because the reason matters more than the result here.

## The SDK does not keep a wire because a consumer calls it

Direction of travel: this SDK is upstream. Consumers adapt to it, not the reverse. If the HTTP roster were genuinely redundant, the right move would be your option 2 — remove it, and `homey-eufy-app` follows. "It would break a sibling repo" is a scheduling problem, not an argument for surface.

So I went looking for whether it *is* redundant. It isn't, and the evidence is in the P2P path this PR keeps:

> `requestFaces` — *"The **HomeBase** replies with the level-1 `CMD_DATABASE` response … `account_id` must be the station `admin_user_id` (a wrong/absent id **or a camera session** answers result `-104`)."*

The P2P roster is read **off a HomeBase**. An account with no HomeBase — standalone cameras only, a topology this SDK supports and documents throughout — has no P2P roster path at all. The HTTP roster is the only wire there.

That inverts your framing of option 2: the P2P path isn't "the better roster on HB3 **anyway**" in a way that makes HTTP droppable, because on a HomeBase-less account there is nothing to be better than. The two are complementary — cloud-held faces via HTTP, HomeBase-held faces via P2P — which is also why the removed doc's "empty roster on HB3" caveat reads as a limitation rather than a verdict.

So it stays on the SDK's own evidence about device topology. That the Homey app happens to call it is a consequence, not the reason.

## A second consumer, and the sweep you asked for

I'd found a different caller before yours: `mega_preview/scripts/control/media_dump.mjs:112` calls `eufy.api.getFaces()` with

```
// TODO(eufylife-keyident): unblock getFaces — see MegaHttpClient.securityAppPost
```

It's inside a `try/catch`, so removal wouldn't have crashed it — it would have degraded silently and made that TODO unreachable, and `private securityAppPost` would have closed the exact door the TODO names. Worse than a compile error, which is at least loud. Your `homey-eufy-app` hit is the compile-time one; together that's two repos.

The sweep you hadn't done, now done:

| repo | result |
|---|---|
| `ha-eufy-sdk-bridge` | zero hits for `getFaces`/`getFacesByIds`/`securityAppPost`/`aiassis`/`get_faces`; only `.api` use is `eufy.api?.auth?.userId`. Resolves a `person_id` through `person_basic_info` over P2P. **Suite 14/14 against this branch.** |
| `ha-eufy-sdk` | Python HACS integration, talks WS to the bridge. No Node SDK dependency. |
| `ha-eufy-sdk-addon` | thin wrapper, `FROM ghcr.io/…/ha-eufy-sdk-bridge` pinned. Unaffected. |
| `homebridge-eufy` | zero hits. Compiles and runs against this branch — 13 accessories, `runtime-ready`, no errors. |

Which also explains why your first pass and mine both missed it: the HTTP roster and the `queryDatabase`/`requestFaces`/`dbChunk` path are different wires, so a grep for one says nothing about the other. Exactly the failure mode the "Kept" section was written about — I just applied it selectively the first time, and you were right to keep pulling.

## The three leftovers

All real, all fixed in `e677b20` — and there were more than five dead imports. I ran the compiler with the unused check on against this branch **and** the base and kept only the difference: **14** specifiers, not 5. `motion.ts` and both barrels (`capabilities/index.ts` ×5, `capabilities/types.ts` ×2) weren't in your list, plus two of the same class from other removals — `command-router.ts` importing `buildDeviceNameBody` for the dropped `renameDevice`, and `McsParser.handshakeComplete` left assigned and never read once its only reader went. The base's own five are untouched.

`param-namespace.ts` was worse than stale: that generator path never existed at *either* spelling. And you're right that `guard-doc-refs.sh` structurally cannot catch it — it only scans `docs/` and `.md`.

## Also in since your review

`87ad263` — a documentation pass over all of `src/`, on the rule that a JSDoc states the declaration and not its audience or its own history. 72 files, comments only, zero code lines touched. It removed the `so a host can…` / `useful for tooling` / second-person register throughout, and about thirty history clauses — several of which had already drifted out of agreement with the code above them.

It also caught a boundary violation neither of us flagged: `transport/http/light-catalog.ts` and the MQTT router **named model capability accessors** (`dev.smartLight()?.setEffect(…)`, `dev.lock()?.getAutoLockState()`) in their JSDoc, while the rule forbids a transport naming a capability at all. `guard:decorrelation` greps imports, so prose walks straight past it. The catalogue module's header even claimed "Nothing here names a capability" two lines above one that did. AGENTS.md now states both rules so it isn't rediscovered.

`npm run verify` green throughout — 179 files, 3363 tests, 264 exports.

One small correction to the first review, no consequence: the per-module assertion was in **20** specs, not eight.

src/client/eufy-mega.ts
martijnpoppen, line 1: One survivor of the `noUnusedLocals` sweep — **line 34 is a fully dead import declaration**:

```ts
import { parseCleanRecordDetail, type CleanRecordDetail } from "../model/clean-record-detail.js";
```

Neither name has another occurrence in the file; both were left behind when `a1106a5` removed the `parseCleanRecordDetail()` facade method. `grep -n "parseCleanRecordDetail\|CleanRecordDetail" src/client/eufy-mega.ts` returns only line 34.

Worth knowing *why* the sweep missed it, since the same gap will recur: a whole-declaration removal surfaces as **TS6192** (`All imports in import declaration are unused`), not the TS6133/TS6196 that the individual specifiers raise. Running the branch and the base and diffing confirms it's this PR's, not pre-existing:

```
HEAD  src/client/eufy-mega.ts(34,1): error TS6192: All imports in import declaration are unused.
HEAD  src/client/eufy-mega.ts(84,3): error TS6196: 'AnyDeviceEvent' is declared but never used.
BASE  src/client/eufy-mega.ts(84,3): error TS6196: 'AnyDeviceEvent' is declared but never used.
```

`AnyDeviceEvent` is the base's own and rightly untouched; the TS6192 is new. Every other file matches the base exactly (the base's fifth, in `webrtc_protocol.spec.ts`, is gone with the folder).

Deleting the line is safe — `parseCleanRecordDetail` is still exported from `model/index.ts`, so consumers keep it; only the facade's own unused import goes.

---

martijnpoppen (approved):
Re-reviewed at `87ad263`. All three findings are fixed, the face roster is back, and both new commits check out under independent verification. Approving — the one item left is a single dead import line, noted inline, not worth another round.

## The fixes, verified

**Dangling `{@link EufyMega.mapFor}`** — reworded to `so allocating one buys nothing`, which says the same thing without the reference. I re-swept every symbol this PR removes (`mapFor`, `renameDevice`, `routeControlRaw`, `inspectAllDevices`, `formatInspection`, `decryptJson`, `splitAdtsFrames`, `buildIntCommandPayload`, `isHandshakeComplete`, `ServerPushEvent`, `GarageDoorPushEvent`, `SmartSafeEvent`, `HB3PairedDeviceMessageType`, `STORAGE_MEMBERS`, `PERSON_DETECTION_MEMBERS`, `WebRTCPeerHandle`, `gen_param_dictionary`, `werift`, …) across non-spec `src/` — **no dangling reference survives**. The remaining hits are live definitions: `COMMAND_NAMES` is the module-local const, `getDeviceMapList`/`getManyDeviceMapContent` are still on `MegaHttpClient` (only the facade forwarders went), and `buildDeviceNameBody` is still exported and cited from `envelope.ts:27` — caller-less now, but that's the same "derived frame awaiting a capture" standing the `Kept` section claims for the 24 `unverified` builders, so it's consistent rather than an oversight.

**`param-namespace.ts`** — the generator justification is replaced by the real one (one small table beside the one-namespace-per-codec guard). That's the better reason; it was true before the script ever existed.

**Dead imports** — cleared, and the sweep found more than I flagged. Two I'd missed entirely, both genuine:

- `buildDeviceNameBody` still imported by `command-router.ts` after `renameDevice` went — an unused *value* import, not just a type.
- `McsParser.handshakeComplete` left assigned-and-never-read once `isHandshakeComplete` went. That one is the better catch: a write-only private field is invisible to a reader-side grep, and the removal correctly takes the `LoginResponse` assignment with it. `MessageTag` stays live via `TAG_TO_TYPE`, so the import is still needed — checked.

**The face roster** — `getFaces`, `getFacesByIds` and a public `securityAppPost` are back with their original signatures. I proved the restoration is sufficient rather than assuming it: built `dist/` at this head and typechecked the exact consumer call shapes against it under `--strict`, and both compile clean.

```ts
const eufy = client;
if (!eufy) throw new Error('not signed in');
return await eufy.api.getFaces();   // homey-eufy-app app.ts:205
```

## `87ad263` is genuinely docs-only

72 files and −696 lines is a lot to take on trust, so I stripped every comment marker from the whole commit diff and looked at what was left. **The only non-comment change in the entire commit is nine lines of new prose in `AGENTS.md`** — the two rules themselves. No executable line moved. `npm run verify` still lands on exactly the same numbers as `a1106a5`: **179 files, 3363 tests, 264 exports.**

Codifying both rules in `AGENTS.md` rather than only applying them is the right call — the audience rule in particular ("naming a consumer dates the doc the moment another one appears") is the kind of thing that silently decays without somewhere to point at.

## One correction to the record, worth carrying forward

The `Kept` section now attributes the face-roster reprieve to *"a probe script in the reverse-engineering repo"*. That's real, but it's the weaker of the two callers. The binding one is **`mega-yfue/homey-eufy-app`**, which depends on this SDK as `"@mega-yfue/eufy-sdk": "file:../eufy-sdk"` — a path dependency, so it compiles against this working tree — and calls the method in two places:

- `app.ts:199-208`, the lazily-built account-level `FaceRoster`;
- `lib/faces.ts:113`, where `eufy.api.getFaces()` **is** the documented contract of the injected `fetch()` dep.

A script with a TODO would have been a soft break. This is a compile error in a sibling app on the next `npm run sync:sdk`. Worth naming in the `Kept` section instead of the probe, because it also settles the HB3 question the removal originally rested on — `lib/faces.ts:129` already absorbs the empty-roster caveat and rate-limits around it:

> which is what happens on a HomeBase-3 account, where `getFaces` *"returns an empty roster … the faces live on the HomeBase"* — costs one request every few minutes rather than one per detection

So the empty roster on this PR's own HB3 fleet was never evidence the endpoint was dead; it's a case a consumer had already handled. That's the sharpest version of the argument the `Kept` section is making, and it's stronger than the one currently written there.

For completeness: I swept `homey-eufy-app` against every other symbol this PR removes and `getFaces` was the only hit — its `personDetected` uses are all the push-event name, which this PR keeps. `ha-eufy-sdk-bridge` and `ha-eufy-sdk` still haven't had that sweep; cheap insurance before merge, given this one landed.

Everything substantive from the first pass still stands — the 540-vs-536 command-map arithmetic, the `1131`/`deviceStatus` storage fix, and the widened property-shape assertion. Nice work on the turnaround.

---

max246 (approved):

lenoxys:
Both inline notes were right, and the second one found a gap in my method rather than just a leftover — so I've closed the class instead of the instance.

## The dead import declaration

Confirmed and fixed. `grep -n "parseCleanRecordDetail\|CleanRecordDetail" src/client/eufy-mega.ts` returned only line 34, and it is this PR's: `a1106a5` removed the facade method and left its import.

Your diagnosis of why my sweep missed it is the useful part. I had been grepping the compiler output for `"is declared but"`, which matches TS6133 and TS6196 — the per-specifier errors. A whole dead declaration raises **TS6192**, different wording, so the grep walked straight past it. That blind spot would have recurred on the next removal.

So rather than delete the line, I turned the check on permanently — `1b7ce61` sets `noUnusedLocals` and `noUnusedParameters` in `tsconfig.json`. Eighteen sites had to be cleared first:

- **this PR's:** the TS6192 above
- **the base's own, now also gone:** `AnyDeviceEvent`, an unread catch binding, `isOutdoorPanTilt` in `camera.ts`, `ParamValue` + `RawDpCodec` in `capabilities/types.ts`
- **specs:** six unused imports, two `topic` parameters → `_topic`, and a `Talkback` constructed for its side effect that bound a name it never read

Two wanted more than deletion:

- `QUERY_COMBINATION_WITH_AI` is a wire constant nothing calls yet, and this PR argues elsewhere that a wire catalogue stays a complete transcription. Deleting it to satisfy a lint would contradict that, so both inner query cmds now sit in one `DB_QUERY` object — the id and its evidence stay, and the catalogue becomes a referenced value.
- `readable_egress.spec.ts` asserted `"audio" extends keyof ReadableOptions ? never : true` in a const nothing read. It is now an `it()` that asserts it, so the check runs where a reader looks for it — hence 3364 tests rather than 3363.

`tsconfig.build.json` extends `tsconfig.json`, so the published build is checked on the same terms.

## `mapFor` — the dangling link, and the narrowing

The `{@link}` went in `e677b20`; the field doc now states the lazy-allocation fact directly instead of leaning on a method that no longer exists. And thank you for grepping the other twenty — I did the same sweep after and agree it was the only one.

On the narrowing, I checked the emit before deciding, and it is narrower than you feared:

```ts
if (store.apply(piece)) this.emit("map", { deviceSn, map: store.snapshot });
```

The event carries **`store.snapshot` — the whole assembled map**, not the piece that changed, and `client/types.ts` says so explicitly: *"Carries the whole snapshot rather than the piece that changed, because the pieces are only useful together."* So everything `mapFor` could answer arrives on the event. What is genuinely gone is the *pull*: a listener attached after convergence waits for the next real change, since `store.apply` returning true is what gates the emit.

I left it removed, on the reasoning that the pull buys little where it would still be empty: the store is in-memory, so after a restart `mapFor` answered an empty store too, and within one process a consumer that cares attaches before connect and holds the last snapshot — one line on its side. If a cold-read path is wanted it should be a deliberate API decision with a name that says so, not a getter over a private map.

`VacuumMapStore` staying exported is not incoherent, though I see why it reads that way: it is a public class with a public constructor whose documented purpose is *"Created empty and filled by `VacuumMapStore.apply`"* — a consumer assembling pieces from the event stream instantiates its own. What is unreachable is the SDK's particular instance, not the type.

Happy to add the pull back under a clearer name if you'd rather not lose the cold read — say so and it's a two-line follow-up rather than a change to this PR.

## State

`npm run verify` green — 179 files, 3364 tests, 264 exports, and the compiler now refuses an unused local or parameter anywhere in the project.

PR #124, opened 2026-09-08
feat(client): expose listLightEffects() for the effect gallery

max246:
## What
Add a thin public delegator `EufyMega.listLightEffects(opts?)` -> `LightEffectSummary[]`.

## Why
`listLightEffects` (transport/http/light-catalog.ts) took the private `MegaHttpClient`, so nothing outside the client could browse the `lightId`s `dev.smartLight()?.setEffect(id)` accepts. A host (the HA bridge) needs the {id, name, colors, buildable} list to populate an effect picker. Account-level HTTP catalogue, not a per-device call; callers should cache it.

## Test
- `tsc --noEmit` clean; `prettier --check` clean; doc-refs / line guards pass locally.


src/client/eufy-mega.ts
```diff
@@ -2000,6 +2004,17 @@ export class EufyMega extends EventEmitter {
...
   }
 
+  /**
+   * Browse the smart-light effect gallery — the catalogue of ids that
+   * `dev.smartLight()?.setEffect(id)` accepts, each with its display name (and preview colours).
+   * An account-level HTTP catalogue, not a per-device call, so it lives on the client; delegates to
+   * {@link listLightEffectsHttp}. Scanning the id band is a handful of `batchget` calls, so a caller
+   * should fetch this once and cache it, not per render.
+   */
+  listLightEffects(opts: { idRange?: [number, number]; ids?: number[] } = {}): Promise<LightEffectSummary[]> {
```
martijnpoppen, line 2014: Two smaller things on the signature itself, independent of whether the method lands:

**The options type is a hand-copy.** `{ idRange?: [number, number]; ids?: number[] }` is retyped here rather than derived from `listLightEffects`'s own parameter, so widening the catalogue scan (a `categories` filter, a `region`) means editing two places and the compiler won't tell you when they drift. Either derive it:

```suggestion
  listLightEffects(opts: Parameters<typeof listLightEffectsHttp>[1] = {}): Promise<LightEffectSummary[]> {
```

…or give `light-catalog.ts` a named exported `LightEffectQuery` interface and use it on both sides — that also gives hosts a type to name when they build a picker.

**`{@link listLightEffectsHttp}` points at the local import alias**, not the exported symbol. The real name is `listLightEffects`; TypeDoc resolves aliases inconsistently, and the alias only exists because of the name collision with this method.

---

martijnpoppen (commented):
The code is fine — `npm run verify` is green on `1d95d44` (182 files, 3432 tests, 277 exports), so it's stronger than the "tsc + prettier" the description claims. But I don't think the method should land as written, because **the premise in "Why" isn't accurate**, and there's a shipped consumer already doing this the other way.

## `listLightEffects` is already reachable from outside the client

> `listLightEffects` (transport/http/light-catalog.ts) took the private `MegaHttpClient`, so nothing outside the client could browse the `lightId`s

Both the function and its result type are already root barrel exports:

```ts
// src/transport/http/index.ts:7
export { listLightEffects, listAiSceneRecommendations, type LightEffectSummary } from "./light-catalog.js";
```

which `transport/index.ts` re-exports via `export * from "./http/index.js"`, and root `index.ts` via `export * from "./transport/index.js"`. And the `MegaHttpClient` it needs is public too — `eufy-mega.ts:1012`:

```ts
/** Raw mega HTTP client, for endpoints not yet wrapped. */
get api(): MegaHttpClient {
  return this.mega;
}
```

So `listLightEffects(eufy.api)` already works today. What was private is the *field*, not the client.

## The Homey app is already doing exactly that, in production

`mega-yfue/homey-eufy-app` consumes this SDK as `"@mega-yfue/eufy-sdk": "file:../eufy-sdk"` and browses the gallery like this — `app.ts:10` and `app.ts:231`:

```ts
import { listLightEffects, type EufyMega, type LightEffectSummary } from '@mega-yfue/eufy-sdk';
…
this.lightEffectsInFlight ??= listLightEffects(eufy.api)
```

Its test suite even pins the shape this PR says doesn't exist — `test/flow.test.ts:540`:

```ts
const sdk = (await import('@mega-yfue/eufy-sdk')) as { listLightEffects?: unknown };
assert.equal(typeof sdk.listLightEffects, 'function', 'listLightEffects must be a barrel export');
```

So the host this is meant to unblock already has what it needs. If the HA bridge can't browse the catalogue, the gap is more likely that it doesn't know about `api` than that the SDK lacks a door.

## …and what the Homey app adds is the part that would actually be worth shipping

The delegator forwards and nothing else. The Homey app wraps the same call in the policy that makes it usable (`app.ts:213-265`): a 1-hour TTL cache, in-flight de-duplication so concurrent callers share one fetch, a failure path that falls back to the last good list instead of throwing into a UI callback, and a `.filter((e) => e.buildable)` so a user can't pick an effect `setEffect` will then refuse. Its own doc says why:

> the catalog is per-account and effectively static, while a Flow autocomplete asks on every keystroke, so refetching each time would hammer the cloud

A bare `listLightEffects()` on the facade gives the HA bridge none of that — it would rebuild the same four things. Either the caching belongs in the SDK (in which case *that's* the PR, and it's a bigger one), or it's host policy and the host should keep it — which is what Homey concluded. A forwarder that adds nothing is the one option that helps neither.

## It also collides head-on with #123

[#123](https://github.com/mega-yfue/eufy-sdk/pull/123) removes six forwarders of exactly this shape from this exact class — `getProductDataPoint`, `getDeviceMapList`, `getManyDeviceMapContent`, `getDeviceParamList`, `megaRequest`, `parseCleanRecordDetail` — on the stated principle that they are *"the facade forwarders the raw client accessor already reaches"*. That accessor is `api`. Both PRs branch from `beta-0.1.0`, so whoever merges second re-opens the question with the other one's reasoning already in `main`. Worth settling the rule once — *does the facade wrap HTTP catalogues, or is `api` the door?* — rather than deciding it twice in opposite directions in the same week.

## Minor, if it does land

- Inline note on the options type being hand-copied instead of derived, and the `{@link}` pointing at a local import alias.
- **The cost note undersells it.** "Scanning the id band is a handful of `batchget` calls" — the default window is `LIGHT_EFFECT_ID_WINDOW = [10001, 10999]`, i.e. 999 ids at `BATCHGET_CHUNK = 100`, so it's **10 sequential `batchget` round trips plus the `discover/list` call, all awaited in series**. "A handful" reads like two or three and invites a per-render call; the Homey app caches it for an hour precisely because it isn't. Say the real number in the doc.

Happy to be wrong if the bridge hits something `api` genuinely can't reach — but as it stands the door is already open, and the neighbouring app walks through it.

---

max246:
Good call — you're right on all counts, closing this.

`listLightEffects` is already a barrel export and `eufy.api` reaches the `MegaHttpClient`, so `listLightEffects(eufy.api)` works today; my "Why" was wrong, and a bare forwarder collides head-on with #123's principle. The HA bridge just didn't know about `api` — now fixed there: it imports `listLightEffects` and calls `listLightEffects(eufy.api, { idRange: [10001, 20999] })` directly (ha-eufy-sdk-bridge 0.1.53), with its own cache + `buildable` filter, matching what the Homey app does. No SDK change needed.

(Also took the cost-note correction — it's ~110 sequential batchget round-trips on a cold cache with the widened window, not "a handful".)

Thanks for the thorough review.

PR #125, opened 2026-09-08
perf: load mqtt and protobufjs on first use — importing this package costs 36 MB, now 24

martijnpoppen:
## What changed

Two dependencies stop loading at import time.

- `transport/mqtt/engine.ts` — `mqtt` loads on first use. Its two call sites (`SecureMqtt.connect`, `probeBrokerInstance`) already returned promises, so the import moved inside them.
- `transport/protobuf.ts` — `protobufjs` loads on first use via `createRequire`, and the push subtree's `.proto` documents parse once on first use (`mcsRoot()` / `checkinRoot()` in `push/proto.ts`) instead of at module scope.

No public API changes. Nothing added to or removed from the barrel.

## Why

A consumer on a memory-limited host profiles at ~125 MB idle against a platform that warns at 150 MB and force-stops around 160. Importing this package is **the largest single block of that before a line of their code runs**.

Measured with one fresh Node process per import (a single process importing several modules in sequence gives meaningless per-module deltas — the second import inherits the first's heap, interned strings and warmed allocator), RSS from `/proc/self/statm` rather than `heapUsed`, because what a memory-limited host kills a process for is resident pages:

| Import | Delta over a 46 MB bare node |
| --- | --- |
| `mqtt` alone | +19 MB |
| `protobufjs` alone | +10 MB |
| `werift` alone | +35 MB — was already lazy before this PR |
| `jpeg-js` alone | +1 MB |
| **the barrel on `beta-0.1.0`** | **+36 MB** |
| **the barrel on this branch** | **+24 MB** |

Both barrel figures re-measured on the current base after merging it in — `beta-0.1.0` has moved since this PR opened, and it costs 3 MB more than when the branch was cut, so the saving is 12 MB rather than the 8 first reported.

Neither dependency is needed to import a client. The secure broker is only reached when a device needs one and a camera-only account never opens one; protobuf is only needed once push registers or a device speaks raw data-points.

**Why `protobufjs` uses `createRequire` and `mqtt` does not.** protobufjs's call sites are *synchronous* — a wire parser driven by socket data, a data-point decoder called from a message handler — and none has an `await` to hide a module load behind. An async loader would have meant either making those paths async, which changes what a consumer depends on, or a load-it-first rule enforced by throwing at runtime in exactly the paths that used to work. A synchronous require defers the cost with neither. mqtt's call sites already returned promises, so it uses `await import()` and stays mockable by `vi.mock`.

**A note for whoever optimises this next.** A dependency's standalone cost is not its marginal one: `mqtt` measures +19 alone, but removing it takes far less than that off the barrel, because most of what it loads something else loads too. Only the barrel number is the number.

## Evidence

n/a — no wire behaviour changes. This is packaging and load order.

The measurement is a `node -e` that reads `/proc/self/statm` before and after a dynamic import, run once per module in its own process. The script that produced the table deliberately does **not** ship here — tooling belongs in the workshop, not the library.

## Verified how

- [x] `npm run verify` is green locally on the merged head — 3364 tests
- [x] Ran against a real device — n/a, no behaviour on the wire changed
- [x] Added or updated specs — six `SecureMqtt`/broker-probe tests now await the client instead of reading it synchronously (see below)
- [x] Docs updated — `engine.ts` and `protobuf.ts` each carry the measurement and the reasoning in their module doc

Six mqtt tests reached for the mocked client synchronously right after `connect()`. With the engine lazy it appears a few microtasks later, so they `vi.waitFor` it. **Nothing either method settles with changed** — same resolution, same rejection, same `end(true)` cleanup assertions.

`beta-0.1.0` merged in to resolve a conflict in `push/fcm.ts`: the base added a `node:timers/promises` import beside the `protobufjs` one this branch replaces. Resolved by keeping both changes. Checked afterwards that no eager `protobufjs` or `mqtt` import survived the merge anywhere, and that `noUnusedLocals`/`noUnusedParameters` (turned on by the base) are satisfied.

## Review pass

Three cuts from @lenoxys, net −58 lines, all taken.

- **`engine.ts` was 36 lines for a three-line idiom** — now 19, nearly all module doc: `??=` rather than `if (!x)`, the promise returned directly rather than `async` + `await`, and `typeof import("mqtt")` in place of the `import type` line.
- **`protobuf.ts` kept a `Map` keyed by the full `.proto` source** to cache two roots, hashing a multi-KB template literal on every parser construction. The roots now live in `push/proto.ts` beside their documents, so both MCS consumers share one naturally and the key — plus the JSDoc that existed only to defend it — is gone. `proto.ts` still pulls no `protobufjs` at import; the built `proto.js` imports only `../protobuf.js`, whose require sits inside a function body.
- **`nextClient` was 17 lines duplicated across two sibling specs** — hoisted to `__tests__/lazy-engine.ts`, taking the clients array, since `vi.mock` hoists per file and each spec keeps its own factory.

Also cut, and worth naming because it was the judgement call rather than slop: the loader's **rejection-eviction branch**. A dynamic import of a declared dependency rejects because it is missing or because module eval throws, and a retry after either fails identically; the case eviction would cover is a transient resource limit mid-load, which is real but speculative. Five lines to restore with evidence if one ever bites.

One correction to the review, since it was load-bearing for the ask: the cited precedent `webrtc/peer.ts` **no longer exists** — `a1106a5` deleted the whole `src/transport/webrtc/` tree and it is on `beta-0.1.0` as of `5dce6e3`. The idiom is right regardless, so it was taken anyway.

And an out-of-scope catch: `engine.ts` said `+33 → +25` while this body says `+36 → +24`. The module doc is the copy a consumer reads, so it now carries the re-measured figures.

## Unverified or deferred

**The remaining +24 MB is this package's own code**, not dependencies: `model/capabilities` (32 modules) and `transport/p2p/command-router` plus what it pulls are roughly half each. Capability *detection* needs every module loaded — you cannot know which apply until each has been asked — so making that lazy means splitting all 32 into a light descriptor and a heavy body. A real refactor, not a corner, and not attempted here.

**Nothing holds the +24 MB in place.** An eager `import` added to a transport module later puts it back and no test or guard notices — as the base moving from +33 to +36 while this PR was open illustrates. A `guard:import-cost` with a ceiling, folded into `verify`, would fix that; it is not in this PR because the threshold needs a tolerance that survives a different Node version and a different machine.

**Bundling is the bigger remaining lever and is deliberately not here.** Measured with esbuild: single-file ESM is **+15 MB**, minified **+13 MB** — larger than this change. It is a packaging decision rather than an internal one, so it should be its own PR with its own discussion. In its favour: `exports` has exactly one entry point, so there is no way for a consumer to load two copies and end up with two sets of the module-level caches this PR introduces. Against: bundled stack traces, and a build step between `tsc` and `dist`. Declarations are untouched either way, so `guard:seal` is unaffected.


src/transport/mqtt/engine.ts
```diff
@@ -0,0 +1,36 @@
...
+      (e) => {
+        // Do not leave a rejected promise memoized: a transient failure would then be permanent for
+        // the life of the process, and the next connect attempt deserves a real one.
+        enginePromise = undefined;
+        throw e;
+      },
+    );
+  }
+  return await enginePromise;
+}
```
lenoxys: **yagni:** this pattern is already in the repo. `webrtc/peer.ts` L33-40 lazy-loads werift for the same reason, in three lines:

```ts
type Werift = typeof import("werift");
let weriftPromise: Promise<Werift> | undefined;
function loadWerift(): Promise<Werift> { return (weriftPromise ??= import("werift")); }
```

Same shape works here verbatim — and `typeof import("mqtt")` drops the `import type mqtt` line at L16 too.

The three things this version adds over that:

- **`if (!enginePromise)` instead of `??=`** — no difference, just longer.
- **`async` + `return await enginePromise`** — an extra microtask to return a promise a caller awaits anyway. `peer.ts` returns it directly.
- **the rejection-eviction branch (L27-32)** — the one genuine difference, and the one I'd push back on. A dynamic import of a declared dependency rejects because it is missing or because module eval throws; neither is transient, so "the next connect attempt deserves a real one" is describing a retry that fails identically. `peer.ts` doesn't do it and nobody has been bitten.

If you're convinced the eviction earns its keep, then it earns it for werift too — and the answer is one `lazyModule()` helper both call, not two divergent implementations of a three-line idiom. What I'd rather not merge is a new 36-line module that is the existing pattern at 5x the size. `AGENTS.md` "Reuse before abstraction" points at this directly: search by operation and data shape, not by proposed name.

The module doc is worth keeping either way — the measurement belongs in the source.

lenoxys (commented):
**Over-engineering pass only** — the change itself is right and the measurement discipline (one fresh process, RSS not `heapUsed`) is the good kind of rigour. Three cuts, all in the scaffolding.

| where | tag | net |
|---|---|---|
| `engine.ts` L16-36 | yagni | ~-18 |
| `protobuf.ts` L36-51 | yagni | ~-8 |
| both mqtt specs | shrink | -17 |

`net: ~-40 lines possible.`

Out of scope but cheap: `engine.ts` L6 says **+33 MB → +25**, the PR body says **+36 → +24**. The body explains why the base moved; the shipped JSDoc is the copy a consumer reads.

src/transport/protobuf.ts
```diff
@@ -0,0 +1,51 @@
...
+ */
+const roots = new Map<string, Root>();
+export function protoRoot(document: string): Root {
+  let root = roots.get(document);
+  if (!root) {
+    root = protobufjs().parse(document).root;
+    roots.set(document, root);
+  }
+  return root;
+}
```
lenoxys: **yagni:** a `Map<string, Root>` keyed by the full `.proto` source, for two documents that are both module constants in `push/proto.ts`. Every `McsParser`/`PushClient` construction now hashes a multi-KB template literal to pick one of two entries.

The sharing it buys is real — `parser.ts` and `push-client.ts` both want the *same* MCS root, so a per-file `??=` would parse it twice. But that's an argument for one lazy root per document, not for a keyed cache. Put them where the documents live:

```ts
// push/proto.ts
let mcs: Root | undefined;
export const mcsRoot = () => (mcs ??= protobufjs().parse(MCS_PROTO).root);
```

Same for `CHECKIN_PROTO`. Both MCS consumers share it naturally, the string key and the `Map` go, and the JSDoc at L39-41 defending the choice of key stops needing to exist — which is the tell. `proto.ts` stays free of `protobufjs` at import; it's still only a function body.

`protobufjs()` itself stays as-is, `raw-dp.ts` needs the engine rather than a root. **shrink** while you're in here: L21-22 are two `import type` lines from `"protobufjs"` — merge, and `cached ??= …; return cached;` is `return (cached ??= …)`.

The `createRequire` decision is correct and well argued — synchronous call sites, no `await` to hide behind. No quarrel there.

src/transport/mqtt/__tests__/secure-mqtt-connect.spec.ts
```diff
@@ -32,6 +32,23 @@ const CREDS = {
...
+ * is unchanged.
+ */
+async function nextClient(index = 0) {
+  return await vi.waitFor(() => {
+    const client = fakeClients[index];
+    if (!client) throw new Error(`no mqtt client #${index} yet`);
+    return client;
+  });
+}
+
```
lenoxys: **shrink:** byte-identical to `broker-instance-probe.spec.ts` L46-62, JSDoc included — 17 lines twice, in sibling files, for six call sites.

Hoisting the whole `vi.mock("mqtt")` factory into a shared fixture fights vitest's hoisting, so I wouldn't ask for that. But `nextClient` is a plain async function with no mock coupling beyond the `fakeClients` array — take it a level up and pass the array, or at minimum let the second copy be one line pointing at the first instead of a second full essay on why the client is late.

The change to the tests is the right one, and the assertion that nothing about what either method settles with moved is the part that matters.

src/transport/mqtt/engine.ts
martijnpoppen: Taken — `engine.ts` is 19 lines now, all but the module doc being the three-line idiom: `??=`, the promise returned directly, and `typeof import("mqtt")` in place of the `import type` line.

One correction to the premise, which does not change the conclusion: **`webrtc/peer.ts` no longer exists.** `a1106a5` ("carry only what is read, derive what was stated twice") deleted the whole `src/transport/webrtc/` tree, and it is on `beta-0.1.0` as of `5dce6e3` — so there is no second implementation to diverge from and no shared `lazyModule()` to extract. Flagging it because "the pattern is already in the repo" was load-bearing for the ask, and it is not true on the current base.

The idiom is still right, so I took it anyway.

**The eviction branch I have also cut, and I want to be straight that this was the judgement call rather than the slop.** Your argument is sound about the realistic failures: a declared dependency rejects because it is missing or because module eval throws, and a retry after either fails identically. The case it would actually cover is a transient resource limit mid-load — real, but speculative, and speculative resilience is exactly what this PR is otherwise removing. Five lines to restore with evidence if one ever bites.

---

src/transport/protobuf.ts
martijnpoppen: Done exactly as suggested — `mcsRoot()` and `checkinRoot()` now live in `push/proto.ts` beside the documents they parse, and `protobuf.ts` is down to the engine plus its doc.

The tell you pointed at was the right one: the JSDoc existed only to defend keying a `Map` by a multi-KB template literal, and once the roots sit beside their documents there is no key to justify. Both MCS consumers share `mcsRoot()` naturally, so the document is still parsed once.

Also confirmed the thing that made this safe — `proto.ts` pulls no `protobufjs` at import: the built `dist/transport/push/proto.js` imports only `../protobuf.js`, whose own require is inside a function body. And took the two shrinks: one merged `import type`, and `return (cached ??= …)`.

---

src/transport/mqtt/__tests__/secure-mqtt-connect.spec.ts
martijnpoppen: Hoisted to `__tests__/lazy-engine.ts`, taking the clients array as a parameter — which is the version that works, since `vi.mock` hoists per file and each spec has to keep its own factory and its own array. The docstring lives there once; both specs are down to an import.

---

lenoxys (approved):
Re-checked at `71ea9ae`. All three cuts taken, plus the stale figures in the shipped JSDoc.

`loadMqtt` is now the same three-line idiom as `peer.ts` — and keeping `.then((m) => m.default ?? m)` there is correct, mqtt's shape needs it where werift's named exports don't. The `as Mqtt` cast turning out to be unnecessary once the branch went is a nice tell. Lazy roots beside their documents in `push/proto.ts` is the better home for the reason you give: both MCS consumers share one parse, and `proto.ts` still pulls no `protobufjs` at import.

`__tests__/lazy-engine.ts` follows the existing `live-source-fixtures.ts` convention and `tsconfig.build.json` excludes `**/__tests__/**`, so it can't leak `vitest` into `dist` — worth confirming on a PR about import cost, and it doesn't.

`tsc -p tsconfig.json --noEmit` clean. Both mqtt specs plus the push subtree green, 15 tests.

**One nit, not blocking.** With `mcsRoot()` memoizing itself, `McsParser.root` and `PushClient.root` are private static getters whose whole body forwards to it — one call site in `parser.ts`, three in `push-client.ts`. Calling `mcsRoot()` directly at those four sites drops both getters and the `Root` type import with them, ~-8. They earned their keep as static *fields* (parse once); as getters over a memoized function they're pure indirection. Fine to leave or fold into a later pass.

The `createRequire` reasoning and the measurement method are the parts of this PR I'd point someone else at.

martijnpoppen:
Took the nit — `cd1f115`. You are right that the getters stopped earning their keep the moment the parse moved into a memoized function: as static *fields* they bought one parse, as getters forwarding to `mcsRoot()` they bought a layer. Four call sites call it directly now, and the `Root` type import goes with the two signatures that needed it. −9 lines.

Flagging the push since it lands after your approval: it is the nit and nothing else — no behaviour, no other file. `npm run verify` green (3364), and the `docs/` build too, which `verify` does not cover.

Thanks for checking `tsconfig.build.json` excludes `**/__tests__/**`. That was worth confirming on this PR specifically and I had not thought to state it — a test helper leaking `vitest` into `dist` would have been a genuinely funny way to lose an import-cost win.

---

PR #126, opened 2026-09-08
feat(p2p): coded frame geometry, and let a caller open one station without warming a fleet

martijnpoppen:
## What changed

Three additive changes, all found while building a still-image path on top of this SDK.

- **`codedGeometry` / `extractParamSets` carry the coded size and crop window**, not just the display size. `CodedGeometry` gains `coded: Size` and `crop: {left, top, right, bottom}`.
- **`EufyMega.connectStation(deviceSn)`** — open the P2P session for one device's station and nothing else.
- **`noBroadcast`** on `EufyMegaOptions` — suppress the `255.255.255.255` local-lookup broadcast.

No breaking changes; nothing removed.

## Why

**Coded geometry.** A consumer decoding a keyframe in-process gets back the *coded* frame, not the display one — 1080 is coded as 1088, because macroblocks are 16 pixels tall and 1080 is not a multiple of 16. The parameter sets already say exactly how much to crop, and this SDK was parsing that and throwing it away, so every caller either shipped 8 rows of padding or re-parsed the SPS itself. The information was already in the bytes being read; only the return type had to grow.

**`connectStation`.** Auto-realtime warms a P2P session to *every wired station on the account*, which is right for a host driving a fleet and wrong for anything else. The cost is not abstract: an unreachable station broadcasts a local lookup once a second for the whole 15 s connect timeout and sends a PPCS lookup to every cloud address in the same tick. A caller that needs one station therefore pays fifteen broadcasts and a cloud-lookup burst per station it never asked about, on every run. Paired with `autoRealtime: false`, this opens exactly what is used.

**`noBroadcast`.** `P2PSessionCfg` has had this field since the local lookup was written and **nothing could ever set it** — the router never passed it and the client never offered it. Plumbed through the same seam `localAddresses` uses. Broadcast frames reach every host on the segment and are transmitted to every associated client of a WLAN at the lowest basic rate. Suppressing it costs the LAN-discovery path only — a station whose record carries a usable `ip_addr` is still reached directly, and the PPCS cloud lookup that connects most topologies is untouched.

Default is unchanged (`false`): broadcast still happens unless a caller opts out.

## Evidence

**Coded geometry** — H.264/H.265 parameter sets. The coded size is `(pic_width_in_mbs_minus1 + 1) * 16` by `(pic_height_in_map_units_minus1 + 1) * 16` adjusted for frame/field, and `frame_cropping_flag` carries the window; this reads what the SPS already declares rather than inferring anything. Confirmed against a real 1080p stream, which reports display 1920×1080 and coded 1920×1088.

**`connectStation` / `noBroadcast`** — no new wire. `connectStation` calls the existing `ensureStation` with the existing `stationKeyOf` mapping; `noBroadcast` gates a send that was already there. The behaviour they change is *how much* this library does unasked, and that is read off the existing code: `LOOKUP_RETRY_MS = 1000`, `CONNECT_TIMEOUT_MS = 15000`, one broadcast plus one datagram per cloud address per tick, per unconnected session, times every wired station.

## Verified how

- [x] `npm run verify` is green locally — 3432 tests
- [x] Ran against a real device — geometry confirmed on a 1080p H.264 camera behind a HomeBase (display 1920×1080, coded 1920×1088, crop applied); `connectStation` used to drive a single station on an account whose other two time out, which is what surfaced the fleet-warm cost
- [x] Added or updated specs — the existing coded/crop cases now assert both sizes exactly
- [x] Docs updated — JSDoc on each of the three

Also ran the `docs/` build (typedoc → guard:docs → vitepress), which `npm run verify` does not cover: an unresolvable `{@link}` passes every local gate and only fails in CI. The first push here did exactly that.

## Review pass

Four cuts from @lenoxys, net −46 lines. One was a real regression this PR introduced and is worth naming here rather than leaving in a thread: **adding `coded`/`crop` broke twelve `toEqual` assertions in the geometry spec, and they were made to pass by weakening them to `toMatchObject` — then a new block re-asserted five of the same fixtures exactly.** The originals ended up strictly weaker than before this PR *and* duplicated. The two sizes are now folded into those five originals, which are back on `toEqual`; the duplicate block is gone; the eight cases where the coded size is incidental keep `toMatchObject`.

The other three were slack: an unreachable plausibility gate (`ue()` is unsigned, so a display size can never exceed the coded one), a `SO_BROADCAST` guard that changed nothing observable because the send is gated on the same flag, and two consecutive exports from the same module.

## Unverified or deferred

Nothing here is a write path — the two P2P changes only reduce what the library does on its own initiative, and the geometry change is a parse of bytes already being parsed.

`connectStation` is best paired with `autoRealtime: false`; used with realtime on, the fleet warm still runs and this is then a no-op on an already-open session. That is documented on the method rather than enforced, because a host may legitimately want both.


src/transport/p2p/__tests__/annexb-coded-geometry.spec.ts
```diff
@@ -162,9 +162,67 @@ describe("codedGeometry — what it refuses to answer", () => {
...
+  it("reads an H.265 conformance window as the same two sizes", () => {
+    const sps = h265Sps({ widthLuma: 1920, heightLuma: 1088, window: { bottom: 4 } });
+    expect(codedGeometry(setsOf(H265.vps, sps, H265.pps))).toEqual({
+      width: 1920,
+      height: 1080,
+      coded: { width: 1920, height: 1088 },
+      crop: { left: 0, right: 0, top: 0, bottom: 8 },
+    });
+  });
+});
```
lenoxys: **delete:** all five fixtures here are byte-identical to cases that already exist above — `(120,68,bottom 4)` is L25, `(80,45)` is L18, `(80,45,profile 100,left 1,right 2)` is L41, `(80,22,!frameMbsOnly,bottom 1)` is L57, `h265 1920x1088 window bottom 4` is L94. Same input, same function, asserted twice.

And the duplication paid for itself twice over: twelve assertions above got downgraded from `toEqual` to `toMatchObject` to keep them passing, so the originals are now strictly weaker than they were before this PR *and* redundant with the new block. That's a net loss of coverage from a diff that adds tests.

Fold `coded` and `crop` into the five originals, keep them on `toEqual`, drop this block. The seven cases where the coded size is genuinely incidental (scaling lists, chroma-branch sweep, ladder rungs) can stay `toMatchObject` — that one's a real judgement call, not slop.

src/transport/p2p/annexb.ts
```diff
@@ -300,15 +321,19 @@ export function codedGeometry(sets: ParamSets): CodedGeometry | undefined {
...
     width > 0 &&
     height > 0 &&
     width <= MAX_CODED_DIMENSION &&
-    height <= MAX_CODED_DIMENSION;
-  return plausible ? { width, height } : undefined;
+    height <= MAX_CODED_DIMENSION &&
+    // A display size larger than the size it is coded at is not a crop, it is a misparse — and it is the
+    // reading a cropping caller would act on by indexing past the end of its decoder's own buffer.
+    coded.width >= width &&
+    coded.height >= height;
```
lenoxys: **delete:** unreachable. `cropWindow` reads all four offsets with `r.ue()`, which is unsigned exp-Golomb and returns `0` on failure — never negative. `width` is `coded.width - left - right`, so it cannot exceed `coded.width` for any input, corrupt bytes included. The gate can't fire.

The misparse it's guarding against is real, but it lands as a *plausible wrong number* — which is what the `MAX_CODED_DIMENSION` and `width > 0` checks above are already for, and which the existing JSDoc on `ue()` says outright.

Also an inline body comment, which `AGENTS.md` rules out — moot if the branch goes.

src/transport/p2p/p2p-session.ts
```diff
@@ -638,10 +638,15 @@ export class P2PSession extends EventEmitter {
...
+        // Only when the broadcast lookup is actually going to be sent. `SO_BROADCAST` is a socket
+        // option nothing else here needs, and a session configured not to broadcast should not be
+        // asking the host for it.
+        if (!this.cfg.noBroadcast) {
+          try {
+            socket.setBroadcast(true);
+          } catch {
+            /* broadcast not permitted — cloud path still works */
+          }
         }
```
lenoxys: **delete:** the send is already gated on the same flag at L743, so with `noBroadcast` set this socket never broadcasts regardless. `SO_BROADCAST` on a socket that doesn't broadcast is a no-op at the kernel — nine lines and a nesting level for zero observable difference.

The PR body frames it as "the socket no longer requests `SO_BROADCAST` at all when suppressed", which is true and isn't a behaviour. Restore the bare four-line try/catch.

src/transport/p2p/index.ts
```diff
@@ -14,6 +14,10 @@ export {
 } from "./shared-live-source.js";
 // `ParamSets` is what `SharedLiveSource.parameterSets` answers with, so a caller has to be able to name it.
 export { type ParamSets } from "./annexb.js";
+// A consumer that DECODES frames itself (rather than muxing them) gets the macroblock-aligned coded size
+// back from its decoder and has to crop to the display size the stream declares — which is what these two
+// answer. Published together because the geometry is unreachable without the sets it reads.
+export { extractParamSets, codedGeometry, type CodedGeometry, type Size } from "./annexb.js";
```
lenoxys: **shrink:** second `export … from "./annexb.js"` directly under the first. Merge into L16 and keep one comment block.

lenoxys (commented):
**Over-engineering pass only** — no correctness or perf claims here, and nothing below is about whether the feature is right. Four cuts, all mechanical.

| where | tag | net |
|---|---|---|
| `annexb-coded-geometry.spec.ts` L172-228 | delete | ~-50 |
| `annexb.ts` L332-335 | delete | -4 |
| `p2p-session.ts` L641-650 | delete | -5 |
| `index.ts` L17-20 | shrink | -1 |

`net: ~-60 lines possible.`

The three additive features themselves are lean — `connectStation` is four lines on existing seams and `noBroadcast` activates a field that was already dead. It's the scaffolding around them that has slack.

src/transport/p2p/__tests__/annexb-coded-geometry.spec.ts
martijnpoppen: Done, and you were right that this one is not cosmetic. I checked the diff against the base and it is exactly as you describe: adding `coded`/`crop` broke twelve `toEqual`s, they were made to pass by weakening them to `toMatchObject`, and then five of the same fixtures were re-asserted in a new block. A diff that adds tests and loses coverage.

Folded the two sizes into the five originals, restored them to `toEqual`, dropped the block. The eight where the coded size is incidental — ladder rungs, chroma-branch sweep, scaling lists, the truncation and multi-SPS cases — keep `toMatchObject`, since pinning a coded size there would assert something the case is not about.

Net −53 lines in this file, and five fewer tests that were never testing anything the others did not.

---

src/transport/p2p/annexb.ts
martijnpoppen: Confirmed and removed. `ue()` returns `zeros === 0 ? 0 : 2 ** zeros - 1 + this.u(zeros)` and sets `failed` with a `0` return on overrun, so every offset is non-negative; `coded.width - left - right` is therefore never greater than `coded.width`, and a negative result is already caught by `width > 0` above. The branch could not fire on any input.

Both the check and its inline comment are gone.

---

src/transport/p2p/p2p-session.ts
martijnpoppen: Restored to the bare try/catch. You are right on both counts — the send at L743 is gated on the same flag, so the option changes nothing observable, and the PR body was selling a no-op as a behaviour. I have corrected that sentence in the body rather than leaving it to stand.

---

src/transport/p2p/index.ts
martijnpoppen: Merged into one export with one comment block.

---

lenoxys (approved):
Re-checked at `98fe1bc`. All four cuts taken, and the spec one taken the right way round — the five originals back on `toEqual` carrying both sizes, the eight where the coded size is incidental left on `toMatchObject`. That distinction was the part worth getting right and you got it.

`coded` also dropped from the destructure when the gate went, so no unused local trailing behind. `tsc -p tsconfig.json --noEmit` clean; `annexb-coded-geometry` + `fmp4-geometry` green, 27 tests.

Nothing left on the over-engineering axis. The three features were lean to begin with — it was only the scaffolding.

PR #127, opened 2026-09-08
fix(rtsp): restore the P2P device-URL read (reportedRtspUrl facade)

martijnpoppen:
## What changed

Restore the P2P read that fetches a camera's live RTSP URL, which #114 removed when it moved the URL to inbound capability state (`decodeState`). Both now decode the same 1145 push; the read is the one that actually delivers it on HomeBase-attached cameras.

- **`p2p-session`** — re-emit `rtspUrl` for a `CMD_NAS_SWITCH` (1145) push whose payload is an `rtsp://` string.
- **`command-router`** — `readReportedRtspUrl` + `awaitRtspUrl`: arm a channel-filtered listener, provoke with 1145 + 1146 (one loop, through the one `"auto"` level decision), await the push, bounded by a single deadline; `undefined` on no route / no reply.
- **`eufy-mega`** — `reportedRtspUrl(sn)` facade over it, documented as the canonical fetch.
- **`rtsp.ts`** — `decodeState`/`url` stay; the `url` member is documented as inbound state for `propertyChanged` consumers, not a second fetch path.

## Why

The URL is P2P-notify only and, on HomeBase-attached cameras, is **not** in the cloud record and **not** pushed by the publish switch alone (both confirmed on hardware — a trace showed 1145/1146 returning no `rtsp://` payload without the read's exact sequence). So there is no state to decode and nothing to mine: the only source is the P2P round-trip that provokes with 1145 + 1146 and reads the push back. Writes only the publish switch and the test-stream provoke, never `SEND_SECURITY_PASSWD`, so a stream a NAS/NVR already publishes keeps its own credentials.

## Evidence

Verified end to end against a HomeBase-attached camera (T8410 behind a T8010): a consumer calls `reportedRtspUrl(sn)`, the URL reads back over P2P, and live video plays. The credentials in the returned URL are the device's own — the read never sets them.

## Verified how

- [x] `npm run verify` is green locally (180 files, 3370 tests)
- [x] Ran against a real device — a HomeBase-attached camera; URL read over P2P, live video played
- [x] Added or updated specs — `read-reported-rtsp-url.spec.ts`: the channel-filter safety case (a push for another camera on the shared session must not resolve this read), the level-1/level-2 provoke wires, the timeout leaving no listener, and the no-account / no-session guards
- [x] Docs updated — `url` member + `decodeState`, and `reportedRtspUrl` as the canonical fetch

## Known limitations (follow-ups, not blockers)

- **Failure modes collapse to `undefined`.** No route / no account id / level-2-not-ready / no-push-before-deadline all return `undefined`; a caller that retries recovers from the transient ones. A typed terminal-vs-retryable reason (like `snapshotLive`'s `retryable`) is a worthwhile refinement but changes the facade contract and its consumer, so it is deferred rather than bundled here.
- **`awaitRtspUrl` is a third hand-rolled reply-wait.** Factoring the shared promise/listener/timeout skeleton across `p2pQuery` / `p2pControlQuery` / `awaitRtspUrl` (or adopting `events.on`) is a transport-wide refactor best done on its own, not folded into this fix.

https://claude.ai/code/session_019s1WfLUuBXxvyJ6rDanUP1


src/model/capabilities/rtsp.ts
```diff
@@ -226,6 +242,28 @@ export const RTSP_MEMBERS = {
...
+   * a NAS/NVR is already consuming: it only nudges the station to state the URL, without withdrawing or
+   * re-publishing. Sent form:auto — a keyed HomeBase serves 1145 on its level-2 seal and drops the
+   * level-1 form, so the provoke must ride the same level the publish does.
+   */
+  requestUrl: method(
+    ({ ctx, sink }) =>
+      (): Promise<void> =>
+        sink.dispatch(setScalar(RTSP_PARAM.STREAM_TEST, 1, ctx)),
+    "Ask the device to report its RTSP URL (starts the stream); read dev.rtsp()?.url after.",
+  ),
```
lenoxys: See the main comment for the shape argument — this is the member it's about.

Two narrower things:

**`answers: true` is missing either way.** Even keeping the current design, this is not a control worth offering: a button that fires a provoke and discards the outcome is exactly what `lock.getAutoLockState` documents `answers` to prevent. As written it lands in the described-controls surface as an action a host could render.

**The description promises what the method doesn't deliver:** *"Ask the device to report its RTSP URL (starts the stream); read dev.rtsp()?.url after."* A described control whose effect a caller has to go read somewhere else, on a poll, with no failure signal, is the thing `AGENTS.md` means by a described write being one a caller can actually offer.

src/model/capabilities/rtsp.ts
```diff
@@ -13,10 +13,25 @@ export const RTSP_PARAM = {
...
+   * persistent setting, but the station only reports the URL once a consumer actually asks the stream to
+   * start, which is this command. So {@link requestUrl} sends this to provoke the push that
+   * {@link RTSP.decodeState} then lifts into {@link STREAM_URL}.
+   *
+   * Distinct from the publish switch on purpose: this does NOT touch the persistent 1145 setting or the
+   * credentials, so sending it to a camera a NAS/NVR is already consuming does not withdraw or
+   * re-publish anything — it only nudges the station to (re)state the URL. `writable:false` in the param
+   * dictionary because it is a provoke, not a stored setting.
+   */
+  STREAM_TEST: 1146,
```
lenoxys: **`writable:false` in the param dictionary** — there's no such thing to set here. `RTSP_PARAM` is a map of plain numbers with no per-entry flags, and `writable` in the published schema is derived in `manifest.ts` from whether a setter function exists. Since `1146` ships as a `method`, it has no schema entry at all, so nothing is being declared `false`. Drop the sentence or state the real mechanism.

**shrink:** this block and `requestUrl`'s JSDoc are two explanations of one thing — both cover what provokes the push, and both carry the same "does NOT touch the publish switch or the credentials, so a NAS/NVR already consuming is undisturbed" paragraph. Keep the wire-level reasoning on the constant, the caller-level contract on the member, and let one `{@link}` do the joining. ~-8.

src/model/capabilities/__tests__/rtsp.spec.ts
```diff
@@ -67,6 +67,19 @@ describe("rtsp capability module", () => {
...
+  it("provokes the URL push with the test-stream command (1146), not the publish switch", async () => {
+    const { acts: actions, sent } = bind<RtspActions>("rtsp", ctx);
+
+    await actions.requestUrl?.();
+
+    expect(sent).toHaveLength(1);
+    expect(sent[0].kind).toBe("set-param");
+    // 1146, its OWN command — not the publish switch (1145), which does not elicit the URL on its own.
+    expect((sent[0] as { param: number }).param).toBe(RTSP_PARAM.STREAM_TEST);
+    expect((sent[0] as { param: number }).param).toBe(1146);
```
lenoxys: **shrink:** the first assertion is a tautology — the code under test uses `RTSP_PARAM.STREAM_TEST`, so comparing the sent param to that constant can't fail independently. The literal `1146` is the one doing work: it pins the wire number against a careless edit to the constant. Keep that line, drop the other. -1

lenoxys (changes requested):
You asked whether this is the best way to implement it. **The fix is right and the diagnosis is excellent — but the shape isn't, and this repo already has the shape it wants, one capability over.**

What's not in dispute: the field test is the good kind of evidence, `1145` alone not eliciting the push is a real finding, and pairing `1146` matches the only capture that ever saw the URL. Keep all of that.

Two things I'd change, one of them blocking.

**1. `requestUrl(): Promise<void>` + "poll it briefly" is the wrong contract (blocking-grade, but see #2 — they share a fix).** The JSDoc tells every caller to send a command and then poll a property for an answer that may never come, with no deadline and no way to tell "not yet" from "never". Every consumer will write that loop, each slightly differently, and most will get the timeout wrong. `AGENTS.md` is explicit that the SDK returns typed data or a typed reason and that presentation is the caller's — a poll loop is neither.

`lock.getAutoLockState` is exactly this problem already solved: a genuine request/reply query, done as an injected provider round-trip, `answers: true`, returning the snapshot. Its own JSDoc even says why — "not a passive property, so it always talks to the device". The URL is the same animal: send `1146`, await the `1145` string push against a deadline, resolve with the URL or a typed reason. That's the `PULL something mid-read or mid-action` case the injected-provider rule names.

#114 was right to delete a transport **facade** that silently re-published `1145`. A provider is not that — it's the sanctioned inversion, and it's how the layer boundary stays intact while the capability still gets to ask a question and hear the answer.

**2. The wire is unconfirmed, and `method()` is the one member kind that can't say so.** Your own "Unverified or deferred" section says it: that `1146` after publish actually makes the URL arrive is *not yet* re-confirmed on hardware. The `unverified` flag exists for precisely this and its JSDoc is almost a description of this PR — *a fire-and-forget write that is wrong looks exactly like success, so it must not ship as a callable method*. But `unverified` lives on `ValueMember`; `MethodMember` has no such field. So shipping the provoke as `method()` routes an unconfirmed write around the gate the repo built for it, and `1146` is fire-and-forget with no ack: if it's the wrong lever, `requestUrl()` resolves cleanly forever and the URL simply never appears.

Either of these unblocks it:

- **Re-drive it on hardware.** You confirmed the negative; confirming the positive is the same session. Then it's verified and this objection evaporates.
- **Or declare it as a member with `write` + `unverified: true`** so the setter isn't installed and the intent path throws the real reason, and promote it with one edit when the capture lands.

What I'd not merge is an unconfirmed fire-and-forget provoke as a present, callable, `void`-returning method whose docs ask the caller to poll.

The smaller stuff is inline. Nothing here is about line count — the diff is already small.

lenoxys:
Followed up on my own objection by inventorying every comparable member in the SDK, because "use a round-trip instead" is worth nothing if the codebase doesn't already agree. It does — and more usefully, **#114's own JSDoc already named this exact outcome and its remedy.**

## The paragraph this PR rewrites

The doc at `rtsp.ts:170-175` that you're replacing said:

> The PROVOKE — that `publish` (1145=1) ALONE elicits that push — is INFERRED … **If it turns out not to, the URL never populates and the answer is a `MediaProvider` pull, not this read.**

Your field test is precisely the condition that sentence was written against. #114 predicted the failure and pre-committed to the remedy; this PR hits the failure and does something else. That's the whole of my objection, and it was in the file before I got there.

## What the SDK already does — 11 members, one convention

Request/reply that returns the value, closest first:

| where | what it does |
|---|---|
| `doorbell.ts:340` `quickResponses` | `provided("media", m => m.p2pQuery && …)` — capability owns the sub-command id and reply parsing, transport owns the wire. Structurally the template for a URL read. |
| `ptz.ts:434` `preset().image` | Returns `undefined` as the **typed reason** ("reply carried no image data"), while a reply that never arrives rejects. That distinction is exactly the one `requestUrl` currently can't express. |
| `ptz.ts:434` `preset().list` | `p2pControlQuery`, correlated notify, `{ timeoutMs }`. |
| `lock.ts:310` `getAutoLockState` | Provider round-trip, `answers: true`, throws on no reply. |
| `camera.ts:537-597` | Seven `MediaProvider` reads, each with its own typed failure. |

And the machinery is already generic and already public: `MediaProvider.p2pQuery` (`contracts.ts:861` → `p2p/command-router.ts:1010`) and `p2pControlQuery` (`contracts.ts:869` → `:1049`). Both arm the listener before sending, correlate by the reply's inner `cmd`, default to 15 s, take `{ timeoutMs }`, and clean up on timeout *and* on a synchronous send failure.

## Where I'd temper that

Two honest caveats, because the fit isn't free:

- **The 1145 envelope is different.** Your own `decodeState` doc says the URL comes back as a bare NUL-terminated string, not the `params` array — whereas `p2pQuery` matches `f.json.cmd` and resolves `payload`. So this may not be a drop-in call to either helper. If it needs a third reply-wait, that's the moment to factor the skeleton out: `p2pQuery` and `p2pControlQuery` are already two hand-rolled copies of one promise/listener/timeout shape, and P2P has no equivalent of MQTT's `waitForMqttMessage` (`mqtt/command-router.ts:426`). `contracts.ts:995` is explicit that the answer is a method on an existing provider, not a third one.
- **26 of 38 members do return `Promise<void>`.** So the majority is nominally on your side. But every one of them is a *control* invoked for its effect — lock, rotate, publish, start cleaning. None is a read dressed as a write.

## Shape A exists exactly once — and it argues against this

`smart-light.ts:399` `refreshState`:

> A request, not a state: reports are pushed on change with no periodic heartbeat, so this is how a caller populates the getters on demand. **The reply arrives asynchronously and does not resolve here.**

Three things about it. It's an `ActionMember`, not a `method()`, so it doesn't ship as a described control. Its docs never tell anyone to poll. And `SMART_LIGHT.realtimeInit()` (`smart-light.ts:476`) fires the same command automatically at connect *"so the typed getters are populated at connect rather than only after the first write"* — the caller-driven provoke is the fallback, not the interface. If `requestUrl` follows that precedent properly, the SDK should be sending 1146 itself after publish, not documenting a poll.

## The clincher

The SDK's stated position on writes whose landing needs confirming is the exact opposite of a caller poll: it does the bounded readback **itself** via `observation` / `refreshEventState` (`eufy-mega.ts:729`) and throws `StateConvergenceError`. There is no member anywhere that hands the caller a timeout loop to write. `requestUrl` + "poll it briefly" would be the first.

None of this touches the finding, which stands and is the valuable part of the PR: 1145 alone does not elicit the URL, and 1146 does. That's the bit hardware told you and nothing else could.

martijnpoppen:
Reopened, squashed to a single commit, and re-scoped. Revisiting the two comments above:

**My own closing note was wrong.** I said the fix was to read the URL from the cloud record and that no SDK change was needed. On the hardware in question the record carries no `rtsp://` at all (the record-only read logged "no reported RTSP URL in the record"), so that was a dead end. The URL is only obtainable via the P2P round-trip — so an SDK change *is* needed, and it's this one.

**@lenoxys — your review holds, and this revision is what it argued for.** The objection was against `requestUrl` + "poll `dev.rtsp()?.url`" — a read dressed as a write with a caller-side timeout loop, which nothing else in the SDK does. That's removed. What's here now is the shape you pointed at: a request/reply facade (`reportedRtspUrl(sn)` → `readReportedRtspUrl`) that **returns** the URL, arms the listener before provoking, is bounded by one deadline, and returns `undefined` as the typed "no reply" — the `ptz.preset().image` / `getAutoLockState` convention, not `smart-light.refreshState`.

On your one caveat — that the 1145 push is a bare NUL-terminated string, not `p2pQuery`'s `json.cmd`/`payload` reply, so it isn't a drop-in: agreed, and that's exactly why it uses a dedicated `awaitRtspUrl` (a channel-filtered wait on the `rtspUrl` event) rather than `p2pQuery`. I've flagged in "Unverified or deferred" that if a third reply-wait ever lands, `p2pQuery` / `p2pControlQuery` / `awaitRtspUrl` are the point to factor the shared promise/listener/timeout skeleton out.

The finding you singled out as the valuable part stands: 1145 alone does not elicit the URL, 1146 does — verified now end to end, with live video playing off the returned URL.

---

src/transport/p2p/command-router.ts
```diff
@@ -1092,6 +1096,84 @@ export class P2PCommandRouter {
...
+      };
+      const cleanup = (): void => {
+        session.off("rtspUrl", onUrl);
+        signal.removeEventListener("abort", onAbort);
+      };
+      if (signal.aborted) return reject(signal.reason ?? new Error("aborted"));
+      session.on("rtspUrl", onUrl);
+      signal.addEventListener("abort", onAbort, { once: true });
+    });
+  }
```
lenoxys, line 1175: **stdlib:** this is the third hand-rolled copy of the promise + listener + abort + cleanup skeleton, after `p2pQuery` (`:1010`) and `p2pControlQuery` (`:1049`). Your own "deferred" note says *"if a third reply-wait shape lands later"* — this is it. Two options, both shorter than twenty lines:

`node:events` already does the whole thing, including buffering (so the re-arm gap you're guarding against cannot exist) and listener removal on abort:

```ts
for await (const [ev] of on(session, "rtspUrl", { signal })) if (ev.channel === channel) return ev.url;
```

No precedent for `events.on` in this repo — every other file imports `EventEmitter` and nothing else — so that's a judgement call, not a slam dunk. The alternative is to factor the skeleton across all three now that there are three, which is what your note proposes and what "Reduce after green" asks for.

Either way I'd rather not merge a third copy with a comment saying a third copy would be the trigger to stop making them.

src/transport/p2p/command-router.ts
```diff
@@ -1092,6 +1096,84 @@ export class P2PCommandRouter {
...
+      const url = this.awaitRtspUrl(session, channel, signal);
+      // Provoke through the ONE level decision (and its RF-resilience repeat), but do not block the read
+      // on the retransmits finishing: the URL push can land after the first datagram, so the read
+      // returns as soon as it arrives (or the deadline aborts), while the repeats run to completion.
+      void this.resolveScalarParam(sn, CommandType.CMD_NAS_SWITCH, 1, "auto").catch((e) =>
+        log.debug(`[p2p] ${sn} RTSP publish-switch provoke failed`, e),
+      );
+      void this.resolveScalarParam(sn, CommandType.CMD_NAS_TEST, 1, "auto").catch((e) =>
+        log.debug(`[p2p] ${sn} RTSP livestream provoke failed`, e),
+      );
```
lenoxys: **shrink:** two provokes, two identical `void … .catch(log.debug)` blocks. One loop over `[CMD_NAS_SWITCH, CMD_NAS_TEST]` with the id in the message says the same thing in two lines. -4

Not awaiting them is right and the reason is well documented — the push can land before the retransmits finish.

src/client/eufy-mega.ts
```diff
@@ -1185,6 +1185,21 @@ export class EufyMega extends EventEmitter {
...
+   *
+   * A thin public door onto the P2P transport (which stays internal otherwise): opens the
+   * station's session on demand, so a viewer adopting a tile can call this directly without one
+   * already existing, and bounds + swallows every failure itself (no route, level-2 not ready, no
+   * push in time) into `undefined` — nothing for this facade method to add. It writes only the
+   * publish switch and the test-stream provoke, never the credentials, so a stream a NAS/NVR already
+   * consumes keeps its own pair.
+   */
+  async reportedRtspUrl(sn: string): Promise<string | undefined> {
+    return this.p2p.readReportedRtspUrl(sn);
```
lenoxys, line 1208: **Four different failures collapse into one `undefined`:** no route, no account id, level-2 not ready, and no push before the deadline. The doc calls that "nothing for this facade method to add", but there is something — whether the caller should try again.

A viewer adopting a tile gets `undefined` and cannot tell "this camera has no RTSP" (terminal, stop asking) from "the HomeBase session wasn't warm yet" (retry in two seconds and it works). It will therefore either retry forever or give up on a camera that would have answered.

The siblings already draw that line: `camera.snapshotLive` rejects with `LiveSnapshotUnavailableError` carrying `retryable`, and `ptz.preset().image` reserves `undefined` for "nothing stored" while a reply that never arrives rejects. `AGENTS.md`'s "typed data or a typed reason" is the same rule. Keeping `undefined` for the genuinely terminal case and rejecting (or returning a reason) for the transient one is a small change and the difference between a tile that recovers and one that doesn't.

**Minor, same method:** `RTSP_URL_READ_TIMEOUT_MS` is fixed at 12 s with no override, where both sibling queries take `{ timeoutMs }`. Fine if deliberate — worth a clause saying so, since 12 s is a long time to hold a UI and no time at all for a cold HomeBase.

lenoxys (changes requested):
Re-checked at `687510b`. **You won the important half of the argument and the design objection is resolved** — this is a request/reply that returns the URL, the poll-loop contract is gone, and the wire is hardware-verified end to end, which retires the `unverified`/`method()` objection entirely (it isn't a capability method any more). The bare-string envelope caveat we discussed is handled honestly rather than forced through `p2pQuery`. Credit where it's due: arming the listener before the provokes, filtering by channel because a HomeBase multiplexes cameras over one session, and one deadline covering both the connect wait and the URL wait are all the right calls.

Verified locally: `tsc --noEmit` clean, all five guards (`capability-ownership`, `decorrelation`, `consumer-agnostic`, `docrefs`, `lines`) pass, and 521 tests across `rtsp` + `p2p` green.

One blocker, one design consistency note, two cuts.

## Blocking: there are no tests, and the PR body says there are

`Added or updated specs (offline, synthetic fixtures) — rtsp capability specs` is checked. The diff against the merge-base touches four files and **not one of them is a spec**. The previous revision's `requestUrl` test went away with `requestUrl` and nothing replaced it, so this revision's net test movement is negative.

That matters more than usual here, because the new logic is exactly the kind that fails silently: a channel filter, an ordering constraint (arm before provoke), an abort path, and a `startsWith("rtsp://")` gate on a wire that also carries a bool. The check I'd actually want is the one the code was written to make safe — two channels on one session, a push for the *other* camera must not resolve this read. Nothing proves that today, and it's the difference between a viewer showing the right camera's URL and the wrong one's. One spec, that case, is enough.

## Not a rule violation, but worth a deliberate answer

`reportedRtspUrl` is the only feature-named **device-wire** read on the client facade. I checked the alternatives before saying so: `getCleanRecords` is a cloud/HTTP account read, and the facade's own `p2pQuery` is `@internal` and generic ("the caller owns the sub-command id"). Every other device-wire read — `snapshotLive`, `record`, `preset().list`, `quickResponses`, `getAutoLockState` — reaches callers as a capability member over an injected provider. `guard:capability-ownership` passes, so nothing is broken; it's that a consumer now gets this URL from `mega.reportedRtspUrl(sn)` while the same value is also `dev.rtsp()?.url`.

Two doors to one value is the part I'd resolve rather than leave: either put it behind the seam the siblings use (`provided("media", …)` on `rtsp`, `answers: true`, per `doorbell.ts:340`), or keep the facade and say in the docs which door is canonical and what the member is for. Right now the member's own JSDoc calls itself the alternative "for a consumer that prefers to read `dev.rtsp()?.url`", which reads like the decision hasn't been made.

martijnpoppen:
Thanks — thorough as ever. Addressed in `d8c2f8d`:

**Blocker — tests (and the body's false claim).** Fixed both. Restored `read-reported-rtsp-url.spec.ts` (the transport spec #114 removed), which leads with the exact case you named: two pushes on one session, the wrong channel first, and the read resolves with the *right* camera's URL — plus the level-1/level-2 provoke wires, the timeout leaving `listenerCount("rtspUrl") === 0`, and the no-account / no-session guards. The "Added specs" box is now true (six tests), not aspirational.

**Two doors to one value.** Made the call in the docs rather than leave it open: `reportedRtspUrl(sn)` is the **canonical fetch** (provokes + returns); the `url` member is inbound state for code already holding a `dev.rtsp()` and reacting to `propertyChanged`, explicitly *not* a second fetch path. Reworded both JSDocs so neither reads as "the alternative". I kept the facade rather than moving to a `provided("media", …)` seam — the read has no `dev`/`ctx` at the call site (a viewer adopting a tile may have no capability object yet), which is the case the facade exists for; happy to revisit if you'd rather it live on the capability.

**Two provokes → one loop.** Done — one loop over `[CMD_NAS_SWITCH, CMD_NAS_TEST]`, same two datagrams.

**Four failures → one `undefined`.** You're right that terminal-vs-transient is a real distinction. I've documented it honestly for now (and noted the fixed 12 s is deliberate, not an oversight) rather than change the contract: the sole consumer already retries `undefined` on its own backoff, so it recovers from the transient case — and a typed `retryable` reason ripples into that consumer (already on `main`), so it's a deliberate follow-up, not this PR.

**Third reply-wait skeleton.** Conceded as a real smell, but I'd rather not fold a transport-wide refactor into an RTSP fix. `events.on` isn't a clean drop-in here — it also surfaces the session's `error` events into the read, changing the failure semantics — and extracting the shared skeleton touches `p2pQuery`/`p2pControlQuery` and their specs. Left it as the explicit listener and called it out under "Known limitations" as the follow-up that retires all three copies at once.

`npm run verify` green (180 files, 3370 tests).

---

src/transport/p2p/command-router.ts
martijnpoppen: Kept the explicit listener for now and flagged the extraction under "Known limitations" as a follow-up. Two reasons not to fold it in here: `events.on(session, "rtspUrl", {signal})` also surfaces the session's `error` events into the read (its async iterator rejects on an emitter `error`), which changes this read's failure semantics; and factoring the shared skeleton properly means touching `p2pQuery`/`p2pControlQuery` and their specs — a transport-wide refactor I'd rather not bundle into an RTSP fix. Happy to do that as its own PR that retires all three copies at once.

---

src/client/eufy-mega.ts
martijnpoppen: Documented both in the facade JSDoc: the four failure modes collapsing to `undefined`, and the 12 s window being deliberate. Left the contract as `undefined` rather than a typed `retryable` reason for now — the sole consumer (the Homey app's `RtspPublication`) already retries `undefined` on its own backoff, so it recovers from the transient case rather than giving up, and adding a typed reason changes the facade signature and ripples into that consumer (already merged). Worth doing, but as a deliberate follow-up rather than in this fix. Noted under "Known limitations".

---

lenoxys (approved):
Re-checked at `e856752`. The blocker is closed and then some.

**The specs are what I asked for and more.** "ignores a push for another channel on the multiplexed station session" is exactly the case the code was written to make safe and the one nothing proved — and it emits the wrong-channel and right-channel pushes *in the same turn*, which is the version that would catch a re-arm gap rather than just a missing filter. The timeout case asserting `listenerCount("rtspUrl") === 0` is the one I'd have forgotten to ask for. Six specs covering both wire levels, no account id, and no resolvable session.

**The two-doors question is answered by deciding, which is all I wanted.** `reportedRtspUrl` is stated canonical in both places, and the `url` member now says what it is *for* — inbound state for code holding a `dev.rtsp()` and reacting to `propertyChanged` — rather than describing itself as the alternative.

**Both declines are well argued, and one corrects me.** You're right that `events.on(session, "rtspUrl", { signal })` surfaces the emitter's `error` events into the read — its iterator rejects on an emitter `error`, and `P2PSession` emits those. That would have changed this read's failure semantics from "ignore, wait for the deadline" to "reject on any session error", which is not a refactor, it's a behaviour change smuggled in as one. My suggestion was wrong on that point and the twenty explicit lines are the right call until the three-copy extraction happens as its own PR.

The `undefined`-versus-typed-reason deferral I'll also take: the reasoning is recorded in the JSDoc, the sole consumer already retries on `undefined`, and changing the facade signature ripples into something already merged. Documented deliberately beats changed hastily.

Approving.

PR #128, opened 2026-09-09
fix(battery): no bogus battery entity on mains cameras (T8425/T8419)

max246:
## Problem
A **wired camera** (T8425 Floodlight Cam; also the T8419 Indoor Cam noted in-code) reports the battery param **1101** as a fixed sentinel (0/100), not a real cell. Because the battery capability detects on `evidenceParams:[1101]`, these mains cams got the capability and a **bogus battery %/icon** (and a charging sensor) in HA/consumers.

## Fix
Keep the `battery` capability (so working-mode / recording still apply — the suite requires T8419 to keep its `workingMode` read), but gate the **physical-battery reads** (`level`, `charging`) behind `notMainsCamera(ctx)` — matched by T-code prefix `MAINS_CAMERA_MODELS = [T8425, T8419]`. Mains cams no longer publish a battery %/charging entity; non-camera devices and real battery cams are unaffected.

Member-level (not capability-level) on purpose: removing the whole capability would also strip `workingMode`, which the model spec expects T8419 to keep.

## Test
- 3433/3433 pass; build, prettier, guards clean.
- New assertion: `level`/`charging` `available` is false for T8425/T8419, true for T8114 + unknown.


src/model/capabilities/battery.ts
```diff
@@ -133,6 +133,22 @@ function recordSetting(param: number, value: number, ctx: CommandContext): Comma
...
+ */
+const MAINS_CAMERA_MODELS = ["T8425", "T8419"] as const;
+
+/** False for a mains camera that only reports 1101 as a sentinel — used to gate the physical reads. */
+const notMainsCamera = (ctx: AvailabilityContext): boolean => {
+  const model = (ctx.model ?? "").toUpperCase();
+  return !MAINS_CAMERA_MODELS.some((prefix) => model.startsWith(prefix));
+};
+
 export const BATTERY_MEMBERS = {
```
lenoxys: The two new consts land **between** `BATTERY_MEMBERS`' doc block and the table itself, so that block now documents `MAINS_CAMERA_MODELS`. It's the table's doc — "Every `battery` feature, declared once … Order is schema order … Exported so a caller can name the table" — and its `@internal` tag now applies to a two-element string array instead of the export it was written for.

Move both consts above the doc block.

src/model/capabilities/__tests__/battery.spec.ts
```diff
@@ -69,6 +69,22 @@ describe("battery capability module", () => {
...
+      return a ? a({ model }) : true;
+    };
+    // Mains cameras (T8425 Floodlight, T8419 Indoor) must NOT publish a battery %/charging entity.
+    for (const model of ["T8425P00", "T8419P00"]) {
+      expect(avail("level", model)).toBe(false);
+      expect(avail("charging", model)).toBe(false);
+    }
+    // Real battery cams (and unknown models) keep them.
+    expect(avail("level", "T8114P00")).toBe(true);
+    expect(avail("level", undefined)).toBe(true);
```
lenoxys: This proves the predicate, not the claim. It reaches through a cast into `members[name].available` and calls it with a hand-made `{ model }` — so it would pass just as happily if `available` turned out not to gate getters at all, which is the one thing that could have made this whole PR a no-op. (It does gate them — I checked — but the test isn't what told me.)

The claim in the PR title is that no battery entity is published. That's `propertiesOf` at the same cost:

```ts
expect(propertiesOf(BATTERY.members, ctxFor("T8425P00")).map((p) => p.name)).not.toContain("battery");
```

Same line count, no cast, and it fails if the wiring changes underneath the flag rather than only if the flag's own body changes.

lenoxys (commented):
Checked the mechanism before trusting it, and it's right: `available` is not just a write gate despite what its JSDoc says — `bindMembers` applies it to the getter too ("One availability decision across getter, setter and manifest", `members.ts:920`) and `propertiesOf` drops the member from the published schema (`:606`). `audio.ts:84` is precedent for using it on read-only members. So the bogus entity really does disappear, and member-level over capability-level is the right call for the reason you give — stripping the capability would take `workingMode` with it.

Three things, none of them blocking, one worth a deliberate answer.

## These cameras are still "battery-powered" to the streaming path

`camera.ts:294`:

```ts
function poweredOf(ctx: CommandContext): "wired" | "battery" {
  return ctx.capabilities?.has("battery") ? "battery" : "wired";
}
```

Keeping the capability — which is the whole point of doing this member-level — means a T8425 still resolves as `battery` there, so its stream gets budgeted like a cell-powered camera when it's on mains. `level`'s own JSDoc already flags this coupling: *"this capability's detection evidence … which is also the fact `MediaProvider` reads to decide a stream needs a power budget."*

So the fix trades a visible wrong entity for an invisible wrong budget. That may well be the right trade — an unnecessary budget is cheap, a phantom battery icon is a support ticket — but right now it's implicit. Worth a line in the JSDoc saying the budget consequence is known and accepted, or a follow-up.

## The model list — a question, not an objection

`device-family.ts` already has `isFloodLight` and `isIndoorCamera`, and `AGENTS.md` says to compose family predicates rather than duplicate device-type sets. But composing those two would over-reach: not every floodlight or indoor cam is mains-only, and you'd be asserting mains for hardware nobody has checked. A list naming exactly the two confirmed devices is the more honest shape, and `WORKING_MODE_DEFAULT_MODELS` at `:443` is precedent in this very file. I think you chose right — worth saying why in the JSDoc so the next reader doesn't "fix" it into a family predicate.

Separately: the PR says T8425 is owned hardware and T8419 comes from the in-code mains-cam note. Those are different evidence bars; the JSDoc claims both as "confirmed on owned hardware".

max246:
Thanks — addressed all five in `97fa80b`:

- **Const placement:** moved `MAINS_CAMERA_MODELS` / `notMainsCamera` **above** the `BATTERY_MEMBERS` doc block, so that block (and its `@internal`) documents the table again.
- **Test:** now asserts the *claim* via `propertiesOf(BATTERY.members, ctx)` — no `battery`/`charging` property published for T8425/T8419, both kept for T8114 and unknown models — instead of reaching through a cast into `member.available`. So it fails if the availability wiring regresses, not only if the flag body changes.
- **List not predicate:** JSDoc now says why an explicit list over `isFloodLight`/`isIndoorCamera` (composing over-reaches; assert mains only for checked hardware; `WORKING_MODE_DEFAULT_MODELS` precedent).
- **Evidence bars:** corrected — T8425 owned-hardware, T8419 from the app's own mains-cam handling.
- **Power budget (the deliberate one):** documented as a known, accepted trade — because the capability stays, `poweredOf` still resolves these as `battery`, so a stream is budgeted as cell-powered; a needless budget is cheap vs a phantom icon = a ticket. Called out that a `poweredOf` refinement is a separate change if we want it.

3433 tests pass; build/prettier/guards clean. No runtime behavior change from these — the deployed bridge image (0.1.54) already behaves correctly.

lenoxys:
@max246 just need to fix the CI/CD and we're good

PR #129, opened 2026-09-09
fix(rtsp): send the served-livestream frame, not the NAS setting alone

martijnpoppen:
## What changed

`rtsp.publish()` sent only the persistent NAS setting (`1145 CMD_NAS_SWITCH`). Publishing is two frames, and it now sends both:

| Frame | Meaning |
| --- | --- |
| `1145` `CMD_NAS_SWITCH` | the persistent NAS/RTSP setting — what `published` reads, what the app's toggle shows |
| `1146` `CMD_NAS_TEST` | start the served livestream — what makes the device report its URL |

- `RTSP_PARAM.TEST_STREAM` (1146) and `streamCommand()` in `rtsp.ts`.
- `publish()` → setting then livestream switch; `withdraw()` → the reverse order.
- New `startStream()` / `stopStream()` drive the livestream half alone, for a camera whose setting stays on while its endpoint follows a recorder.
- The `rtspStream` property write is **unchanged** — the intent path resolves to one command by construction, so it stays the setting alone. A spec pins that asymmetry so a caller who writes the property and expects a served endpoint hits a stated limit rather than a silent one.
- `url`'s evidence note corrected: the provoke is grounded on the livestream frame, no longer inferred from the publish switch.
- `docs/rtsp.md`: the two frames, and a "When nothing plays" checklist ending at the serving device's own firmware.

## Why

Writing 1145 alone leaves `published` reading `true` while `url` stays `undefined` forever — and that URL is the only place the credentials live (the device regenerates them on every publish toggle, the cloud record lags a cycle). So a caller has no address to point a player at, and the failure looks exactly like a device that is serving nothing.

1146 was dropped in 75125eb, which moved the URL read from a transport facade to inbound capability state. That move was right; it took the start-stream frame with it on the assumption that the setting alone provokes the push. The `url` member has carried the admission since:

> The PROVOKE — that `publish()` (1145=1) ALONE elicits that push — is **INFERRED**: the only live capture also sent `CMD_NAS_TEST` (1146, since removed), so the switch has not been seen to elicit the URL on its own. **If it turns out not to, the URL never populates.**

It turns out not to.

## Evidence

The sequence `1145=1` then `1146=1` is what the one live capture that ever produced a URL sent, on a standalone camera, which served its endpoint under it (the capture behind 14488b5 — "confirmed end to end on a standalone camera"). The publish switch has never been observed to elicit a push on its own.

1146 is `nasTestStream` in `param-dictionary.ts`, provenance `apk`.

Both frames go out through `setScalar(..., "auto")` — the one level decision the normal publish already used, so a keyed HomeBase seals the livestream switch at level 2 and a standalone camera at level 1. No new wire form is introduced here.

Claims already in the file were **narrowed, not widened**: `STREAM_SWITCH` records 1145 alone flipping `DESCRIBE` 404→200 on real devices, so nothing here says the setting serves nothing. What it says is the setting alone never reports the URL — which is the part a caller cannot work around.

## Verified how

- [x] `npm run verify` is green locally — 3375 tests, typecheck, all five guards, build, ESM check, examples typecheck
- [x] Ran against a real device — no device in this environment
- [x] Added or updated specs (offline, synthetic fixtures) — the publish/withdraw frame pair and its order, `startStream`/`stopStream` alone, `form: "auto"` on the livestream switch for a HomeBase-attached camera, and the property write staying a single setting frame
- [x] Docs updated

## Unverified or deferred

**The frame pair is not re-confirmed on hardware by this PR.** Both ids and the sequence are grounded in the capture cited above, and both are plain scalar writes on a wire the module already drives — but nobody has re-run `publish()` → `url` on a device since the switch was restored. The fix is falsifiable in one read: if `url` populates after `publish()`, it is right.

**The level-2 wire form for 1145/1146 on a HomeBase is untouched and still open.** eufy-sdk sends these as a level-2 direct-binary body (`buildDirectBinaryBody`, 136 bytes, GCM signCode 8). Other clients send the int-string body shape for these two commands, ECB-encrypted at whatever level was negotiated. Those are different protocol generations with different key negotiations, so porting that form would be a guess, not a fix — but if `url` still never lands on a HomeBase-attached camera after this change, that is the next thing to look at.

**Not the same problem as a station that closes 554.** Some station firmware accepts and acknowledges both frames and still never serves an attached camera; no frame a caller sends changes that. Before this PR the two failure modes were indistinguishable from a player — now `url` separates them, which is what the docs checklist is for.

https://claude.ai/code/session_017dHzL1VseTdSH4tL4LadaY

---
_Generated by [Claude Code](https://claude.ai/code/session_017dHzL1VseTdSH4tL4LadaY)_


martijnpoppen:
Closing this — #127 covers the finding with hardware evidence this PR doesn't have, and field testing since has shown this PR doesn't address the reported symptom.

**Why it shouldn't land**

- The core finding — 1145 alone doesn't elicit the URL, 1146 does — is already in #127, verified end to end on a device. This PR inferred it from the repo's own earlier capture; #127 proved it.
- The `publish()` change here serializes the provokes: `await dispatch(1145)` then `await dispatch(1146)`, and `sendIntStringCommand` replays `DIRECT_CMD_SENDS = 5`× at 200ms, so 1146 leaves ~1s after 1145. #127 deliberately fires both un-awaited after arming a listener, and reports that a trace showed no `rtsp://` payload without that exact sequence. So this PR ships a described control whose two-frame behaviour was never confirmed — against "Describe only a confirmed wire... A described control that always fails is worse than none."
- It conflicts with #127 in `rtsp.ts` (both rewrite the same three JSDoc blocks).

**Field result that closed it out**

The reporting setup is a battery eufyCam 3 Pro behind a **T8030 HomeBase 3**. A raw RTSP `DESCRIBE` against the station returns `ECONNREFUSED` — an active refusal, so the host is up and nothing is bound to 554 — using a URL copied from the vendor app itself. The station accepts the publish and never opens the port, which is the bropat/eufy-security-ws#538 shape and unreachable from any client-side frame ordering.

Worth recording that #127's proof is a T8410 behind a **T8010** — a wired indoor camera behind a HomeBase 2. Neither PR has evidence on a T8030, and a battery camera is the case the vendor app itself restricts continuous recording away from.

**One item here that is a real gap in #127, raised there rather than lost:** after #127, `readReportedRtspUrl` writes `1146=1` and nothing in the SDK ever writes `1146=0` — `withdraw()` clears 1145 only, so a started livestream has no off switch.

---

PR #130, opened 2026-09-09
perf: bundle dist into one ESM file, and read a control command's result

martijnpoppen:
## What changed

Two independent things, both small.

**`dist` is bundled into a single ESM file.** `npm run build` is now `tsc` followed by
`scripts/bundle.mjs`. Runtime dependencies stay external, read from `dependencies` so adding one
cannot silently change what is inlined. Declarations are untouched: `tsc` still emits the whole
`.d.ts` tree and `exports.types` still points into it, so only `.js` and `.js.map` are replaced.

**`commandResult` now fires for a control reply, not only a media one.** A reply with no JSON body is
the int32 result of the command just sent, and both wrappers answer that way — media over
`CMD_SET_PAYLOAD` (1350), device control over `CMD_CONTROL_PAYLOAD` (1700). Only the first was read.

## Why

The package has one entry point and no subpaths, so a consumer evaluates all of it on import and pays
for the shape it is emitted in. ~200 module records, per-file source maps and compilation units cost
far more resident memory than the same code in one file:

| Shape | Barrel cost | `dist` JS | Files |
| --- | --- | --- | --- |
| the tree | +28.4 MB | 4.3 MB | 131 |
| bundled | **+18.3 MB** | 0.96 MB | 1 |

10 MB off every consumer's baseline before a line of their code runs, and more than the lazy-loading
pass in #125 took. It matters most to a memory-limited host: the consumer that prompted the
measurement idles around 125 MB against a platform that warns at 150 and force-stops near 160.

The `commandResult` gap costs more than a missing value. A station refusing a command reached the
caller as silence, which is exactly what a command the station never answered looks like — three
clean rejections read as no reply at all for two probe runs before the wrapper, rather than the
station, turned out to be the difference.

## Evidence

**Bundling:** Node 22.22.2 x86-64, one fresh process per import, RSS from `/proc/self/statm`, three
runs each and stable to 0.5 MB. No device involved.

**`commandResult`:** the 1700-carries-its-result-like-1350 finding comes from probe runs against real
hardware, written up in the consumer's `cmd-snapshot.md`. `-104` (file not found) and `-108`
(refused) are the codes seen. The code is carried through unmapped — naming them is a caller's
business and the set is not known to be closed.

## Verified how

- [x] `npm run verify` is green locally — format, typecheck, all five guards, build, `check:esm`,
      examples, 3369 tests
- [x] Ran against a real device — no device in this environment; neither change sends anything new
- [x] Added or updated specs (offline, synthetic fixtures) — `command-result.spec.ts` drives the
      private frame handler and fails on the 1700 case alone against the previous condition, which I
      confirmed by stashing the fix and re-running
- [x] Docs updated, or n/a — the measurement is written up in the consumer's `sdk-import-cost.md`;
      nothing in this repo names a consumer

Also checked before bundling: nothing in `src/` depends on module file location. The only
`import.meta.url` is `createRequire` in `transport/protobuf.ts`, which resolves `protobufjs` the same
way from `dist/index.js` as from `dist/transport/protobuf.js`, and the `.proto` documents are
referenced in prose rather than read from disk. All 266 exports survive `check:esm`, and the Homey
consumer validates at its strictest level against the bundled package with no source change.

## Unverified or deferred

- **Not minified.** It measures ~2 MB better and costs readable frames in a package three consumers
  debug. The lever stays available if a budget gets tight.
- **Bundled stack traces.** Source maps chain back to the TypeScript with contents inlined, so the
  emitted `.js` being gone does not break them — but I have not read a real bundled trace from a
  running consumer.
- **No RSS number from a constrained host.** Every figure here is from a workstation. Whether 10 MB
  is the difference between crashing at idle and not is the consumer's to measure.
- `requestImage`'s JSDoc still says its send path needs live-device validation. This PR does not
  change that; it only makes the refusal readable when someone does validate it.

---
_Generated by [Claude Code](https://claude.ai/code/session_017u6pJrkPeWGqjnW4vbLjTB)_


scripts/bundle.mjs
```diff
@@ -0,0 +1,64 @@
...
+  const out = [];
+  for (const entry of readdirSync(dir)) {
+    const path = join(dir, entry);
+    if (statSync(path).isDirectory()) out.push(...emittedJs(path));
+    else if (entry.endsWith(".js") || entry.endsWith(".js.map")) out.push(path);
+  }
+  return out;
+}
+
+for (const file of emittedJs(DIST)) rmSync(file);
```
lenoxys: **stdlib:** the recursive walk is `readdirSync`'s job since Node 18.17, and this package requires >= 24.5. Ten lines plus a `statSync` per entry become:

```js
for (const f of readdirSync(DIST, { recursive: true }))
  if (f.endsWith(".js") || f.endsWith(".js.map")) rmSync(join(DIST, f));
```

Verified on this Node 24.19. That also drops the `readdirSync`/`statSync` pair from the import list.

Smaller, and take it or leave it: esbuild has `allowOverwrite: true` for exactly the bundle-in-place case, which would remove the `.bundle` staging directory and both renames. Delete the emitted `.js` first, then write `dist/index.js` — esbuild has the entry in memory by then. I haven't run it, so treat that one as a suggestion rather than a finding.

src/transport/p2p/__tests__/command-result.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+  it("reports nothing for a reply that carries a JSON document", () => {
+    // A document is the answer itself; only a bodyless reply is a bare result code.
+    const body = Buffer.concat([Buffer.from(JSON.stringify({ cmd: 1306, count: 0 }), "utf8"), Buffer.from([0])]);
+    expect(results(1700, body)).toEqual([]);
+  });
+
+  it("reports nothing for a body too short to hold one", () => {
+    expect(results(1700, Buffer.from([0, 1, 2]))).toEqual([]);
+  });
+});
```
lenoxys, line 74: Good specs — five cases, both negative codes from real probe runs, and the JSON-body and short-body exclusions are the two that matter most.

One case missing, and it's the one the broadening introduces: a 1700 reply carrying a **non-JSON string** body. The gate is `!frame.json && data.length >= 4`, so any four-plus bytes of text now decode as an int32 and emit a `commandResult` with a nonsense code. 1350 was narrow enough that this never came up; 1700 is the general control wrapper.

If no such reply exists on 1700, one spec asserting that pins the assumption. If one does, the gate needs to be tighter than "not JSON and long enough".

lenoxys (changes requested):
Both changes look sound on their merits and the `commandResult` one is genuinely valuable — a station refusing a control command reaching the caller as silence is the worst kind of bug, indistinguishable from no reply, and you found it the hard way across two probe runs. The measurement discipline is the same as #125's and just as trustworthy.

My objection is that they're in one PR.

## Split them

#125 deferred bundling in these words: *"a packaging decision rather than an internal one, so it should be its own PR with its own discussion."* This is that PR, plus an unrelated P2P frame-decode fix riding along.

They share nothing. The decode fix is nine lines, hardware-evidenced, covered by five new specs, and mergeable today. The bundle changes what every consumer downloads, replaces 131 emitted files with one, and carries the risks you list yourself: unread stack traces, no RSS figure from a constrained host, and a source-map chain that hasn't been exercised. Bisecting a consumer regression to "the release that bundled *and* changed frame decoding" is exactly the position you don't want to be in, and reviewing them together means the small certain win waits on the large uncertain one.

Land the `commandResult` fix on its own — I'd approve that today. Keep the bundle here with its discussion.

## On the bundle itself

The design is right: deps external so the #125 lazy loads survive, declarations untouched, `exports.types` still pointing into the `.d.ts` tree, and you checked the one `import.meta.url` in `transport/protobuf.ts` resolves the same from `dist/index.js`. Inlining `external` from `dependencies` so adding a dep can't silently change the bundle is the good kind of paranoid.

**The source map is the risk worth closing before merge, and it's one command.** You wrote "I have not read a real bundled trace" — you don't need a running consumer for the part that can actually be wrong. `dist/index.js.map` either carries `sourcesContent` with the TypeScript in it or it doesn't:

```
node -e "const m=require('./dist/index.js.map');console.log(m.sources.length, m.sourcesContent?.length, m.sources.slice(0,3))"
```

If `sourcesContent` is absent or the sources point at the `.js` files the script then deletes, every consumer trace is unmappable and the answer is `sourcesContent: true` in the esbuild options. I couldn't run it here — `esbuild` isn't installed in this checkout.

Also worth stating in the script's doc: nothing holds the bundle in place, same as #125's note about the lazy loads. A `build` that silently stops bundling looks identical to one that does.

scripts/bundle.mjs
martijnpoppen: Both taken, in b2461b6.

`readdirSync(DIST, { recursive: true })` replaces the walk, and `statSync` leaves the imports with it.

I did run the `allowOverwrite` one before taking it, since you flagged it as untested. It works, with one correction to the sequencing you suggested: deleting the emitted `.js` first would remove the inputs esbuild still has to read. It resolves and loads the whole graph before writing anything, so overwriting the entry in place is safe on its own — build first, then delete everything except `index.js` and its map. Output is identical: 0.96 MB, one `.js`, one `.js.map`, 131 `.d.ts`, 266 exports through `check:esm`, and +18.3 MB RSS on the same measurement as before.

That removes the staging directory and both renames, so the script is now build-then-prune with no temporary anywhere.

---

src/transport/p2p/__tests__/command-result.spec.ts
martijnpoppen: You're right, and it's worse than a missing spec — the gate itself was wrong. Fixed in b2461b6.

I went looking for whether a non-JSON body can reach that branch on 1700, and it can, without needing a device to prove it. `handleFrame` leaves a body as ciphertext whenever the decrypt above declines: the level-1 path only runs when the body is 16-byte aligned, and the level-2 path keeps the raw payload when `decryptLevel2` returns nothing. Ciphertext is not JSON and is comfortably over four bytes, so `>= 4` would have read its first word and emitted a fabricated code — for a command whose answer was in fact never recovered. As you say, 1350 was narrow enough that it never came up.

So rather than pin the assumption, I tightened the gate: the body IS the int32, so `data.length === 4`. Anything longer is a different kind of frame. Two specs cover it — a 16-byte ciphertext-shaped body and a plain text one — and both fail against the previous condition, which I checked by putting `>= 4` back and re-running.

Worth noting the narrowing also applies to 1350, where the same fabrication was possible and pre-existing. Nothing in the library reads `commandResult` — it's a diagnostic for external listeners — so there was no internal caller to break, and I'd rather it under-report than invent a code.

---

lenoxys:
## Independent corroboration, from frame lengths rather than probe runs

Came at the `commandResult` half of this from the opposite direction — measuring plaintext sizes in a decrypted pcap of the app doing things — and it lands on the same conclusion, which seems worth recording since the methods share no assumptions.

Every control write in two captures got a reply of **36 B on the wire → 4 B of plaintext** after the sign-8 GCM overhead (16 tag + 12 nonce + 4 sub-header). Four bytes is an int32, and it matches "a reply with no JSON body is the int32 result of the command just sent" exactly.

| outer cmd | write plaintext | reply plaintext |
| --- | --- | --- |
| 1700 (notification style, JSON body) | 69 B | **4 B** |
| 1246 (working mode, 136 B direct-binary struct) | 136 B | **4 B** |

## One case that may fall outside this PR

**1246 is not a wrapper.** It is a direct outer command — the param id itself, with the 136-byte `[u32 channel][u32 value][account_id]` body — and it still answered with a 4-byte reply, twice, on a standalone camera.

If the new read is keyed on the wrapper being 1350 or 1700, a direct command's result would stay silent, and by this PR's own argument that is the case where silence is most misleading: the direct-binary switches are exactly the writes with no other confirmation path.

Worth a check on whether the reply is recognised by "no JSON body on the session" or by "outer cmd is one of the two wrappers". I have not read the diff closely enough to tell which.

## Also observed, and not the same thing

On a `1350` write of sub-command 2730 the station additionally sent a **`1351` NOTIFY_PAYLOAD, signCode 2**, echoing the request payload back verbatim including its `transaction` field — alongside the 4-byte result, not instead of it. Two distinct replies to one write.

Noting it because the playbook's standing note says a state notify goes to OTHER clients and not to the setter; here the setter received it. Not investigated further, and there may well be a second session on the phone that explains it.

martijnpoppen:
You called it — it was keyed on the wrapper, and 1246 stayed silent. Fixed in 3256728.

**The gate is now the shape, not the id:** a non-media frame whose whole body is four bytes is the result of the command just sent, whatever it arrived under. Media is excluded because its bodies are never control plaintext. A spec covers 1246 directly, and it fails against the wrapper allowlist — I checked by putting the allowlist back and re-running.

Your framing of why is the part that decided it: a direct-binary switch is the write with the least other confirmation to fall back on, so an allowlist kept exactly the wrong ones quiet.

**On your frame-length measurement — that is stronger corroboration than it may look.** The four-byte rule got into this branch by a different route: `handleFrame` leaves a body as ciphertext when the decrypt declines (level-1 needs 16-byte alignment, level-2 can return nothing), so `>= 4` would have read ciphertext's first word and reported a code the station never sent. That argument is about what can *reach* the branch and says nothing about the wire. Your 36 B → 4 B across two captures measures the wire directly and lands on the same number. Two methods, no shared assumptions. I have recorded it in the code comment as measured rather than inferred, since it is now both.

**The 1351 echo I have left alone.** It is a separate reply carrying a JSON body, so it falls out on `!frame.json` and never reaches this branch — the four-byte result beside it is what gets read, which is the behaviour you would want either way. Whether the setter should be receiving that notify at all is a real question and not this PR's; your second-session hypothesis sounds likely, and it wants a capture with the phone app closed to settle.

Gate is green: 3373 tests, all guards, `check:esm`.

---

lenoxys (approved):
Re-checked at `3256728`. Everything I raised is addressed, the gate came back better than what I asked for, and **I built the branch and closed the source-map risk myself** rather than leaving it to a follow-up.

## The source map, settled

Installed esbuild and ran `npm run build`:

```
sources: 120        .ts sources: 120   .js sources: 0
sourcesContent: 120 entries, 0 empty
sample: ../src/transport/p2p/annexb.ts -> "/** * Annex-B elementary-stream helpers — …"
```

esbuild consumed tsc's input maps and chained all the way back to TypeScript, with the source text inlined — so a consumer's traces map without needing any source on disk. That was the one item I said should not stay deferred, and it's fine.

Also verified on the built output: `check:esm` gives 266 exports, `dist/index.d.ts` and all 131 declarations survive with `exports.types` still pointing at them, one `.js` and one `.js.map` at 0.96 MB, and — the part that matters most — `import("mqtt")` and `requireFrom("protobufjs")` are still there as deferred loads with zero dependency internals inlined. Bundling did not undo #125. `tsc --noEmit` clean, 514 p2p tests green.

## The gate

I asked for one missing spec; you found the gate was wrong instead, which is the better outcome. Two things I'd not have got to on my own:

- **Ciphertext can reach that branch.** `handleFrame` leaves a body encrypted whenever the decrypt declines, and ciphertext is neither JSON nor four bytes — so `>= 4` would have read its first word and reported a code the station never sent, for a command whose answer was never recovered. That's a fabricated value, worse than the silence this PR set out to fix.
- **`data.length === 4` with shape instead of an id allowlist.** Keying on 1350/1700 would have kept 1246 quiet, and a direct-binary switch is the write with the least other confirmation to fall back on — the allowlist would have silenced exactly the wrong ones.

Nine specs now, including 1246, the media exclusion, over-long bodies and non-JSON text. And you're right about the `allowOverwrite` sequencing — deleting first would remove the inputs esbuild still has to read. My version of that suggestion was wrong; build-then-delete is correct.

## Dropping the split objection

That was my main one, so I'll say plainly why I'm not holding it.

What I was managing was reviewing an unverified bundle alongside an unverified decode change and having no way to attribute a consumer regression. Neither half is unverified now: the bundle's output I have inspected directly, and the decode gate is corroborated by two methods that share no assumptions — probe runs against hardware, and plaintext frame lengths from a decrypted capture. The bisect argument doesn't disappear, but it's a preference about release hygiene, not a defect, and it isn't worth holding two verified changes over.

One residual, and it's the same gap #125 noted about its lazy loads: nothing holds the bundle in place. A `build` that quietly stops bundling looks exactly like one that doesn't, and the +10 MB comes back with no test failing. Worth a line under "Unverified or deferred" so the next person knows it's unguarded rather than guarded.

Approving.

PR #131, opened 2026-09-09
perf(http): stop the v2 thumbnail search climbing a ladder it has already failed

martijnpoppen:
## What changed

`decodeImageV2`'s coarse geometry search now walks the ladder in ascending MCU order and breaks at the first candidate that fails, instead of decoding every rung for each subsampling and keeping the largest that filled.

One helper (`ladderByMcuCount`), one `break`, and the module doc's cost note rewritten with measured figures. Plus a new spec that guards the cost rather than the output.

## Why

Reconstructing a v2 push thumbnail cost **~150 MB of resident memory and ~1.5 s of blocked event loop**. Almost none of that was the thumbnail.

Measured on the 176×144 fixture: **98 candidate decodes totalling 29.7 Mpx, of which 29.2 Mpx was spent by candidates that then threw** — three of them at 1920×1080. A decoder allocates a candidate's full component buffers before it can discover the scan does not fill it, so every oversized rung was paid for in full and then discarded.

Whether a candidate decodes is monotone in its MCU count: the plaintext scan carries a fixed number of MCUs, and any frame demanding more runs out of entropy-coded data and throws. Walked in ascending MCU order the first failure is therefore the last candidate worth attempting, and the largest frame that decodes is the one before it. **Nothing about which geometry wins changes** — only which candidates get built.

## Evidence

The monotonicity is a property of the format, not an observation: `jpeg-js` throws when the scan runs out, and MCU count is what determines how much scan a frame consumes. It is checked against both fixtures below.

|  | before | after |
| --- | --- | --- |
| **176×144** decodes | 98 | **26** |
| pixels decoded | 29.7 Mpx | **0.68 Mpx** |
| largest frame attempted | 1920×1080 | **320×180** |
| peak RSS | +158 MB | **+48 MB** |
| wall clock | 1490 ms | **194 ms** |
| **264×200** decodes | 111 | **38** |
| pixels decoded | 30.8 Mpx | **1.69 Mpx** |
| peak RSS | +173 MB | **+52 MB** |
| wall clock | 1703 ms | **290 ms** |

Output is **byte-identical** on both fixtures — same SHA-256 (`021129c2a83e15e4…`, `936e42330e13dd45…`) before and after — and the existing reconstruction specs pass unchanged.

## Verified how

- [x] `npm run verify` is green locally — 3366 tests, all six guards, build, `check:esm`, examples typecheck
- [x] Ran against a real device — **no.** Both fixtures are the synthetic 176×144 and 264×200 ones already in the repo
- [x] Added or updated specs (offline, synthetic fixtures) — `decode_v2_search.spec.ts`
- [x] Docs updated — the module's cost note; it claimed about a second per thumbnail and said nothing about memory, which was the larger half

The new spec exists because the reconstruction specs **could not have caught this**: they check that the right image comes out, and would pass just as happily if the search reached that answer the expensive way. It mocks `jpeg-js` to count decoder calls and read each candidate's SOF geometry, so it asserts an exact property — "the search does not attempt frames it can prove are too large" — where a stopwatch on a shared CI host would be flaky. Bounds are an order of magnitude below the unbounded walk rather than pinned to today's figures, so adding a ladder rung will not break it.

Checked that it bites: reverting the source gives `expected 98 to be less than 60` and `expected 111 to be less than 60`.

## Unverified or deferred

- **Not run against a live 632×472 4:4:4 thumbnail**, which is the geometry the module doc records this decoder being cracked against. The monotonicity argument covers it and its MCU count is far below anything the ordering skips, but the fixtures here are both small and synthetic. Worth one run before this is relied on.
- **The final `jpegEncode` is untouched.** It measures +12.8 MB on the first fixture: `jpeg-js`'s `encode()` constructs a new `JPEGEncoder` per call, and building its Huffman and bitcode tables costs the same regardless of picture size. Real, but an order of magnitude below what this removes, so it did not justify vendoring an encoder here.

---
_Generated by [Claude Code](https://claude.ai/code/session_01X9sVubw54TbWgrfHy5YgSE)_


src/transport/http/decodeImageV2.ts
```diff
@@ -264,9 +287,10 @@ export function decodeImageV2(data: Buffer): Buffer | null {
   let best: { spread: number; subsampling: number; width: number; height: number } | null = null;
   for (const subsampling of [2, 0, 1]) {
     let filled: { width: number; height: number; img: Decoded } | null = null;
-    for (const [width, height] of LADDER) {
+    for (const [width, height] of ladderByMcuCount(subsampling)) {
       const img = decodeCandidate(tail, width, height, subsampling);
-      if (img && (!filled || width * height > filled.width * filled.height)) filled = { width, height, img };
+      if (!img) break;
```
lenoxys, line 292: The monotonicity is over MCU count, but the thing you `break` on is `decodeCandidate` returning null — and null is the observable, not "the scan ran out". If `jpeg-js` ever fails a mid-ladder rung for some other reason, the walk stops early and `decodeImageV2` quietly returns a smaller geometry than it used to. The old code was immune to that by construction: it tried everything and kept the largest that filled.

Byte-identical output on both fixtures is good evidence, but both are small and synthetic, and your own deferred note names the gap — the 632x472 4:4:4 thumbnail is the geometry this decoder was originally cracked against and it hasn't been run. Its MCU count is far below anything the ordering skips, so I'd expect it to be fine; it's the one run I'd want before relying on this.

Not blocking, and I wouldn't add a failure-mode discriminator on speculation — the cheap version is just that one fixture.

(Ties are fine, for what it's worth: equal MCU counts consume the same scan, so they succeed or fail together, and V8's stable sort keeps ladder order among them.)

lenoxys (approved):
This is the model the other open PRs should be measured against. The change makes the code do strictly less — a helper, a `break`, and a rewritten cost note — and the justification is a property of the format rather than a benchmark artefact: the scan carries a fixed number of MCUs, so failure is monotone in MCU count and the first failure is the last candidate worth building. Nothing about which geometry wins changes; only which candidates get built.

Nothing to cut on the over-engineering axis. Recomputing the sort once per subsampling is three sorts of a tiny array — memoizing it would be the over-engineered version.

**The spec is the part I'd point other PRs at.** It asserts the cost, which the reconstruction specs structurally cannot see (they'd pass just as happily if the answer were reached the expensive way), it counts decoder calls instead of timing a shared CI host, its bounds are an order of magnitude loose so a new ladder rung won't break it, and you confirmed it fails when the source is reverted. That last step is the one people skip.

Approving. One question below that I don't think should hold it up.

PR #134, opened 2026-09-09
feat(camera): notification style, and 6020's real meaning

lenoxys:
## What

Cameras gain `notificationStyle` — how much a detection push carries: text alone, a thumbnail, or text then a thumbnail. Read AND write, both confirmed.

```ts
dev.camera()?.setNotificationStyle(NotificationStyle.IncludedThumbnail);
```

## Why 6020 was wrong before

The app's own id -> name table carries **two** aliases for 6020, `ENTER_OTA` and `INDOOR_PUSH_NOTIFY_TYPE`, and the two halves of this repo disagreed about which wins: `param-dictionary.ts` said `enterOta`, while the generated `transport/p2p/commands.ts` already said `CMD_INDOOR_PUSH_NOTIFY_TYPE`. A capture settles it — on a camera this id is the notification style, so the dictionary entry is repointed and its provenance raised to `verified`.

## Evidence

A standalone outdoor camera (T8171 — its own station, channel 0, reached over the cloud/WAN path rather than LAN-direct):

```
outer P2P cmd 1700 (CONTROL_PAYLOAD), signCode 8, AES-256-GCM
{"commandType":6020,"data":{"value":N,"transaction":"<epoch ms>"}}
```

Three taps through the app's picker. Every frame decrypted with that station's own level-2 key, derived from the `GATEWAYINFO` reply present in the same capture; every value read back on the cloud parameter within ~10s.

| Option | `value` |
| --- | --- |
| Text Only | 1 |
| Included Thumbnail | 2 |
| Text First, Then Thumbnail | 3 |

Independently, this is the mapping `docs/cloud/push.md` recorded as `1 TEXT / 2 THUMB / 3 ALL` but had never confirmed on a device.

### How the parameter was identified

Not by decryption. By diffing the cloud parameter record either side of each tap: `update_time` moves only when a parameter is written, so it names the parameter with no key involved, and the decryption then only had to confirm the frame's shape.

That distinction mattered, because only 6020 moved on **all three** taps — `wifiRssi` and battery telemetry moved on some of them, and drift like that is exactly what a single before/after comparison would have mistaken for the control.

## Decisions worth challenging

- **`transaction` is sent.** It is in the frame that was verified; a bare `{value}` body is untested, and `light`'s 1700 writes omitting it are not evidence for this command. Dropping it would ship a frame nobody has seen work.
- **A value outside the enum is rejected, not clamped.** Every neighbouring value is itself a real style, so a coerced one selects the wrong notification and reports success.
- **The reply is acknowledged but not claimed.** The station answers on the same wrapper ~600ms later, so this write is not fire-and-forget — but the reply's shape is not asserted anywhere and nothing reads it yet.

## Not covered

- Whether 6020 also serves `ENTER_OTA` on some other model or generation. Only the camera meaning is confirmed.
- The **LAN-direct** path. This was verified over cloud/WAN only, and those are genuinely different code paths.

## Verification

`npm run verify` green — 3381 tests, all guards, build, ESM check, examples typecheck.


lenoxys:
## Two cross-checks from a second capture

Mapping an unrelated setting on the same camera (`battery.workingMode`, param 1246) put the frames of both commands side by side in one capture, which sharpens two claims here.

**The plaintext length matches the JSON exactly — nothing hidden.**

| | outbound | reply |
| --- | --- | --- |
| 1700, notification style | 101 B wire -> **69 B** plaintext (the JSON) | 36 B -> **4 B** |
| 1246, working mode | 168 B wire -> **136 B** plaintext (direct-binary struct) | 36 B -> **4 B** |

69 B is precisely `{"data":{"value":N,"transaction":"<13 digits>"},"commandType":6020}`. So the decrypted body carried no field beyond `value` and `transaction` — the frame this PR ships is the whole frame, not a subset of one.

**The reply is a 4-byte body, and it is not specific to this command.** The same 36 B -> 4 B shape came back on 1246, whose write wire this repo already calls verified. So the ack is how control writes answer generally, which is worth knowing given how much of the codebase is built around writes that cannot be confirmed. The JSDoc now states the observed length and nothing more — the 4 bytes were measured, not decrypted, and nothing reads them.

Reading that result code would let the SDK confirm a control write instead of firing and forgetting. That is a transport-level change well outside this PR, and it deserves its own issue.

**Also worth recording as a negative result:** 1246 on this model needed no change at all. `WORKING_MODE_MAPS.DEFAULT` already says `{0: Optimal Battery Life, 1: Optimal Surveillance, 2: Customize Recording}`, and tapping all three options through the app produced exactly 0 / 1 / 2. The existing map was right.

martijnpoppen (approved):

lenoxys:
## HW-tested, and the test nearly reported the opposite

Replayed `setNotificationStyle` from this SDK against a standalone T8171. **It lands.** The frame goes out sign-8 on the negotiated level-2 key, LAN-direct, and the setting changes.

Getting there corrected two things I had wrong, both worth recording because they will mislead the next person the same way.

### The cloud parameter is a weak oracle for a control write

My first two runs reported **FAILED**. They polled the cloud parameter for 100s and it had not moved.

It moved at **321 seconds**.

| | write → cloud reflects |
| --- | --- |
| the app writing 6020 | seconds |
| this SDK writing 6020 | ~5 minutes |

The app holds its P2P session open; a one-shot write closes it, and a battery camera then syncs on its own schedule. So a read taken shortly after reports the OLD value, and treating that as failure mistakes a landed write for a dropped frame — which is exactly what I did, twice, and then went hunting for a signCode bug that was not there.

### The timely confirmation is the reply, not the cloud

The station answers within about a second — the 4-byte int32 result. That is what the app acts on, and it explains the behaviour that started this: when a broken certificate killed the session, the app's picker **timed out and reverted**. It was waiting on that reply, never on a cloud read.

On this model that reply does not open under the session decrypt, so #130's `commandResult` does not fire for it and the code stays unread. Reported there separately; the app clearly reads it on this same camera, so it is openable in principle.

### One conclusion I withdrew

An intermediate diagnosis was that the transport sends level-1 to a standalone camera while the app used level-2. The debug trace disproves it: this standalone camera **does** negotiate a level-2 key (`cipher_id 196`) and `sendBySessionLevel` picks L2 off `hasLevel2Key`, not off topology.

Worth flagging anyway: `command-router.ts` still states "A standalone camera never negotiates a level-2 key" as the reason `form: "auto"` exists. That claim is false for this camera. It is not this PR's to fix, but something downstream is reasoning from it.

### Also caught by an earlier capture

The outbound plaintext is 69 B, which is exactly `{"commandType":6020,"data":{"value":N,"transaction":"<13 digits>"}}` — so the frame shipped here is the whole frame, with no field the decrypt might have hidden. The SDK's own send measures 69 B too.

PR #135, opened 2026-09-09
fix(camera): name the top video-quality tier for its rank, not for a resolution

lenoxys:
## The bug

`resolveVideoQuality(3)` answered `"3K HD"` for every camera. On a 2K camera that is a resolution the device does not have, and it is the *only* name the SDK offered for that camera's top tier.

> The branch name says `per-model`, which is where this started. The second commit throws that away — see below. The name is left alone rather than churned.

## Why not a per-model label map

That was the first attempt, and the comment above `VIDEO_QUALITY_TIERS` even prescribed it. Three things say it is wrong:

**1. The resolution is not discoverable.** A camera does not report its top resolution — the cloud record's `product` is `null` and no parameter carries it. The app's own model registry spells a resolution into only some older families, and neither confirmed model is among them:

| model | app family | resolution in the name |
| --- | --- | --- |
| T8170 | `BATTERY_SOLO_CAM_8170` | no |
| T8425 | `FLOODLIGHT_8425` | no |
| T8171 | *absent from the registry* | — |

So a label map is hand-maintained per model, indefinitely, with no source to check it against.

**2. The app is not consistent with itself.** On one T8171, the SAME tier 3 of the SAME sensor is:

- `2K HD` in the **Recording Quality** picker
- `2K` in the **Streaming Quality** picker

There is no single label to be right about.

**3. A resolution label is presentation.** The repo's rule is that presentation belongs to the caller. The tier's RANK is the device fact; what to call the pixels is not.

## What this does instead

`VideoQuality.Max` — tier 3 named for its rank. The two lower tiers keep their resolutions, which are 720P and 1080P on every camera confirmed so far.

No table, no model parameter on any resolver, and one name that is valid on every camera — so the name type does not have to span models. (The map version needed that and did not have it: `"2K HD"` was not assignable to `VideoQualityName`, so a 2K camera's only valid name could not be typed. That defect is gone with the map.)

## Evidence

The same setting mapped through the app's picker on two standalone cameras, each option read back on the cloud parameter after the tap:

| model | tier 1 | tier 2 | tier 3 |
| --- | --- | --- | --- |
| T8170 | HD (720P) | Full HD (1080P) | 3K HD |
| T8171 | HD (720P) | Full HD (1080P) | 2K HD |

Tier **numbering** is shared; the top tier's resolution is the sensor's.

## Behaviour change, called out deliberately

`setVideoQuality("3K HD")` no longer resolves. `Max` replaces it on every camera — and the old name was wrong on half the confirmed ones, while still landing on a real tier, so nothing looked broken at any layer.

## Deliberately not done

Both cameras also write a flat mirror of the active tier to param **1023** (`videoRecordQuality`) — an integer where 2731 is a base64 JSON blob, so it looks tempting as a read alias. The meaning already has an owner in `camera.videoQuality`, and an alias only earns its place if some device reports 1023 *without* 2731. Neither confirmed camera does.

Also untouched: the write hardcodes `mode: 0` while the read lifts whichever mode `cur_mode` names. Both cameras reported `cur_mode: 0` throughout, so there is no evidence either way, and guessing the field is what the repo forbids.

## Verification

`npm run verify` green — 3364 tests, all guards, build, ESM check, examples typecheck.


lenoxys:
Folded into #136.

Splitting these was a mistake. This PR renamed the top tier on `VideoQuality`; #136 renames `VideoQuality` itself — so reviewing this one meant reviewing a symbol that is already obsolete in the stack, and a consumer would take two breaking renames of one symbol instead of one.

The `Max` decision also depends on evidence from #136's capture run: the same T8171 calls the same tier of the same sensor `2K HD` in the Recording picker and `2K` in the Streaming picker. That is what rules out chasing the app's label, and it is not visible from this PR alone.

Both commits are in #136's history, including the per-model map and the commit that deletes it — that trail is worth keeping.

PR #136, opened 2026-09-09
feat(camera): the two quality settings, told apart — live view and recording

lenoxys:
> Supersedes and absorbs #135. They were split, which was wrong: #135 renamed the top tier on `VideoQuality`, this renamed `VideoQuality` itself, so reviewing the first meant reviewing an already-obsolete symbol — and it inflicted two breaking renames of one symbol instead of one. The `Max` decision is also justified *by the evidence gathered here*, so it never stood alone.

## The trap

A camera has **two** quality settings, and the SDK modelled one of them under a name that did not say which:

| app picker | sub-command | app's own name | mirror param |
| --- | --- | --- | --- |
| **Streaming Quality** | 2730 | `multicamSetVideoQuailty` | 1020 |
| **Recording Quality** | 2731 | `multicamSetRecordQuailty` | 1023 |

The vendor's names invert what they suggest — "Video" is the live view, "Record" is what gets stored. `camera.videoQuality` read 2731, so it was the *recording* quality, and the JSDoc on that id literally read "Video / streaming quality": it was describing the other setting.

Adding the live-view one next to a `videoQuality` that means recording would have been permanent. The pair is now `recordingQuality` and `streamingQuality`.

They are separate wires, established by observation and not by name: changing Streaming Quality moved **1020 and 2730** and left 1023 and 2731 untouched. Changing Recording Quality did the reverse.

## The top tier is named for its rank, not a resolution

`resolveVideoQuality(3)` answered `"3K HD"` for every camera. On a 2K camera that is a resolution the device does not have, and it was the only name offered for that camera's top tier.

A per-model label map was the first attempt — the comment in the source even prescribed one. It is in this branch's history and then deleted, because three things say it is wrong:

**1. The resolution is not discoverable.** The cloud record's `product` is `null` and no parameter carries it. The app's own model registry spells a resolution into only some older families, and neither confirmed model is among them:

| model | app family | resolution in the name |
| --- | --- | --- |
| T8170 | `BATTERY_SOLO_CAM_8170` | no |
| T8425 | `FLOODLIGHT_8425` | no |
| T8171 | *absent from the registry* | — |

**2. The app is not consistent with itself.** On one T8171, the SAME tier 3 of the SAME sensor is `2K HD` in the Recording picker and `2K` in the Streaming picker. There is no single label to be right about — and that is evidence only this branch's second capture run could produce.

**3. A resolution label is presentation**, which the repo's rules give to the caller. The tier's rank is the device fact.

So: `Max`. No table, no model parameter on any resolver, one name valid on every camera. The two lower tiers keep their resolutions, which are 720P and 1080P everywhere confirmed.

## Evidence

Every option tapped through the app's picker on two standalone cameras, each read back on the cloud parameter.

**Streaming** (0-3):

| tier | T8170 | T8171 |
| --- | --- | --- |
| 0 | Auto | Auto |
| 1 | HD (720P) | HD (720p) |
| 2 | Full HD (1080p) | Full HD (1080p) |
| 3 | 3K HD | 2K |

**Recording** (1-3): identical tiers, `3K HD` on T8170 and `2K HD` on T8171 — no tier 0.

The streaming frame, decrypted from a T8170 capture, all four values byte-exact:

```
outer cmd 1350 (SET_PAYLOAD), signCode 8
{"cmd":2730,"payload":{"transaction":"<epoch ms>","quality":N,
 "channel":0,"mode":0,"primary_view":0},"account_id":…}
```

The station echoes the same payload back as a `1351` NOTIFY_PAYLOAD carrying the same `transaction`.

## Tier 0 is the difference that matters

Recording has no `Auto`. The two domains are **not** interchangeable: `0` is a real tier on one wire and out of range on the other, and a test pins that. Sharing one tier set would have made `setRecordingQuality(0)` look valid.

## The read is 1020, not 2730

2730 carries a per-mode config; 1020 carries the tier in force, which is what the property means.

It is also the name the param dictionary already gives 1020 — and `Device` keys state by **name**, so a member claiming 2730 under this name would have collided with it on a device reporting both. That fails silently (`undefined` forever), not loudly.

## Breaking changes, called out deliberately

- `camera.videoQuality` / `setVideoQuality` → `recordingQuality` / `setRecordingQuality`, and the same for `VideoQuality`, `VIDEO_QUALITY_TIERS` and the three resolvers.
- `setRecordingQuality("3K HD")` no longer resolves; `Max` replaces it on every camera. The old name was wrong on half the confirmed ones while still landing on a real tier, so nothing looked broken at any layer.

Below 1.0 the repo's rule is to move the shape rather than keep a parallel one, so there are no deprecated aliases.

## Worth challenging

- **`transaction` is sent on 2730 but not on 2731.** Each matches its own captured frame; neither is inferred from the other.
- **The envelope adds `mChannel`/`mValue3`, which the captured app frame does not carry.** `setPayload` always emits them and the recording sibling's write was confirmed live that way, so the family accepts them — but that is the sibling's evidence, not this command's.
- The 2731 write still hardcodes `mode: 0` while the read is the resolved tier. Both cameras reported `cur_mode: 0` throughout, so there is no evidence either way, and guessing is what the repo forbids.
- **Nothing here is HW-tested from the SDK.** The frames come from the app. Read paths are confirmed on both cameras.

## Deliberately not done

Param **1023** (`videoRecordQuality`), recording's flat mirror, stays unmodelled: the meaning already has an owner, and a read alias only earns its place if some device reports 1023 without 2731. Neither confirmed camera does.

## Verification

`npm run verify` green — 3381 tests, all guards, build, ESM check, examples typecheck.


martijnpoppen (approved):

lenoxys:
## HW test: the write is withdrawn, the read stays

Replayed the captured 2730 frame from this SDK against a standalone T8171. **Nothing changed on the device**, over 40 minutes. So the setter is gone: `unverified: true`, no `write`, the intent path throws, and the read ships ahead of it.

### Why this is not "the SDK cannot reach the camera"

A control experiment in the same session, same camera, same envelope:

| write | sub-command | result |
| --- | --- | --- |
| `recordingQuality` | 2731 | **landed**, device reported it back |
| `streamingQuality` | 2730 | no change, ever |

So the session, the level-2 key, the channel and the envelope are all fine. Something about 2730 specifically is not.

### What I ruled out, and what I did not

**Ruled out: `transaction`.** It was the obvious suspect — the two failing writes carried it and the working one did not. Removed it and re-tested: 2730 still did nothing. So it is not the cause, and it is no longer sent.

**Not ruled out: `mChannel`/`mValue3`.** The app's captured 2730 frame carries neither, and `setPayload` always emits them. The recording sibling's confirmed write emits them too and works, so the family tolerates them — but that is 2731's evidence, not 2730's. Testing this needs transport plumbing to omit them, which is out of scope here.

### The honest verdict is weaker than "it fails"

The cloud parameter is a poor oracle for a control write. On this camera it lagged a **confirmed** 6020 write by ~5 minutes (see #134), so 40 minutes of no movement is suggestive but not proof. The reliable confirmation is the reply the station sends within a second, and that reply does not open under the session decrypt on this model.

So the truthful position is not "this write is broken" but **"there is currently no reliable way to tell whether it lands"** — which is exactly what the unverified rule is for. Read paths may ship ahead of writes, and verification is per direction.

### Unchanged by any of this

Everything the PR establishes about the *distinction* stands, and none of it depended on our own send: the two settings are separate wires with separate parameters and separate domains, and the `Max` naming rests on the app's own labels. Only our ability to drive 2730 is in question.

PR #137, opened 2026-09-09
feat(doorbell): live-view quality and video encoding format, both inside 1705

lenoxys:
## What

A doorbell exposes two menus the SDK modelled neither of — **Streaming Quality** (Auto/Low/Medium/High) and **Video Encoding Format** (high/low compression). They turn out to be *one* parameter.

```
1705 = quality + (highCompression ? 5 : 0)      quality: 0 Auto, 1 Low, 2 Medium, 3 High
```

| | low compression | high compression |
| --- | --- | --- |
| Auto | 0 | **5** |
| Low | **1** | **6** |
| Medium | 2 | **7** |
| High | **3** | **8** |

Bold = written from the app and read back off the parameter. The plain two follow the same arithmetic and were **not** individually written; the JSDoc says so rather than implying otherwise.

The model earned some trust on the way: after Low=6 / Medium=7 / High=8 / Auto=5 it predicted that switching the encoding format would move High from 8 to **3**, then that Low on low compression would be **1**. Both held before being observed.

## Wire

Captured byte-exact on a T8210 behind a HomeBase:

```
outer P2P cmd 1705 itself — no wrapper — signCode 8, on the doorbell's channel
[u32 channel][u32 value][account_id ASCII pad 128]        = 136 B
```

The same direct-binary struct its **1703** and **1704** siblings already use in this file, both `verified` here. The station answers with an unencrypted level-1 frame carrying `0`.

## Why this is not two ordinary members

Two settings in one integer breaks the usual shape three ways, and each is handled rather than worked around:

- **`videoQuality`** owns 1705 and lifts its half out with `decode`. Its setter is `writtenElsewhere`, because rebuilding the integer needs the *other* half's current value and a member's `write` receives only the new value and the context.
- **`highCompressionEncoding`** declares no param and reads the owner's value. Two members claiming 1705 would leave the second reading a name nothing is stored under — `undefined` forever, silently. A paramless member publishes no schema entry, which prevents that structurally rather than by convention.
- **Both setters are `method`s**, which do get the state reader, and each preserves the half it does not own. Tests pin exactly that: setting quality on `8` sends `6`, setting low compression on `8` sends `3`.

## One validator, shared

Both halves go through the same guard so they cannot disagree about what a value means. It rejects:

- **4 and 9** — in range, but outside the encoding. Reading 9 as `9 % 5` would answer a quality of 4, which is not one.
- **`null` and `""`** — `Number()` maps both to 0, and 0 is a real setting here (Auto, low compression), so an absent reading must not decode as one. The type is checked before converting.

## Not done

**Not HW-tested.** The frame is the app's, not ours. Worth saying plainly given what happened elsewhere today: on a *standalone* camera I could not confirm a send at all, because the cloud parameter lagged a landed write by minutes and the reply that would settle it in a second does not decrypt there.

This doorbell is the better case — it is behind a HomeBase and its reply comes back **unencrypted**, so a send from the SDK can actually be judged here. That test is the obvious follow-up and I would rather it be a separate, deliberate step than a claim in this PR.

## Verification

`npm run verify` green — 3383 tests, all guards, build, ESM check, examples typecheck.


lenoxys:
## HW-tested — and the test rejected the first design

Replayed from this SDK against the T8210. The write works. The **shape** did not, and the device is what caught it.

### What broke

The first shape took one setting and preserved the other by reading it back. Two writes in one session:

| step | expected | actual |
| --- | --- | --- |
| `setVideoQuality(Low)` on 8 | 6 | **6** ✓ |
| `setHighCompressionEncoding(false)` | 1 | **3** ✗ |

The state reader is the snapshot the device was bound with. So the second call composed against the value from *before* the first — it re-read `8`, took `8 % 5 = 3`, and silently reverted the quality to High while appearing to do exactly what was asked.

No unit test would have found that. The reader was faithfully mocked in the spec, and the spec passed.

### The fix is a smaller API, not a smarter one

The wire cannot set one half alone — every frame carries the whole integer — so `setVideoQuality(quality, highCompression)` takes both. `setHighCompressionEncoding` is gone.

That removes the stale copy instead of refreshing it, and puts the choice where it can be made correctly: a caller reads both properties and passes what it wants, rather than the SDK guessing from a copy it has no way to trust. Requiring both arguments is the API telling the truth about the wire.

### Re-tested, same session, three writes

```
setVideoQuality(Low, high)   -> 1705 = 6    PASS
setVideoQuality(Low, low)    -> 1705 = 1    PASS   <- the step that broke before
restore(High, high)          -> 1705 = 8    PASS
```

Each read back off the parameter, and the device is left as it was found. The spec now pins that middle step specifically.

### Status

This is the only write in today's batch that is confirmed end to end from the SDK. #134 and #136 both ship their writes as `unverified` for the opposite reason — on a standalone camera the cloud parameter lagged a landed write by minutes and the reply that would settle it does not decrypt, so nothing there could be judged. This doorbell is behind a HomeBase and answers in clear, which is what made the difference.

max246 (approved):

PR #139, opened 2026-09-10
fix(ptz): correct the property table — two properties go, one moves to the id the app writes

lenoxys:
The `ptz` property table claimed three things, and all three were wrong. This corrects the table: two properties go, one moves.

Every wire below was observed on an indoor pan-tilt camera (deviceType 31, mains, standalone) — one cloud snapshot per app tap, the frames captured byte-exact in the same run.

## `panAngle` (1206) and `tiltAngle` (1207) — removed

Both were `provenance: "guessed"`, and both ids are already claimed:

| property | id | what the id actually is |
| --- | --- | --- |
| `panAngle` | 1206 | `CMD_SET_BITRATE` |
| `tiltAngle` | 1207 | `CMD_SET_MIRRORMODE` — the live-confirmed `camera.imageFlipped` write |

1207 was the visible bug: on a pan-tilt camera, flipping the image read back as a tilt angle. It was tracked as `KNOWN_CROSS_OWNER` debt in `property-id-integrity.spec.ts`.

**Measured, with a positive control.** The camera reports 1207 but **not** 1206 — a camera whose tilt were 1207 and pan 1206 would report both — and its 1207 read `"0"`, last written **894 days ago**, on a camera panned and tilted many times since.

- **Phase A, many pan and tilt moves from the app:** the only parameters that moved on that device were `1135` (tfcard status, same value), `1224` (arming mode, same value), `2001` (a camera enable, ours), `6053` (an unnamed bool that flipped when the camera woke). No parameter appeared. **1207 did not move** — same value, same 894-day-old timestamp.
- **Phase B, the positive control — toggle the 180° image rotation:** 1207 went `"0"` → `"1"` within ~30 s, and it was the **only** parameter on the device that moved. Restored and confirmed in the app.

There is no parameter to repoint these two to. Position is not reported as a parameter by any model: it arrives only while the camera moves, as a `CMD_DOORBELL_SET_PAYLOAD` (1700) binary frame of float `(pan, tilt)` records — which `PTZ.decodeEvent` already turns into `ptzNotify` `kind: "position"`. The two properties were a second, wrong answer to a question the event already answers.

## `rotationSpeed` — moved from 1018 to 6015

1018 is `CMD_CLOSE_AUDDEC`. The real parameter is **6015**, found by walking the app's own Slow/Mid/Fast control:

| app option | wire | value |
| --- | --- | --- |
| Slow | `{"commandType":6015,"data":{"value":1}}` | `1` |
| Mid | " | `3` |
| Fast | " | `5` |

6015 was the only parameter that moved on each tap. Then **confirmed through this member**: 1 and 5 sent consecutively in one session and each read back on the cloud param, restored to 3. Two consecutive writes because a setter composing its payload against bind-time state passes the first and reverts on the second — a unit test with a mocked reader cannot see that.

The finding worth keeping: **6015 was absent from the device's parameter list until the first write.** It materialises when the setting is first used. That is why an earlier fleet probe found no device reporting a rotation speed — the feature was never missing, it had simply never been configured. Generally: "no device reports X" is not evidence that X is unsupported.

2 and 4 are accepted but undescribed. The endpoints and the midpoint are observed, the wire is not known to reject the gaps, and a `scalar` says that honestly where an enum of three would claim the scale has three positions.

## Result

- `PTZ.properties` is `["rotationSpeed"]` — one real property, on a verified id, with a confirmed setter.
- `KNOWN_CROSS_OWNER` is **gone** from `property-id-integrity.spec.ts`: the one-owner rule now holds with no cross-owner exemption at all, so the next collision fails the build instead of joining a list.
- Two entries off the ungrounded-id list (`KNOWN_UNLISTED`).

## Known wart, not introduced here

The param dictionary calls **6015 `lockParam`**, beside `6010 pwList` / `6011 statusInServer` / `6012 statusInLock` / `6014 updatePw` — the smart lock's block. The id is reused across product lines and the camera's meaning is the observed one. The one-owner rule keys on `${line}:${param}` so nothing collides, but a flat id → name map cannot express an id meaning two things in two lines.

Also observed: the app's **Pan-tilt Calibration** button wrote no parameter at all — P2P-only, so a `calibrate()` method needs the frame, not the cloud record.

`npm run verify` green: 3390 tests.


lenoxys:
## Confirmed on hardware — the pan-tilt camera itself

The PR argued from the command table plus a prior capture. Ran the experiment live on the Indoor PT camera (deviceType 31), on the cloud parameter record, with a positive control.

**Baseline.** The camera reports `1207` but **not** `1206` and **not** `1018`. A camera whose tilt angle were 1207 and pan angle 1206 would report both. Its `1207` read `"0"`, last written **894 days ago** — on a camera that has been panned and tilted many times since.

**Phase A — many pan and tilt moves from the app's live view.** Every parameter on that device whose `update_time` moved: `1135` (tfcard status, same value), `1224` (arming mode, same value), `2001` (the enable, mine), `6053` (an unnamed bool that flipped when the camera woke from a 2-day power-off). No parameter appeared. **`1207` did not move — same value, same 894-day-old timestamp.** Nothing angle-shaped moved anywhere.

**Phase B — the positive control: toggle the 180° image rotation.** `1207` went `"0"` → `"1"`, written within ~30 s, and it was the **only** parameter on the device that moved. Restored to `"0"` afterwards and confirmed in the app.

So 1207 is the image flip, measured on the one device class where a tilt angle could have existed — fixed cameras have no angle at all. There was nothing to repoint these three properties to, and now that is observed rather than derived.

### Side finding, not part of this PR

That mirror latency is on a camera with **no HomeBase**. The working assumption has been "a write is only confirmable on a HomeBase-attached device"; this device is standalone and mirrored in ~30 s. The real axis looks like **mains vs battery**, not attached vs standalone — this one is standalone and mains. Tracking separately; it does not affect anything here, since this PR writes nothing.

lenoxys:
## Correction — one claim in the PR body is now wrong

The body says *"There is nothing to repoint them to"*. That holds for `panAngle` and `tiltAngle`, and it is now **wrong for `rotationSpeed`**.

A later hunt on the same camera walked the app's **Pan-tilt Rotate Speed** control, one snapshot per tap:

| app option | parameter written | value |
| --- | --- | --- |
| Slow | `6015` | `1` |
| Fast | `6015` | `5` |
| Mid | `6015` | `3` |

`6015` was the **only** parameter that moved on that device for each tap, and — the part worth keeping — it was **absent from the device's parameter list before the first tap**. It materialises on first write. That is why the earlier fleet probe found no device reporting a rotation speed: the feature exists, it had simply never been set on any device in the account.

**This PR does not change.** `1018` is still not the rotation speed — the app writes `6015` — so shipping a typed getter on `1018` was wrong in the way this PR describes, and it still goes. Repointing is a different change with its own evidence: a real id, an observed 1..5 scale, and a write that can be confirmed (this camera is mains, so its mirror answers in ~30 s). It gets its own PR rather than being smuggled into a subtraction.

One thing that PR will have to settle first: the param dictionary calls `6015` **`lockParam`**, next to `6010 pwList`, `6011 statusInServer`, `6012 statusInLock`, `6014 updatePw` — the smart lock's parameter block. So `6015` is an id **reused across product lines**, and the camera's meaning is the observed one. The one-owner rule keys on `${line}:${param}`, so the two can coexist, but the dictionary's flat id → name map cannot express it.

Also observed, for whoever writes that PR: the app's **Pan-tilt Calibration** button wrote **no parameter at all** — it is P2P-only, so a `calibrate()` method needs the frame, not the cloud record.

max246 (approved):

PR #140, opened 2026-09-10
feat: sound detection, and pan-tilt rotate speed on the id the app writes

lenoxys:
Stacked on #139 — that PR removes three properties pointing at ids the app uses for something else; this one adds back the **one** of them that turned out to exist, on its real parameter, plus the three sound-detection settings found in the same session. Merge #139 first, or squash the two together.

Every wire here was identified by walking the app's own control with one cloud snapshot per tap, captured byte-exact in the same run, and then **written through these members** on a T8410 (mains, standalone, own channel) and read back.

## `ptz.rotationSpeed` → 6015

| app option | wire | value |
| --- | --- | --- |
| Slow | `{"commandType":6015,"data":{"value":1}}` | `1` |
| Mid | " | `3` |
| Fast | " | `5` |

6015 was the only parameter that moved on each tap. The part worth keeping: **6015 is absent from the device's parameter list until the setting is written once.** That is why an earlier fleet probe reported no device carrying a rotation speed — the feature was never missing, it had simply never been configured on any device in the account. It also means "no device reports X" is never evidence that X is unsupported.

## `camera.soundDetection` / `soundDetectionSensitivity` / `soundDetectionType` → 6043 / 6044 / 6046

The app shows sound detection beside motion; the SDK modelled none of it.

| param | payload | values |
| --- | --- | --- |
| `6043` | `{"status": 0\|1}` | off / on |
| `6044` | `{"index": 1..5}` | 1 lowest, 3 mid, 5 highest |
| `6046` | `{"type": 1\|2}` | 1 crying only, 2 any sound |

Each command names its payload field differently — `status`, `index`, `type` — so there is no shared setter for the family. Not an envelope detail; it is per command.

Three members on `camera` rather than a new capability: they are evidence-gated either way, so a device that does not report 6043 shows none of them, and a module of its own would have bought a barrel entry plus a detection block for no behavioural difference. They sit beside `notificationStyle`, the same 60xx indoor-camera family on the same wrapper.

## Write confirmation

12 writes, each through the bound surface (`dev.camera().setX()` / `dev.ptz().setX()`), each read back on the cloud param, device restored to what it was found on:

```
soundDetection             false → true → false   ✓ ✓ ✓
soundDetectionSensitivity  1 → 5 → 3              ✓ ✓ ✓
soundDetectionType         1 → 2 → 1              ✓ ✓ ✓
rotationSpeed              1 → 5 → 3              ✓ ✓ ✓
```

**Two consecutive writes per member on purpose:** a setter that composed its payload against state read at bind time passes a single write and silently reverts on the second, and a unit test with a mocked reader cannot see it. None did.

Judged on a **mains** device deliberately. Its mirror answers in ~30 s — that is a property of power, not of having a HomeBase, and every device reporting these four settings is either standalone or battery. The tool is on the workshop branch (`scripts/observe/confirm-write.mjs`).

## Where the types stop short of the values

`soundDetectionSensitivity` is a `scalar` in 1..5, not an enum of three. The app exposes three positions and those three are what was measured; 2 and 4 are not known to be rejected, so the type says "a number in 1..5" and the description names the three that were observed. `soundDetectionType` **is** an enum, and a value outside it is refused rather than coerced — both neighbouring values are real triggers, so arming the wrong one and reporting success is worse than an error.

## Known wart, not introduced here

The shared param dictionary calls **6015 `lockParam`**, beside `6010 pwList` / `6011 statusInServer` / `6012 statusInLock` / `6014 updatePw` — the smart lock's block. The id is reused across product lines and the camera's meaning is the observed one. The one-owner rule keys on `${line}:${param}` so nothing collides, but the dictionary's flat id → name map cannot express an id that means two things in two lines.

`npm run verify` green: 3445 tests.


lenoxys:
Superseded — the split was chronological rather than by subject. `rotationSpeed` was deleted in #139 and re-added here, which is churn a reviewer sees for no reason: across the pair its net change is "1018 → 6015, plus a setter".

Re-cut by subject, and the two are now independent rather than stacked:

- **#139** — the whole `ptz` property-table correction: `panAngle`/`tiltAngle` removed, `rotationSpeed` moved 1018 → 6015. One story, and `rotationSpeed` is never deleted and re-added.
- **#141** — sound detection on `camera` (6043/6044/6046), based directly on `beta-0.1.0`. It shares no file with the ptz change, so it never needed to be stacked at all.

Same commits, same hardware evidence, nothing dropped.

PR #143, opened 2026-09-10
feat(solix): same-account login + read Anker Solix devices

max246:
## What

Adds a self-contained `src/solix/` module that lets **one Anker/eufy account log in and read its Anker Solix power-station devices** — the same account credentials the eufy client already uses.

Solix's cloud runs the **same `algo_ecdh` passport** as the `eufy_mega` account stack, re-skinned under `app-name: anker_power` with its own key-exchange bootstrap key and API host. So the login handshake reuses the existing crypto primitives verbatim; only a few constants differ.

## API

```ts
import { SolixClient } from "@mega-yfue/eufy-sdk";

const solix = new SolixClient({ email, password, countryCode: "GB" });
const r = await solix.login();                 // { status: "ok" | "2fa" }
if (r.status === "2fa") await solix.submitVerifyCode(code);

await solix.getDevices();        // account's bound Solix devices
await solix.getSites();          // sites/systems
await solix.getUserMqttInfo();   // per-user AWS-IoT cert/key/endpoint/thing
```

## How it works

- **Login (encrypted):** domain-estimate → localKey-bootstrapped ECDH key exchange → `/passport/login` with an encrypted, signed body. Reuses `prepareKeyExchange` / `encryptLoginPassword` / `signRequest` / `encryptBody` from `core/crypto`, passing the Solix `localKey`. 2FA follows the eufy passport's `fa_info.info` semantics.
- **Reads (plain):** authenticated resource calls are plain JSON carrying only `x-auth-token` + `gtoken = md5(user_id)` — no per-request encryption or signature. (This is the key detail: sending an algo_ecdh envelope on reads is what the server rejects.)

Constants (`app-name`, hosts, key-exchange `localKey`, endpoints) live in `src/solix/constants.ts`.

## Scope / non-goals

- Self-contained; **does not touch the eufy Device/capability model**. Returns the vendor's typed JSON rather than modelling Solix as capability objects.
- Real-time device state over the returned AWS-IoT MQTT credentials is **out of scope** here (this PR gets you authenticated cloud reads + the MQTT bootstrap).

## Validation

- Reversed from a live capture of the Solix app (system-cert mitm, no pinning); wire scheme cross-checked against the open-source `anker-solix` project.
- **Login proven end-to-end against the live cloud**; plain reads return `200` with the expected payloads (`get_user_mqtt_info` returns full per-user IoT credentials).
- Offline round-trip unit test with a synthetic server doing the real server side of the handshake (real crypto both directions) — no network, no account data.
- `npm run verify` passes locally: format, typecheck, all guards, build, seal, esm, examples, **3436 tests**.

https://claude.ai/code/session_0165pQMCLK7JLkqWfwRqShXy


max246:
Superseded by #153 — the whole Solix stack is combined into one PR. Closing this stacked PR.

PR #148, opened 2026-09-10
feat(camera): label recordingQuality tiers so hosts render a Video Quality dropdown

max246:
## What

Makes **Recording Quality** show up as a proper dropdown (720P / 1080P / 3K) in a host UI, matching **Streaming Quality** which already does.

`streamingQuality` already carries `enumValues` (so it classifies as a labelled choice / HA `select`). `recordingQuality` had only `decodedValues` (the tier set) with **no labels**, so `classify()` on the host side couldn't turn it into a dropdown. This adds `enumValues: RECORDING_QUALITY_TIERS` — the same shape as `streamingQuality` and `suction`.

`decode` still owns the value (the active tier lifted out of the composite `2731` config); `enumValues` only labels it, keyed to match `decodedValues`.

## Why (user report)

A tester had no dropdown for video / streaming quality in HA. Root cause: HACS builds a `select` only from a writable enum **property** carrying `enumValues`. Streaming quality already qualified; recording quality didn't. This closes that gap — **no HACS change needed**, the bridge just needs a rebuild against this SDK.

## Test

`npm run verify` passes except **2 pre-existing failures** in `action-specs.spec.ts` (`doorbell.videoQuality` setter reachability) that are present on `beta-0.1.0` before this change — they come from #137, not this PR, and aren't touched here.


lenoxys:
@max246 just need to fix CI/CD pipeline but lgtm

lenoxys (approved):
lgtm

PR #149, opened 2026-09-10
fix(doorbell): register setVideoQuality in the setter-reachability allowlist

max246:
## What

Fixes the 2 failing tests on `beta-0.1.0` in `action-specs.spec.ts` (suite *"published schema — every writable property has a reachable setter"*):
- `doorbell.videoQuality is settable, not writable-in-name-only`
- `names an elsewhere-setter only for a property that still needs one`

## Cause

PR #137 (`feat(doorbell): live-view quality … inside 1705`) correctly added `doorbell.videoQuality` as a **`writtenElsewhere: true`** property: both halves of param 1705 (live-view quality **and** encoding format) must be written together, so it's set through the real `setVideoQuality(quality, highCompression)` method rather than a per-property `write`. That's the intended pattern (same as `lock.locked` → `lock`, `locate.locating` → `locate`).

But the test's curated `SETTER_ELSEWHERE` allowlist — which names the setter for each `writtenElsewhere` property, so the guard can confirm it's reachable — was **not updated**. So the guard rightly flagged `doorbell.videoQuality` as having no reachable setter, and the allowlist as out of sync.

## Fix

Add `"doorbell.videoQuality": "setVideoQuality"` to `SETTER_ELSEWHERE`. The capability was correct; only the allowlist was stale. Test-only change.

## Effect

- Unblocks `beta-0.1.0` (back to green) and every PR based on it — notably #148 (video-quality dropdown), whose CI was red purely from inheriting these 2 failures.

## Test

`npm run verify` green — **3486 tests pass** (the 2 formerly-failing now pass, nothing else moved).


lenoxys (approved):
lgtm

PR #147, opened 2026-09-10
feat(solix): Solarbank grid-tie battery read capability

max246:
## What

Adds a `battery` capability to `SolixDevice` for the **grid-tie Solarbank / home-battery** family (A1790 = Solarbank E1600, A17C\* = Solarbank 2/3), with typed read accessors — the battery analogue of `energyMeter`, and the pair to the AE1X0 meter already shipped.

Accessors: `soc`, `batteryEnergy`, `batteryPower`, `chargePower`, `dischargePower`, `solarInputPower`, `outputPower`, `homeLoadPower`, `gridToBatteryPower`, `isCharging`, and the **temperature sensors** — `batteryTemperature`, `casingTemperature`, `bmsTemperature`, plus `temperatures()` which collects every temp-keyed channel (covers per-expansion-pack temps).

Detection: `SOLARBANK_MODELS` prefix match adds `battery` + `solarInput`, catalog-independent (like `SOLIX_METER_MODELS` for the meter).

## Grounding / honesty

Field **names** are the app's own, recovered from the Anker app's compiled-Dart strings (`libapp.so`, packages `charging/` + `ak_soft_ems/` — see `SOLIX_SOLARBANK_SPEC.md`). But unlike `energyMeter` — whose one confirmed tag came from a **live ff09 frame** — **no Solarbank telemetry frame has been captured yet**, so the wire→field binding is unconfirmed. Each accessor reads a set of candidate keys (JSON snake_case + Dart camelCase) and returns `undefined` until a real frame populates it. Documented on `SolixBattery`; tighten the candidate keys once a frame is captured or `blutter` recovers the parser.

## Scope

Read-only. The writable settings (output/discharge power limit, charge on/off, working mode, SoC bounds, backup reserve) are fully mapped in the spec but land in a **follow-up** — their exact payloads (`mode_type` enum ints, JSON schema) aren't literal in the binary and must be confirmed before writing to a real battery.

## Test

`npm run verify` green (all guards, build, **3451** tests). New specs cover product-code detection and the accessor/temperature reads.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #152, opened 2026-09-11
feat(solix): keep telemetry alive — arm reporting + heartbeat

max246:
## What

Fixes Solix realtime telemetry **stopping after a while** (and only resuming when you open the app).

A Solix device only *pushes* `param_info` while a client keeps **requesting** it. The app re-arms on every foreground resume plus a periodic heartbeat; our `SolixMqtt` only *subscribed*, so after the server's reporting window closed it went silent until the app re-armed it.

## How (captured live, mirrors the app exactly)

Captured the app's MQTT over the AWS-IoT plane (Zygisk `SSL_write` hook) — full protocol in `SOLIX_MQTT_ARMING.md`. It's **plain JSON**, not protobuf:

- **Arm** → publish `requestDeviceInfo` (`cmd:17`, `sign_code:1`) to `cmd/{app}/{pn}/{sn}/req`, whose `payload.data` is a base64 **ff09** request (request-type tag `a1=0x22`; an `info` and a richer `realtime` a2/a3 variant).
- **Heartbeat** → publish `cmd:10` to `dt/{app}/{user}/power_site`.
- Also subscribes to the full topic set (`basic_info`, `state_info`, `cmd/.../app/res`), not just `param_info`.

`SolixMqtt.watch()` now subscribes the full set, arms on connect, and starts a re-arm/heartbeat timer (`armIntervalMs`, default **25 s** — inside the ~30 s cadence / keepalive 60; set `0` for the old subscribe-only behaviour). Publishes are best-effort (errors emitted, not thrown; timer survives). `close()` clears the timer.

`buildFf09Request()` reproduces the two captured frames byte-for-byte — the XOR-checksum convention is **verified against the real captured `data` frames** in the tests.

## Test

`npm run verify` green — **3454 tests** (+3: the ff09 request builder + the captured-frame checksum convention).

Stacked on #147.


max246:
Superseded by #153 — the whole Solix stack is combined into one PR. Closing this stacked PR.

PR #155, opened 2026-09-11
release: 0.1.1

lenoxys:
Documentation only. `dist/` is identical to 0.1.0 — the version exists because npm renders a README from
the tarball, so correcting one requires publishing.

## What the README now states

**The example compiles against the published package.** `ptz` is the pan-tilt accessor, a
`PushSemanticEvent` carries `deviceSn`, and both an accessor and the method it returns are optional:
capabilities are evidence-gated, so a device exposes exactly the features it reported and the
optionality is how the type says one may be absent.

**Install carries the two commands, the Node floor, and that ffmpeg is optional.** Nothing else — the
dependency list and the provenance claim are out, because npm renders both on the package page next to
the install command, authoritatively.

**The architecture section is out.** `AGENTS.md` owns the layer graph and the decorrelation rule, and
`src/` is one flat public surface a consumer never navigates. A pointer replaces it.

134 lines → 110.

## Verified

- corrected example typechecked against `@mega-yfue/eufy-sdk@0.1.0` from npmjs, `--strict --module nodenext`
- `npm run verify` — 182 files, 3487 contracts, `ESM OK: 270 exports`
- `0.1.1-beta.1` published and installable

## Deferred

The README is the only TypeScript here no gate reads. A `guard:readme` extracting fenced `ts` blocks
would need a preamble convention for the identifiers a snippet leaves undeclared, which is a decision
about the gate rather than part of this release.


max246:
Do we have any conributions rules in place? how to do it?

lenoxys:
> Do we have any conributions rules in place? how to do it?

we do : https://github.com/mega-yfue/eufy-sdk/blob/main/CONTRIBUTING.md

max246 (approved):

PR #157, opened 2026-09-12
feat(arming): allow setting custom1 (mode_type 3)

holland-built:
## What

Adds `custom1` to `ArmingMode`, so `setMode("custom1")` is accepted instead of refused with `mode: 3 is not a valid value (must be one of 0/1/63)`.

## Why, and the evidence

I hit the refusal driving a HomeBase 3 from Home Assistant. My "don't alert" posture is a custom mode configured in the eufy app, and `home`/`away` can't reproduce it — so the mode is unreachable from the API even though the station sits in it happily.

`ARMING_MODE_WIRE` already maps `custom1 -> 3` and `armingCommand()` is mode-agnostic (it drops the wire value into `{mode_type, user_name}` on cmd 1224), so `ArmingMode` was the only gate.

I've read the doc comment explaining why the write domain is deliberately narrower — a fire-and-forget write that looks like success whatever the device does is exactly the failure mode worth refusing. So, to be precise about what I actually have:

- **Not** a byte-exact capture. I don't have the app traffic.
- **Live-confirmed** on a T8030, 2026-09-12: `setMode("custom1")` accepted, station reported `armingModeChanged` back within the convergence window. Done twice, cycling `custom1 -> home -> custom1`. Bridge log:

```
[bridge:event] push in: armingModeChanged sn=T8030P1323330EBA → broadcast to 1 frontend client(s)
[bridge:event] push in: armingModeChanged sn=T8030P1323330EBA → broadcast to 1 frontend client(s)
```

The post-write convergence check is what makes this meaningful rather than fire-and-forget: a mode the station ignored would surface as `StateConvergenceError`, not silent success.

I've documented it as live-confirmed rather than folding it into the captured set, so the distinction you drew stays visible. If you'd rather hold out for a real capture, that's a completely reasonable call — say so and I'll close this. I'm happy to collect a wire capture if you tell me what you need.

The other five (`schedule`, `custom2`, `custom3`, `off`, `geo`) stay refused. I only tested `custom1`.

## Changes

- `ArmingMode` gains `custom1`, placed in wire order so `SETTABLE_MODES` reads `0/1/3/63`
- Doc comments updated: `custom1` moves to a new live-confirmed bullet; counts corrected (six -> five uncaptured, three -> four settable)
- `arming.spec.ts`: `custom1` out of the refusal table, plus a positive case asserting `mode_type: 3` reaches the wire
- `manifest.spec.ts`: published action domain `[0, 1, 63]` -> `[0, 1, 3, 63]`

## Verification

- `npx vitest run` — 3487 passed, 182 files, 0 failed
- `npx tsc --noEmit` — clean

## Downstream

This surfaces through `ha-eufy-sdk-bridge` to the `ha-eufy-sdk` HA integration, where the select already offers all nine modes from `enumValues` and the four settable ones will now actually work.


holland-built:
Understood — re-opened against `beta-0.2.0` as #158.

Rebased rather than just moving the base: `beta-0.2.0` sits at release 0.1.0 while `main` is at 0.1.1, so retargeting this branch would have pulled `main`'s five extra commits into the diff. #158 carries the single commit. Tests re-run on that base — 3487 passing, `tsc --noEmit` clean.

Thanks for the clear redirect.

PR #158, opened 2026-09-12
feat(arming): allow setting custom1 (mode_type 3)

holland-built:
Re-targeted from #157, which I'd opened against `main` before reading the release flow. Rebased onto `beta-0.2.0` rather than moving the base, so this carries a single commit — `beta-0.2.0` is at 0.1.0 while `main` is at 0.1.1.

Updated after review: re-read `AGENTS.md` and corrected three violations in the original submission — a real station serial in the PR body, consumer/stack names in the PR body, and a JSDoc line pointing at this discussion instead of citing peers. Details at the bottom.

## What

Adds `custom1` to `ArmingMode`, so `setMode("custom1")` is accepted instead of refused with `mode: 3 is not a valid value (must be one of 0/1/63)`.

## Evidence

`ARMING_MODE_WIRE` already maps `custom1 -> 3` and `armingCommand()` is mode-agnostic — it places the wire value into `{mode_type, user_name}` on cmd 1224 — so `ArmingMode` was the only gate.

I've read the rule that unverified write wires throw rather than guess, and the reasoning: on a fire-and-forget wire a wrong frame is indistinguishable from success. So, precisely what I have and don't have:

- **Not** a byte-exact capture of the app sending mode 3. I don't have the app traffic.
- **Live-confirmed** on a T8030, 2026-09-12: `setMode("custom1")` accepted, station reported its own MODE_SWITCH push within the convergence window. Two writes, cycling `custom1 -> home -> custom1`.

The convergence check is what I'd argue makes this more than fire-and-forget: a mode the station ignored surfaces as a convergence failure, not as success. That is weaker than a capture and the wire table says so — `custom1` goes under a new "confirmed live (not byte-captured)" bullet rather than joining the captured three.

If that doesn't meet the bar, closing this is a completely reasonable call. I have the hardware and am happy to collect a proper capture if you tell me what you need from it.

The other five (`schedule`, `custom2`, `custom3`, `off`, `geo`) stay refused — untested, and I'm not extrapolating from one.

## Changes

- `ArmingMode` gains `custom1`, in wire order so `SETTABLE_MODES` reads `0/1/3/63`
- Wire-table JSDoc: new live-confirmed bullet for `custom1`; counts corrected (six -> five uncaptured, three -> four settable)
- `arming.spec.ts`: `custom1` out of the refusal table, plus a positive case asserting `mode_type: 3` reaches the wire
- `manifest.spec.ts`: published action domain `[0, 1, 63]` -> `[0, 1, 3, 63]`

## Verification

`npm run verify` on this branch — clean. Format, typecheck, all five guards (decorrelation, lines, docrefs, consumer-agnostic, capability-ownership), build, ESM check, examples typecheck, and 3487 tests across 182 files.

## AGENTS.md corrections

| Rule | What I'd done | Fixed |
|---|---|---|
| No real device data anywhere | Pasted a real station serial in the PR body log excerpt | Serial gone; no capture log quoted |
| Consumer-agnostic | PR body named a specific host stack and its integration | Removed; no consumer named |
| Shipped `src/` cites peers only | JSDoc said "See the PR discussion for the log" | Rewritten to state the protocol fact inline |

The last one would have dangled in `dist/` for consumers and inverted the direction `docs/` is generated in. Apologies for the noise — worth reading the file before the first submission, not after.


lenoxys (changes requested):
Hello, 

Can I ask you to run against AGENTS.md ?

Thanks!

holland-built:
Done — read `AGENTS.md` and found three things I'd broken, all mine:

1. **Real device data.** The original body pasted a station serial in a log excerpt. Removed, and I've checked the diff and commit message are clean too.
2. **Consumer-agnostic.** The body named a specific host stack and its integration. Removed.
3. **`src/` cites peers only.** My JSDoc line said "See the PR discussion for the log" — that dangles in `dist/` and inverts the direction `docs/` is generated in. Rewritten to state the protocol fact inline.

Force-pushed. `npm run verify` is clean end to end — format, typecheck, all five guards, build, ESM, examples, 3487 tests.

On the substance: the one thing I can't satisfy is a byte-exact capture, and the wire table records that honestly rather than claiming parity with the captured three. If the live convergence confirmation isn't sufficient evidence for a write, I'd rather you close this than have a weaker standard land in the table — say the word. I have the hardware if a capture is what you need.

lenoxys (changes requested):
AGENTS.md pass looks right — serial, consumer name and the doc-ref are all clear now. Thanks for going back through it.

On the evidence: keep it. The hard rule says a write ships once it is *confirmed on a real device*, not once it is byte-captured — byte-exact is the convention the table grew, not the bar. A station reporting back the mode it landed in is confirmation, and the wire shape here was already captured; the only new claim is that an app-defined integer is accepted on it. A new tier in the table is the honest way to record that. Not closing this.

Three things before it merges.

**1. `ArmingMode` is the domain of two writes.** It also types `setAlarmDelayConfig`'s mode, which puts the same integer into `mode_id` on cmd 1255. No runtime gate there — the union *is* the gate — so this also opens 1255 to mode 3. That wire has no capture for 3, no known GET, and no readback, which is exactly "a described control whose state cannot be read back is worse than none". Your evidence is 1224-only and should stay that way: give the alarm-delay parameter its own union over the byte-captured three and leave `ArmingMode` as `setMode`'s domain. Its JSDoc already promises this ("a delay can only be configured for a mode whose mode_id integer is captured"), so today the doc and the type disagree.

**2. The member's `description` still refuses custom1.** It says it sets only three and names custom1 among the refused, sitting in the same declaration as `values: [0, 1, 3, 63]`. That string is published — it is what a caller reads to offer the control — and one-feature-one-entry means it moves with the numbers. Same staleness in the barrel's arming one-liner, the `ARMING` module doc, and `SET_ARMING` ("those three are the whole of ArmingMode", "the remaining six"). `ARMING_MODE_WIRE` tells you to flip the `SET_ARMING` note when a mode is added — that's the step that got missed. Also re-read the `ArmingMode` opener: as it stands it now claims byte-exact capture for four values, which overstates your own case.

**3. Trim the new prose to the fact.** "Weaker evidence than a byte-exact capture, and recorded as such: ..." is argument, and the ✅/⚠️ split already carries it — a JSDoc states the declaration, not its case. Hardware, date and the convergence observation are the ground truth; the reasoning is already in your commit message, which is where it belongs. And drop the `// ✅ settable` on the wire line: membership in `ArmingMode` is what makes a mode settable, the ⚠️ lines exist only to mark the exceptions, and away/home/disarmed carry nothing.

Nit while you're in there: the positive `it.each` above already asserts the full frame for the other three. `[ArmingMode.custom1, 3]` in that table covers user_name/channel/mValue3 too, and the standalone case can go — its read stub is unused.

lenoxys (commented):
Same three items as above, anchored to the lines they land on.

src/model/capabilities/arming.ts
```diff
@@ -24,6 +24,8 @@ export const ArmingMode = {
   away: "away",
   /** Armed for occupancy — reduced/perimeter protection while home (wire value 1). */
   home: "home",
+  /** Custom 1 — a user-defined posture configured in the app (wire value 3). */
+  custom1: "custom1",
```
lenoxys, line 32: This is the blocking one. `ArmingMode` is not only `setMode`'s domain — it also types `setAlarmDelayConfig`'s `mode`, and `alarmDelayCommand` puts that same integer into `mode_id` on cmd **1255**. There is no runtime check on that path, the union *is* the gate, so this line silently opens 1255 to mode 3 as well.

1255 has no capture for 3, no known GET, and no readback — the case the rule calls worse than none. Split the domains: keep `ArmingMode` as `setMode`'s, and give the alarm-delay parameter its own union over the byte-captured three. `setAlarmDelayConfig`'s JSDoc already promises exactly that ("a delay can only be configured for a mode whose `mode_id` integer is captured"), so as it stands the doc and the type disagree.

src/model/capabilities/arming.ts
```diff
@@ -13,7 +13,7 @@ const STATION_CHANNEL = 255;
  * real station. `ArmingMode` is both the const value-object (`ArmingMode.home`) and the union type of its
  * values, so callers pass the named constant: `setMode(ArmingMode.home)`.
  *
- * Deliberately NARROWER than the set a device may report. The other six modes are ones the app itself
+ * Deliberately NARROWER than the set a device may report. The remaining five modes are ones the app itself
```
lenoxys, line 20: Counts moved, prose didn't. The opener two lines up still reads "the three whose write was captured byte-exact" — after this change it claims byte-exact provenance for four values, which is stronger than what you're arguing.

Same staleness in three more places this diff doesn't touch: `mode`'s `description` (still lists `custom1` among the refused, in the same declaration as `values: [0, 1, 3, 63]` — and that string is published, it's what a caller reads to offer the control), the `ARMING` module doc ("3 of its 8"), the arming one-liner in `capabilities/index.ts`, and `SET_ARMING` ("those three are the whole of `ArmingMode`", "the remaining six"). `ARMING_MODE_WIRE` tells you to flip the `SET_ARMING` note as each mode is added — that's the step that got missed.

src/model/capabilities/arming.ts
```diff
@@ -92,7 +94,11 @@ export const ARMING_CMD = {
  *
  * ✅ WRITE WIRE-CAPTURED (byte-exact, a T8030 2026-07-23; all three re-confirmed live 2026-08-05, each
  * reporting its own MODE_SWITCH push within ~5s): `away` 0, `home` 1, `disarmed` 63.
- * ⚠️ WRITE NEVER CAPTURED: `schedule` 2, `custom1` 3, `custom2` 4, `custom3` 5, `off` 6, `geo` 47. The app
+ * ✅ WRITE CONFIRMED LIVE (not byte-captured): `custom1` 3 — a T8030 2026-09-12, accepted and reporting
+ * its own MODE_SWITCH push within the convergence window on each of two writes, cycling
+ * custom1→home→custom1. Weaker evidence than a byte-exact capture, and recorded as such: the station
+ * confirming the mode it landed in is what distinguishes this from a fire-and-forget write.
```
lenoxys: These two lines are argument, not ground truth — the ✅/⚠️ split already says which tier the evidence is in, so a reader doesn't need it re-justified. Keep hardware, date and the convergence observation; drop from "Weaker evidence…" on. The reasoning is already in your commit message, which is where it belongs.

src/model/capabilities/arming.ts
```diff
@@ -101,7 +107,7 @@ const ARMING_MODE_WIRE: Record<ArmingMode, number> & Record<string, number> = {
   away: 0,
   home: 1,
   schedule: 2, // ⚠️ reportable, NOT settable — see the doc comment above
-  custom1: 3, // ⚠️ reportable, NOT settable — see the doc comment above
+  custom1: 3, // ✅ settable — confirmed live 2026-09-12, see the doc comment above
```
lenoxys: Drop the comment, keep `custom1: 3,`. Membership in `ArmingMode` is what makes a mode settable — the ⚠️ lines exist only to flag the exceptions, and `away`/`home`/`disarmed` carry nothing. As written it restates the type, and the date is already in the doc block above.

src/model/capabilities/__tests__/arming.spec.ts
```diff
@@ -82,11 +81,21 @@ describe("arming capability module", () => {
...
     });
 
+    it("sends custom1 (mode_type 3) — confirmed live on a T8030 2026-09-12", async () => {
+      const readCtx: CommandContext = { ...ctx, paramIds: new Set([ARMING_CMD.SET_ARMING]) };
+      const { acts, sent } = bind<ArmingActions>("arming", readCtx, {
+        read: (p) => (p === "armingMode" ? { value: 0 } : undefined),
+      });
+      await acts.setMode("custom1");
+      expect(sent).toHaveLength(1);
+      expect(sent[0]).toMatchObject({ payload: { mode_type: 3 } });
```
lenoxys: Nit: the positive `it.each` at the top of this describe already asserts the complete frame for the other three. `[ArmingMode.custom1, 3]` in that table covers `user_name`, `channel` and `mValue3` too, where this checks `mode_type` alone via `toMatchObject`. The whole block can go — its `readCtx` and `read` stub are unused, nothing here reads `acts.mode`.

holland-built:
All four addressed in a26146a.

**1. `ArmingMode` is the domain of two writes.** You're right, and this was the real defect — not the evidence tier. `setAlarmDelayConfig` takes the same union and `alarmDelayCommand` puts that integer into `mode_id` on 1255, with no runtime gate on the path, so adding `custom1` opened a second wire as a side effect. My live confirmation was gathered on 1224 and says nothing about 1255, which has no capture for mode 3 and no GET to read one back — so shipping it there is exactly the described-control-with-no-readback case.

Split: new `AlarmDelayMode` over the byte-captured three types the alarm-delay parameter, `ArmingMode` stays `setMode`'s with four. `setAlarmDelayConfig`'s JSDoc already promised this, so doc and type agree now. I also extended the `ARMING_MODE_WIRE` maintenance note to say which union a new capture widens, and that widening one is not licence to widen the other — that note is what failed to stop this.

**2. Counts moved, prose didn't.** Fixed all of it, including the two you flagged outside the diff. The `mode` member's published `description` was the worst of them — it named `custom1` among the refused while sitting next to `values: [0, 1, 3, 63]`, and that string is what a caller reads. Also corrected: the `ArmingMode` opener (it did claim byte-exact provenance for four), `SET_ARMING`, `ARMING_MODE_WIRE`, `SETTABLE_MODES`, the `mode` member doc, the `ARMING` module doc (which also said 8 where the read reports 9), and the barrel's arming one-liner.

**3. Trim to the fact.** Dropped from "Weaker evidence…" on; hardware, date and the convergence observation stay. `// ✅ settable` is gone from the wire line. Agreed on both — the tier markers carry it and the argument belongs in the commit.

**Nit.** Taken. Folded `[ArmingMode.custom1, 3]` into the positive `it.each` and deleted the standalone block with its unused read stub. Added one note on the describe so the table doesn't imply all four rows are byte-captured.

While rewriting I'd drafted the new JSDoc explaining the bug it fixed, which `AGENTS.md` rules out — rewrote all three blocks as declarations.

`verify` green on the beta base: prettier, typecheck, all five guards, build, 271 ESM exports, examples, 3487 tests.

Unrelated, but worth flagging since it affects any agent running the gate: the five `guard:*` scripts use `grep --include`, which BusyBox grep doesn't support. On an Alpine/BusyBox box all five print grep's usage text and exit 0 — a false pass on every one. CI's GNU grep is fine, so this is local-only. Happy to send a separate PR switching them to `find -name … -exec` or a `grep -r` form that works on both, if you want it.

lenoxys (changes requested):
Verify green on my side too, and all three items are addressed properly — the split is complete (`ArmingMode` now appears only on the 1224 path: the wire table's key type, `armingModeOf`, `armingCommand`, `accepts<ArmingMode>()`), the counts are clean everywhere I grepped for them, and it isn't a breaking change since `AlarmDelayMode` is exactly the signature `setAlarmDelayConfig` had before. Thanks for taking the second write seriously rather than arguing the point.

One thing left, and it's the reason I'm not merging yet.

**The narrowing has nothing holding it in place.** The whole point of `a26146a` is type-level, and vitest does not typecheck — merge the two domains back together and all 3487 tests still pass. That is a fix that can be undone by accident, and the next person to widen `ArmingMode` is the one who will do it. `tsc --noEmit` covers the specs (`include: ["src/**/*"]` — only the *build* config drops them), so one line in `arming.spec.ts` is a real CI gate:

```ts
// @ts-expect-error custom1 is confirmed on cmd 1224 only — the alarm-delay domain excludes it
void ((): AlarmDelayMode => ArmingMode.custom1)();
```

It passes as written and fails with an unused-`@ts-expect-error` the moment the two domains become one again, which is exactly the alarm worth having. Put it next to the alarm-delay describe block, where a reader hits it in context.

Nits, take or leave:

- The alarm-delay specs still call `setAlarmDelayConfig(ArmingMode.away, ...)`. It typechecks — `ArmingMode.away` is the literal `"away"` — but `AlarmDelayMode.away` makes the split visible at the place that exercises it.
- `describe("setMode / buildCommand (wire captured live on a T8030, 2026-07-23)")` now carries a row from a different date. The new `//` lines above it say so; the sibling tables in this file use a `/** */` block, which is also what the comment rule asks for.
- `AlarmDelayMode`'s three entries repeat `ArmingMode`'s docs verbatim, and deriving it by omission would be shorter — but it would then widen silently on the next `ArmingMode` addition, which is the bug you just fixed. Explicit is correct here; I'm noting it only so nobody "tidies" it later.

Not yours to fix, but worth an issue while it's fresh: `alarmDelayCommand` has no runtime gate at all — an untyped caller can hand it any string, and the wire table resolves `custom1` to 3 or a non-key to `undefined`, shipping `mode_id: undefined`. Predates this PR and this change doesn't widen it.

Add the pin and I'll merge.

lenoxys:
Docs CI/CD must be fixed as well

max246:
Can you set resolved comments if they are been addressed?

holland-built:
Pin added in ec64249, and the docs build is fixed.

**The pin.** You were right that nothing held the narrowing in place — I checked before adding it, and merging the two unions back together leaves all 3487 tests passing. Added as given, next to the alarm-delay describe, and verified both directions: clean as written, and `TS2578: Unused '@ts-expect-error' directive` the moment `custom1` goes into `AlarmDelayMode`.

**Docs CI.** Mine, from a26146a, and not a `guard:docs` hit — `{@link ARMING_CMD.ALARM_DELAY_CONFIG}` in the `ArmingMode` and `AlarmDelayMode` comments resolved to a symbol the reference doesn't include, one warning per comment, four total, and the build treats warnings as errors. Both now name cmd 1255 in plain text. Ran the docs toolchain locally to confirm rather than guessing again: typedoc 0 warnings, `guard:docs` OK, VitePress build complete.

Both nits taken — the alarm-delay specs call `AlarmDelayMode.away` where they exercise the split, and that note is a `/** */` block like its siblings. `AlarmDelayMode` stays explicitly listed; your reasoning for not deriving it is in the commit message so the next person finds it there.

Threads resolved, @max246.

Filed the runtime-gate finding as #163 rather than folding it in here — confirmed it first: `method()` wraps the builder with no argument validation, so the type is the only gate, and `ARMING_MODE_WIRE` being `Record<string, number>` means a non-key indexes to `undefined` silently. Worth noting while I was in the labels: `AGENTS.md` names `needs-triage` as canonical but it doesn't exist in the repo's label set, so #163 is `bug` only.

One last thing from earlier, still standing if you want it: the five `guard:*` scripts use `grep --include`, which BusyBox doesn't support — on a musl/BusyBox box all five print grep's usage and exit 0, a false pass on every one. CI's GNU grep is unaffected, so it only bites local work. Happy to send it as its own PR.

lenoxys (approved):
The pin is in and it works. I checked it the only way worth checking a type-level guard — by breaking it: adding `custom1` back to `AlarmDelayMode` makes `tsc --noEmit` fail with `TS2578: Unused '@ts-expect-error' directive` on `arming.spec.ts:160`, and restoring it goes green. So the narrowing now has something holding it in place, and the person who widens `ArmingMode` next will be told at build time rather than shipping a `mode_id` nobody has captured.

The JSDoc above it is the right kind: it says what the pin is for and why nothing else could enforce it, so the line does not read as a curiosity to be tidied away. Both nits taken too — the alarm-delay specs call `AlarmDelayMode.away`, which makes the split visible at the place that exercises it, and the dated note is a `/** */` block like its siblings.

The TypeDoc fix is real and was worth catching here rather than after merge: `{@link ARMING_CMD.ALARM_DELAY_CONFIG}` resolved to a symbol the reference does not include, one warning per comment, and warnings are errors in that build. Plain text naming cmd 1255 loses nothing — the surrounding prose already carries the association.

On the original question, for the record: a station accepting mode 3 and reporting it back within ~5s, twice, with the convergence observation, is confirmation on a real device, which is the bar `AGENTS.md` actually sets. The `⚠️ WRITE CONFIRMED LIVE (not byte-captured)` tier records the difference honestly instead of flattening it into the captured rows, and `custom1` is now correctly settable on 1224 and correctly absent from 1255.

Verified: `npm run verify` green on my side (182 files, 3487 tests), and the full docs build too — typedoc, `guard:docs`, VitePress, exit 0.

One procedural note: CI had never actually run on this PR. Fork pull requests need a maintainer to release the workflow runs, and both were sitting on `action_required` — so the green you and I were both reading was local-only. I have approved them; CI and docs are now genuinely green on `ec64249`, which is what should be true before this merges.

Approving. `alarmDelayCommand`'s missing runtime gate stays out of scope as agreed — it predates this and this change does not widen it; I will open it as its own issue.

Issue #159, opened 2026-09-12
Long-lived secure-MQTT connects under an account-scoped client id, so two installs on one account evict each other


martijnpoppen:
## What happens

`SecureMqtt.id` falls back to the certificate CN, `{user_id}-{app_name}`:

```ts
// src/transport/mqtt/secure-mqtt.ts:95
get id(): string {
  return this.o.clientId ?? this.o.credentials.thing_name ?? "";
}
```

and `EufyMega.startMqtt()` never passes a `clientId`:

```ts
// src/client/eufy-mega.ts:1720
const transport = new SecureMqtt({ credentials: creds, logger: this.opts.logger });
```

So every long-lived connection identifies itself by the **account and line**, with nothing per install in it. Two SDK processes signed into the same account — a Homey app and a Home Assistant integration, two Homeys, a restarted process racing its own previous socket — connect under an identical client id, and AWS IoT treats a duplicate client id as a takeover and evicts the incumbent. Each eviction is a `close`, mqtt.js reconnects after `reconnectPeriod` (5 s), and that connect evicts the other one: both sides flap indefinitely, neither ever holding a usable realtime channel.

This is not a new finding for the repo — `topics.ts` already names the exact mechanism as the reason the Clean line was not given a scope of its own:

> a scope of its own would open a second connection under the SAME `thing_name` — AWS IoT treats a duplicate client id as a takeover and evicts the incumbent, so both lines would flap.
> — `src/transport/mqtt/topics.ts:29`

The same hazard applies between *processes*, and there the SDK has no control over what else is signed in.

The one-shot path already solves this. `MqttCommandRouter` builds a per-install id and passes it explicitly:

```ts
// src/transport/mqtt/command-router.ts:372
if (!this.mqttUuid) this.mqttUuid = generateMqttUuid();
const clientIdFor = () => buildAppShapedClientId({ appName: "eufy_security", uid: creds.user_id ?? "", mqttUuid });
```

`SecureMqttOptions.clientId` and `buildAppShapedClientId()` / `generateMqttUuid()` all exist. The long-lived transport is the one path that doesn't use them.

**Expected:** two independent installs on one account each hold their own realtime connection.
**Actual:** they take the connection from each other, on a ~11 s cycle, for as long as both run.

### Suggested fix

Give `startMqtt()` the same treatment as the command router — a client id carrying a stable per-install component:

```ts
const transport = new SecureMqtt({
  credentials: creds,
[...]

PR #160, opened 2026-09-12
fix(mqtt): identify a realtime connection by install, not by certificate

martijnpoppen:
## What changed

`startMqtt()` passed no `clientId`, so `SecureMqtt.id` fell back to the certificate's `thing_name`. The long-lived transports now connect under the app's own id shape, `android-{app_name}-{uid}-{uuid}-{ts}`, with the uuid derived from this client's `openudid`.

- `app-client-id.ts` — `mqttUuidFrom(installId)`, hashing a stable install id into the 16-hex shape `buildAppShapedClientId` takes. It replaces `generateMqttUuid`, so one operation has one owner; the one-shot path in `command-router.ts` takes it too and its ids stop changing every process.
- `mega-client.ts` — `openudid` exposed as a `readonly` field. No new option: it is already "the stable per-install device id", already persisted, and already what `CONTENDED_SESSION_HINT` tells a caller to make unique per client.
- `secure-mqtt.ts` — `isNotAuthorized(err)`, the CONNACK refusal a client cannot retry its way out of; and an info line when a connection is established, without which a log cannot distinguish a channel that never connects from one that connects and is evicted every few seconds.
- `eufy-mega.ts` — the id is built in `startMqtt()`; `connectMqtt(creds, clientId?)` connects and then wires the fan-out, so a discarded attempt reports nothing to a consumer.

## Why

`{user_id}-{app_name}` names the account and the line and carries nothing of the client, and the broker resolves a duplicate client id by evicting the incumbent. Two clients on one account therefore take the channel from each other indefinitely, neither holding one — the same takeover `topics.ts` already cites as the reason the clean line was not given a scope of its own, except that between processes the SDK controls neither side. Refs #159.

Deriving the uuid from `openudid` rather than generating one keeps it stable across restarts, so a client resumes as itself and takes over its own stale session instead of doubling up beside it. It is hashed because the client id travels the wire in clear and should not carry the identity it came from.

**How far this separates two clients: exactly as far as their `openudid` does.** A caller that supplies none is given one derived from the ACCOUNT (`md5("eufy-mega:" + email)`), so two such clients still evict each other — the same condition under which their logins already displace each other, which is what `CONTENDED_SESSION_HINT` exists to say. Per-caller `openudid` is the contract, stated on both declarations and asserted in a spec rather than left to inference. Making that default distinct means a persisted random in the HTTP identity path, with its own 2FA consequences; that is a separate change.

## Evidence

The app-shaped id is confirmed granted on the `eufy_security` credential: `ensureSecurityMqttFor` has connected with it (account's own cert, `buildAppShapedClientId`) since the broker-instance probe landed, and the grant/deny pattern there is per instance, never per client id. The eviction behaviour itself is the vendor broker's AWS IoT semantics and is already stated in `topics.ts`.

What is NOT evidenced: that the `default` and `eufy_life` credentials' policies also accept a client id that is not the certificate CN. Same broker, same issuing endpoint, so it is a reasonable inference — but an inference, and if one of those policies pins the id to the CN the connect is refused and that line loses realtime.

A refused connect therefore falls back to the certificate's name, since a shared channel beats none — **only** a refusal. A connect that fails any other way rejects the bring-up, whose memo `ensureMqttStarted` clears, so the next one asks under this client's own id again: a dropped socket must not be what moves a process onto the shared name, which a transport that connects then keeps until `disconnect()`.

A field report consistent with the takeover (two scopes, ~11 s reconnect cycle, a window where both were connected) is in #159. It is corroboration, not proof: the same trace also fits a shared network or broker-instance cause.

## Verified how

- [x] `npm run verify` is green locally — format, typecheck, all four guards, build, ESM check, examples, 3497 tests.
- [ ] Ran against a real device — **no.** Nothing here can be confirmed without the live broker, which is why the fallback exists. (This box was ticked by a stray click on the PR page; unticking it, since it is not true.)
- [x] Added or updated specs (offline, synthetic fixtures) — `src/client/__tests__/mqtt-client-id.spec.ts` drives `ensureMqttStarted`, since the memo is what the claims are about: the id identifies the client and not the certificate, the credential's own `app_name` wins over the scope's, a restart keeps its id, two clients differ only if their `openudid` does, a refusal falls back and that transport is kept, a reset does not fall back and the next bring-up asks under the client's own id, and a discarded attempt announces no connect. Plus one in `ensure-security-mqtt.spec.ts` pinning the one-shot path's now-stable id.
- [x] Docs updated, or n/a — JSDoc only; no prose file describes this.

Also re-synced into a consuming app and re-ran its own gate (typecheck, tests, manifest validation) to confirm the added exports, the removed `generateMqttUuid` and the `readonly` field break no consumer build.

## Unverified or deferred

- The client id shape on the `default` and `eufy_life` credentials, as above. The fallback is what makes shipping it safe; a live confirmation on either line would let it go.
- **What narrowing the fallback costs:** a policy refusal that arrives as a bare TCP reset rather than a CONNACK is indistinguishable from a transport failure, so it no longer falls back and that line gets no realtime until someone looks. That is the side worth failing on — loud and fixable with evidence in hand, where the alternative pinned a process to the colliding name silently — but it is a real narrowing, and `isNotAuthorized`'s JSDoc says so.
- Two clients that configure no `openudid` still collide, as above. Narrowed in the claim, not closed in the code.

## Base branch

Targets `beta-0.2.0`, which currently sits at `release: 0.1.0` and is five commits behind `main` — so this PR's diff carries those five alongside the fix. Retarget it (or bring the beta branch up to `main` first) if that is not the intended staging branch for this release.


src/client/eufy-mega.ts
```diff
@@ -1707,17 +1708,51 @@ export class EufyMega extends EventEmitter {
...
-   * The inbound decode is gated by the reporting device's own capabilities, so one line's decoder never
-   * runs against another's traffic, and the DP frame is unwrapped here — the layer that may import the
-   * transport — so a capability reads tags without owning any framing.
+   * Identified by a client id carrying this install's own `openudid`, not by the certificate's name:
+   * that name is `{user_id}-{app_name}`, which every client on the account shares per line, and a
+   * duplicate client id is a takeover the broker resolves by evicting the incumbent. The id shape is
+   * the app's own (`android-{app_name}-{uid}-{uuid}-{ts}`, see {@link buildAppShapedClientId}), which
+   * the broker grants on the `eufy_security` credential; a credential whose policy refuses it falls
+   * back to the certificate's name, since a shared connection beats none. The fallback is not
+   * remembered — the next bring-up asks under this install's own id again.
```
lenoxys: **Blocking — this sentence is not true of a successful fallback.**

`ensureMqttStarted` clears its memo **only when the attempt rejects** (`:493-495`). A fallback that connects is stored in `mqttReady` and `transports`, and from then on mqtt.js reconnects *that* transport, under the certificate's name, for the lifetime of the process. Nothing re-asks under the install's id until an explicit `disconnect()` clears the map.

So the guarantee holds only in the case where the fallback *also* fails — which is the one case where it does not matter. In the case that matters it inverts: one bad first handshake and the process is pinned to the colliding name, which is the exact failure #159 opened for.

This is a `dist/`-shipped JSDoc, so a consumer reads it as fact ("A JSDoc … ground-truth", `AGENTS.md`). Either make it true — clear the scope's memo when the fallback is taken, so the next bring-up retries the install id — or state the real behaviour: the fallback holds until the next `disconnect()`.

The spec at `mqtt-client-id.spec.ts:111` does not catch this; see the comment there.

src/client/eufy-mega.ts
```diff
@@ -1707,17 +1708,51 @@ export class EufyMega extends EventEmitter {
...
     const auth = this.mega.auth;
     if (!auth) throw new Error("login() first");
     if (!this.registry.list().length) await this.getDevices();
 
     const creds = await this.getUserMqttInfo(mqttAppName(scope));
-    const transport = new SecureMqtt({ credentials: creds, logger: this.opts.logger });
+    const perInstall = buildAppShapedClientId({
+      appName: creds.app_name ?? mqttClientAppName(scope),
+      uid: creds.user_id ?? auth.userId,
+      mqttUuid: mqttUuidFrom(this.mega.openudid),
```
lenoxys, line 1736: **Blocking — `openudid` is not per-install when the caller does not set it.**

`mega-client.ts:333-334`:

```ts
this.openudid = cfg.openudid ?? saved?.openudid ?? createHash("md5").update(`eufy-mega:${cfg.email}`).digest("hex").slice(0, 16);
```

The third branch is derived from the **account**, not the install. Two hosts on one account run in separate processes with separate stores, so `saved?.openudid` is empty in both; neither passes `cfg.openudid`; both derive the same value. Same openudid → same uuid → same client id → they still evict each other.

`mqttUuidFrom` is doing exactly what it says. The problem is upstream: the input is only per-install for a caller that made it so. Since #159's scenario is two hosts that *don't* know to, the fix does not reach its own motivating case by default.

The counter-argument is sound as far as it goes: `CONTENDED_SESSION_HINT` (`mega-client.ts:226`) already tells a caller to "give each client its own openudid", so the contract arguably predates this PR. If that is the intended reading, it should be stated outright — here and in the PR body — rather than left to a reader who will otherwise take "this install's own `openudid`" at face value. What would settle it in code: fall back to a persisted random value rather than a hash of the email, so an unconfigured caller is distinct by default. That is a change to the HTTP identity path with its own 2FA implications, so it may well belong in a separate PR — but then this PR's claim needs narrowing, not the code.

src/client/eufy-mega.ts
```diff
@@ -1707,17 +1708,51 @@ export class EufyMega extends EventEmitter {
...
+    try {
+      return await this.connectMqtt(creds, perInstall);
+    } catch (e) {
+      this.opts.logger?.warn(
+        "[smqtt] the broker refused a per-install client id; connecting under the certificate's own name, " +
+          "which another client signed in to this account will take over",
+        e,
+      );
+      return await this.connectMqtt(creds);
+    }
```
lenoxys, line 1748: The `catch` is unconditional, so it treats "the policy refuses this client id" and "the socket died" as the same event. Only the first deserves a fallback; the second deserves a retry under the same id.

Paired with the memo (see the JSDoc comment above), the consequence is not just a wasted handshake as the PR body suggests — a single `ECONNRESET` on a first bring-up moves the process onto the colliding name and *keeps* it there. The cost of conflating the two is the whole fix, not one extra round trip.

The spec at `:122` shows the shapes are distinguishable in practice: an authorisation refusal arrives as `"Connection refused: Not authorized"`, a transport failure as `ECONNRESET`. Narrowing the `catch` to the refusal wording — or, failing that, clearing the memo so a transient error cannot become permanent — closes it.

Also: the warn fires on every refusal, including the transient ones, and says the broker refused the id when it may have said nothing at all.

src/client/eufy-mega.ts
```diff
@@ -1707,17 +1708,51 @@ export class EufyMega extends EventEmitter {
...
+  }
+
+  /**
+   * Build one secure-MQTT transport, wire its decode and fan-out, and connect it.
+   *
+   * `clientId` overrides the certificate's own name (`{user_id}-{app_name}`), which identifies the
+   * ACCOUNT and the line rather than this install: the broker treats a second connection under the
+   * same client id as a takeover and evicts the incumbent, so two installs on one account hold the
+   * channel from each other indefinitely. Omitting it connects under that shared name.
```
lenoxys: Two `AGENTS.md` comment rules, both non-blocking:

1. **Duplication.** These four lines restate the takeover mechanism already given at `:1711-1717`, near-verbatim. Two copies drift; when the broker behaviour is next re-confirmed, one of them will be edited and the other will not. `{@link startMqtt}` and delete.
2. **"A JSDoc does not narrate its own history."** The parameter is defined by what identification *used to* be — "overrides the certificate's own name", "Omitting it connects under that shared name". `clientId` does not override anything; it *is* the id, and omitting it selects the certificate's. The takeover reasoning is genuinely valuable, but it is why the *commit* exists, and the commit body already carries it well.

The third paragraph (`:1750-1752`) is the one that belongs on this declaration — it describes what `connectMqtt` actually does.

src/client/eufy-mega.ts
```diff
@@ -1707,17 +1708,51 @@ export class EufyMega extends EventEmitter {
...
+   * remembered — the next bring-up asks under this install's own id again.
    */
   private async startMqtt(scope: MqttScope): Promise<SecureMqtt> {
     const auth = this.mega.auth;
     if (!auth) throw new Error("login() first");
     if (!this.registry.list().length) await this.getDevices();
 
     const creds = await this.getUserMqttInfo(mqttAppName(scope));
-    const transport = new SecureMqtt({ credentials: creds, logger: this.opts.logger });
+    const perInstall = buildAppShapedClientId({
```
lenoxys: Nit: `perInstall` names the axis, not the value — it holds a client id. `perInstallClientId` reads correctly at the two use sites, and distinguishes it from the uuid, which is the part that is actually per-install.

Related, and the reason the naming matters here: `mqttUuidFrom(this.mega.openudid)` reaches through the HTTP client to derive an MQTT identity. It works, and exposing the field beat inventing an option — but the reach is what makes the correctness of `openudid` this file's problem, and it is the second reader with no obvious link to the first.

src/transport/http/mega-client.ts
```diff
@@ -292,8 +292,12 @@ export class MegaHttpClient {
...
   private pending2fa = false;
   private tokenExpiresAt = 0;
-  /** Stable per-install device id — the auth token is bound to it. */
-  private openudid: string;
+  /**
+   * Stable per-install device id: `openudid` as configured, as restored from the session store, or as
+   * derived from the account when neither supplied one. The auth token is bound to it, and two clients
+   * sharing it displace each other's session.
+   */
+  readonly openudid: string;
```
lenoxys, line 304: This JSDoc is the most accurate description of the problem in the diff — "**or as derived from the account** when neither supplied one" — and it is the sentence that makes the client id built from it not per-install. See `eufy-mega.ts:1728`.

Worth being explicit that "two clients sharing it displace each other's session" now has a second meaning: before this change it cost the HTTP token, after it, the realtime channel too. Same input, two failure modes, one of which is the bug this PR is closing.

On the widening itself: `private` → `readonly` publishes a transport-internal identity to the package surface for one internal reader. Preferring it over a new config option was the right call, and the field is genuinely the stable install id — just note that once it is public it is API, and `AGENTS.md` "No backward-compat below 1.0" cuts the other way here: it is easier to widen later than to narrow. Not a blocker.

src/transport/mqtt/app-client-id.ts
```diff
@@ -35,3 +35,13 @@ export function buildAppShapedClientId(input: AppClientIdInput): string {
...
+
+/**
+ * The `mqttUuid` for an identity that is already stable and persisted, hashed to the 16-hex shape
+ * {@link buildAppShapedClientId} expects. Deterministic, so the same install produces the same value
+ * on every run and two installs produce different ones; one-way, so the id it is derived from is not
+ * recoverable from a client_id that travels the wire in clear.
+ */
+export function mqttUuidFrom(installId: string): string {
+  return createHash("sha256").update(installId).digest("hex").slice(0, 16);
+}
```
lenoxys, line 43: The function is right — deterministic, one-way, correct shape, and the hashing rationale is the good kind of JSDoc.

The issue is that it now sits directly beside `generateMqttUuid` (`:35`) with the **same contract**: produce the 16-hex `mqttUuid` segment. Two owners for one operation, and only one caller was converted — `command-router.ts:372` still takes the random per-process value. `AGENTS.md` "Reuse before abstraction" §1 (search before creating) and §3 (reduce after green, "resolve every match before running the full verification gate").

The reconciliation is small and strictly a win. `command-router.ts:217` types `deps.mega` as `MegaHttpClient`, and `openudid` is `readonly` as of this diff, so:

```ts
const mqttUuid = mqttUuidFrom(this.deps.mega.openudid);
```

replaces `:372-373`, and `generateMqttUuid`, the `mqttUuid` field (`:254`) and the lazy-init all go. The one-shot path gains the same restart stability this PR argues for on the long-lived one — its ids stop changing every process, which is the property the PR body says matters.

If the random value is deliberate there, that is fine, but it needs a line saying why the one-shot path wants instability where the persistent path wants the opposite.

src/transport/mqtt/topics.ts
```diff
@@ -70,6 +73,15 @@ export function mqttAppName(scope: MqttScope): string | undefined {
...
 }
 
+/**
+ * The `{app_name}` segment a scope's credential is issued under — the default line's is `eufy_mega`,
+ * which is the same segment the certificate CN carries (`{user_id}-{app_name}`). Unlike
+ * {@link mqttAppName} this is total: it answers the name itself, not the optional request header.
+ */
+export function mqttClientAppName(scope: MqttScope): string {
+  return scope === EUFY_LIFE ? EUFY_LIFE : EUFY_MEGA;
+}
```
lenoxys: This is a new exported function plus a new module constant (`:39`), duplicating `mqttAppName`'s `scope === EUFY_LIFE ? … : …` cascade (`:72`), for one call site and a two-way default. The single caller already has a `??` chain:

```ts
appName: creds.app_name ?? mqttAppName(scope) ?? "eufy_mega",
```

Same behaviour, one expression, and one less symbol on the public surface — `topics.ts` is re-exported through the mqtt barrel, so this is package API for an internal default.

The "total vs optional" distinction the JSDoc draws is real and well spotted; it just does not need a second function to express. If the pairing stays, the two should share one scope-keyed map so a third scope cannot be added to one cascade and missed in the other — which is the failure mode two parallel ternaries on the same type invite.

src/client/__tests__/mqtt-client-id.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+  });
+
+  it("is stable for one install and different for another", async () => {
+    await start(makeClient("a".repeat(16)));
+    await start(makeClient("a".repeat(16)));
+    await start(makeClient("b".repeat(16)));
+
+    const uuid = (i: number) => String(constructedWith[i].clientId).split("-")[3];
+    expect(uuid(0)).toBe(uuid(1));
+    expect(uuid(2)).not.toBe(uuid(0));
```
lenoxys: This proves `mqttUuidFrom` is deterministic, not that the id is stable across a restart — both clients are handed `openudid` explicitly, so the assertion never touches the path that decides what `openudid` *is*.

That matters because the derivation is where the defect lives (see `eufy-mega.ts:1728`): the `cfg → store → md5(email)` chain in `mega-client.ts:333-334` is unexercised, and this spec is structurally unable to catch two unconfigured clients colliding. It passes today and would pass with the bug fixed or unfixed.

The spec that would earn the claim: two clients constructed with **no** `openudid` and separate stores, asserting the uuids differ. That currently fails, which is the point. A companion — one client, a store that already holds an openudid, asserting the uuid survives reconstruction — is the actual restart case the PR body claims (`"a restart resumes as itself"`).

src/client/__tests__/mqtt-client-id.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  it("does not remember the fallback — the next bring-up asks under the install's id again", async () => {
+    const eufy = makeClient("a".repeat(16));
+    connect.mockRejectedValueOnce(new Error("Connection refused: Not authorized")).mockResolvedValue(undefined);
+    await start(eufy);
+    constructedWith.length = 0;
+
+    await start(eufy);
+
+    expect(constructedWith[0].clientId).toMatch(/^android-/);
```
lenoxys: This calls `startMqtt` directly, which bypasses `ensureMqttStarted` — and the memo in `ensureMqttStarted` is the thing that does the remembering. So the spec passes for a reason unrelated to its own title: `startMqtt` is stateless across calls, and always was.

The behaviour the name describes is the opposite of what the code does. Through the real path, a fallback that *succeeds* is memoised in `mqttReady`/`transports` and is not re-attempted under the install id at all. See the JSDoc comment at `eufy-mega.ts:1716`.

Driving this through `ensureMqttStarted` instead of `startMqtt` — first bring-up refused then falling back, second bring-up asserting the install id returns — is what pins the claim. It should fail as written today.

src/client/__tests__/mqtt-client-id.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    expect(constructedWith[0].clientId).toContain("-eufy_mega-");
+    expect(constructedWith[1].clientId).toContain("-eufy_life-");
+  });
+
+  it("follows the credential's own app name when it reports one", async () => {
+    await start(makeClient("a".repeat(16), { ...CREDS, app_name: "eufy_home" }));
+
+    expect(constructedWith[0].clientId).toContain("-eufy_home-");
+  });
```
lenoxys: These two are one assertion with two inputs: the `app_name`-absent branch and the `app_name`-present branch of the same `??`. One spec with both cases covers it and reads better as a statement about precedence — the credential's own name wins, the scope default fills in.

Worth noting `CREDS` carries no `app_name`, so `:85` is *already* exercising the absent branch on its `"default"` leg, and `:88` then sets it to `undefined` explicitly for `eufy_life`. The setup implies a distinction between the two that does not exist.

src/client/__tests__/mqtt-client-id.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+function makeClient(openudid: string, creds: Record<string, unknown> = CREDS) {
+  const eufy = new EufyMega({ email: "t@example.com", password: "x", openudid });
+  Object.defineProperty((eufy as any).mega, "auth", {
+    configurable: true,
+    get: () => ({ userId: UID, authToken: "t" }),
+  });
+  const dev = { sn: "T8000P0000000000", category: "eufy_mega" } as unknown as EufyDevice;
+  vi.spyOn((eufy as any).registry, "list").mockReturnValue([dev]);
+  vi.spyOn(eufy, "getMqttDevices").mockReturnValue([]);
+  vi.spyOn(eufy, "getUserMqttInfo").mockResolvedValue(creds as never);
```
lenoxys: Non-blocking, and specs are exempt from the docref rule — but three private reach-ins to stand one client up (`(eufy as any).mega`, `defineProperty` over the `auth` getter, `(eufy as any).registry`) means any rename inside the facade breaks these specs without changing behaviour. Given they are the only defence for the client id shape, that is the wrong direction of fragility.

Not asking for a redesign here. Flagging it because `makeClient` is the piece most likely to need widening when finding 1 is addressed — testing an *unconfigured* client means not passing `openudid`, and a store seam would be needed for the restart case anyway. If it grows a store parameter, that is the moment to reconsider the reach-ins rather than add a fourth.

lenoxys (changes requested):
## Two-axis review + over-engineering pass

Reviewed at `8c628c0d` against fixed point `d37f00a8` (the parent of this PR's only substantive commit), with **#159 as the spec**. The other five commits in the diff are already on `main` and are carried only because this targets `beta-0.2.0`; the PR body already discloses that, so nothing below concerns them.

Three passes: **Spec** (does it do what #159 asked), **Standards** (`AGENTS.md`), and an over-engineering pass. The work is well-argued and the fallback instinct is right. Two findings block, and both are the same species: **shipped prose asserting behaviour the code does not have.** That is the one kind of defect this repo's own standard singles out ("A JSDoc … ground-truth"), because `src/` ships in `dist/` and a consumer reads it as fact.

### Blocking

**1. The headline requirement is unmet in the default case.** #159: *"**Expected:** two independent installs on one account each hold their own realtime connection."*

`mqttUuidFrom(this.mega.openudid)` is deterministic — and so is `openudid` when no caller supplies one. `mega-client.ts:333-334` derives it as `md5("eufy-mega:" + cfg.email).slice(0, 16)`: **account** identity, not install identity. Two hosts on one account, separate processes so separate stores, neither passing `openudid`, therefore land on the same openudid → same uuid → the same client id, and still evict each other. Only the shared-CN half of the collision is closed.

The mitigation is real but partial, and worth stating in the PR body rather than leaving to inference: `CONTENDED_SESSION_HINT` (`mega-client.ts:226`) already instructs a caller to "give each client its own openudid", and `MegaClientConfig.openudid` predates this PR — so "no new option needed" holds. But the spec's own scenario is a host that does not know to set it, and that host still flaps. `openudid` now also silently governs MQTT identity, which raises the cost of getting it wrong.

Restart-stability itself is sound — resolved in the constructor from cfg → store → derivation, before login. No concern there.

**2. "The fallback is not remembered" is false for a successful fallback.** `ensureMqttStarted` clears its memo **only on rejection** (`eufy-mega.ts:493-495`). A fallback that *succeeds* is memoised in `mqttReady`/`transports`, and mqtt.js then reconnects it under the CN for the lifetime of the process — cleared only by an explicit `disconnect()`. Since the `catch` does not distinguish `"Not authorized"` from `ECONNRESET`, one transient first-attempt error pins the process to the colliding name: precisely the failure this PR exists to remove. Inline on the JSDoc and on the `catch`.

### Not inline (outside the diff)

**3. `command-router.ts:372` was left behind.** `mqttUuidFrom` lands beside `generateMqttUuid` with the same contract — produce the 16-hex `mqttUuid` segment — but only the long-lived path was converted. The one-shot path keeps a random per-process value, so two owners for one operation now exist, unreconciled inside the same change. `AGENTS.md` "Reuse before abstraction" §1/§3. Worth noting that `deps.mega` is a `MegaHttpClient` (`command-router.ts:217`) and `openudid` is `readonly` as of this diff, so converting it is a three-line change that also *deletes* `generateMqttUuid`, the `mqttUuid` field and its lazy-init — and hands the one-shot path restart stability for free.

**4. #159's secondary requirement is untouched.** *"**Secondary: the failure is invisible** … Logging the established connection — or surfacing rapid connect→close cycling as its own signal."* Nothing in the diff touches `secure-mqtt.ts`; the only log remains `[smqtt] reconnecting`. The PR body does not claim it, so this is not a broken promise — but it is load-bearing for *this* change specifically. Given the PR's own central caveat is that the `default` and `eufy_life` policies are an inference and the fallback is what makes shipping safe, there is currently no way to tell from a log whether the fallback fired. The fix and its own confirmation signal are the same missing line.

**5. Minor, disclosed-adjacent.** The failed attempt's `client.end(true)` fires `close`, so the discarded transport re-emits `disconnect` to the consumer for a connection that never existed. The `error` leak is disclosed in the PR body; this one is not.

### Over-engineering pass

Net **-30 lines** available, none of it load-bearing: `mqttClientAppName` + the `EUFY_MEGA` const collapse to one `??` at the single call site; `generateMqttUuid` goes with finding 3; `connectMqtt`'s JSDoc restates four lines already given on `startMqtt`; two specs assert one thing. All inline.

Explicitly **not** flagged: the `try`/`catch` fallback and `private openudid` → `readonly` are both the *lazy* choice — a fallback beats a new config option, an existing field beats a new one. Keep both.

### Verdict

Requesting changes on findings 1 and 2 only. Both are addressable without redesign — 2 is a code fix or an honest JSDoc, and 1 may be purely a documentation matter if per-caller `openudid` is the intended contract, in which case the JSDoc and PR body should say so outright instead of describing the id as per-install. Everything else is optional.

<sub>Standards: 3 hard, 4 judgement. Spec: 4 findings. Over-engineering: -30 lines.</sub>

martijnpoppen:
Both blocking findings were right, and both were the same mistake: prose asserting more than the code does. Pushed `59ee69c`.

**1. `openudid` is not per-install when the caller does not set it.** Correct — `md5("eufy-mega:" + email)` is account identity, so two unconfigured clients still collide and the motivating case is not reached by default. I narrowed the claim rather than the derivation, which is the reading you allowed: per-caller `openudid` is the contract. It now says so on both declarations — `startMqtt` ("separates two clients exactly as far as their `openudid` does"), and `MegaHttpClient.openudid`, which names the two things keyed on it and that both fail the same way when it is shared. A spec asserts the limit outright: two clients configuring none get the same uuid. Changing the fallback to a persisted random is an HTTP-identity change with 2FA consequences, as you say — its own commit, not this one.

**2. "The fallback is not remembered" was false.** Also correct, and worse than a wrong sentence: an unconditional `catch` meant one `ECONNRESET` moved a client onto the shared name and the memo kept it there for the process — the collision this PR exists to remove, reached by accident. Fixed in code rather than prose: `isNotAuthorized` (in `secure-mqtt.ts`, since the CONNACK is the transport's vocabulary) gates the fallback, and anything else rejects the bring-up, whose memo `ensureMqttStarted` then clears. The JSDoc now says the fallback holds until `disconnect()`, which is what happens.

**The trade that narrowing makes, stated plainly:** a policy refusal that arrives as a bare TCP reset instead of a CONNACK no longer falls back, so that line gets no realtime at all until someone looks. I think that is the right side to fail on — it is loud and fixable with evidence in hand, where the pin was silent and permanent — but it is a real narrowing of the safety net, and the JSDoc on `isNotAuthorized` says so.

**3.** Done. `command-router.ts` takes `mqttUuidFrom(this.deps.mega.openudid)`; `generateMqttUuid`, the field and the lazy init are gone, and the one-shot ids stop changing every process. A spec pins that.

**4.** Added — `[smqtt] connected (<app_name>)` at info, beside the existing reconnect warn. You were right that it is load-bearing here specifically: without it a log cannot show that the fallback fired. It carries the credential scope, not the client id, which contains the account id.

**5.** Fixed together with the disclosed `error` leak: the fan-out is wired once the connection stands, so a discarded attempt reports no connect, no disconnect and no error for a connection nobody had. The connect it just completed is announced explicitly; errors raised during the handshake are held only to keep the emitter from throwing, and reported once the transport is one a consumer owns.

**Over-engineering pass.** Took all of it: `mqttClientAppName` and `EUFY_MEGA` collapse into the `??` at the single call site (`topics.ts` is untouched now), `connectMqtt`'s JSDoc keeps only the paragraph that describes `connectMqtt`, the two `app_name` specs are one precedence spec, and `perInstall` is `ownClientId`.

**Specs.** You were right that they proved less than their titles. They now drive `ensureMqttStarted`, since the memo is what the claims are about, and reach the derivation the old ones stepped over: a restart keeps its id (via a store that already holds one), two unconfigured clients share one, a refusal falls back and is kept, a reset does not fall back and the next bring-up asks under the client's own id again. On `makeClient`'s reach-ins — it grew the config seam you predicted rather than a fourth `as any`; the three that were there are unchanged.

`npm run verify` green: 3497 tests. Re-synced into a consuming app and re-ran its gate too.

---

lenoxys (approved):
Re-reviewed at `59ee69c` in a clean worktree, checking the code rather than the replies. Both blockers are genuinely closed, and the second one in the way that matters — in code, not in prose.

**Finding 2** is the substantive fix: `isNotAuthorized(e)` gates the fallback, anything else rethrows, and `ensureMqttStarted`'s catch then clears its memo, so a dropped socket can no longer pin a process to the shared name. The spec drives the ECONNRESET path through the real bring-up and asserts the next attempt goes out under `android-…`, which is the assertion the old suite was structurally unable to make. `isNotAuthorized` living in `secure-mqtt.ts` is the right home — the CONNACK is transport vocabulary — and its JSDoc names the residual hole (a refusal arriving as a bare reset) instead of leaving it to be discovered.

**Finding 1** takes the documentation route, which is what I said would settle it, and takes it properly: the limit is stated on `startMqtt`, on `MegaHttpClient.openudid`, in the PR body twice, and — the part that makes it durable — asserted as behaviour in a spec, so two unconfigured clients sharing one uuid is now a pinned fact rather than an unexamined default. Leaving the derivation alone is correct; a persisted random there is an HTTP-identity change with 2FA consequences and deserves its own commit and its own review.

The rest all landed: the one-shot path shares `mqttUuidFrom` and `generateMqttUuid` is gone from the surface entirely (checked `dist`), `[smqtt] connected (<app_name>)` gives the fallback a signal it can be seen by, the fan-out is wired only once a connection stands so a discarded attempt reports nothing it shouldn't, and the whole over-engineering pass was taken. `npm run verify` green here too — 3497.

One nit, not worth a round trip and not a reason to hold this: the title still says "by install" when the correction's whole point is that the id separates *clients* exactly as far as their `openudid` does. The body and both JSDocs are accurate; only the title lags. Worth fixing in the squash message if you're editing it anyway.

Approving. The `openudid` default and #159's flap-visibility signal are both worth their own issues rather than riding along here.

PR #162, opened 2026-09-12
feat(p2p): typed refusals for a station, and traces for the waits before them

lenoxys:
## Why

Nothing reaches a station before its session connects, and an attached camera's media start has no unencrypted form, so it needs the level-2 key too. Both waits could end with no record at all, and every ending was the same plain `Error`. Seven causes, one string, none naming a station.

| Ending | Whose problem |
|---|---|
| the session never connected | station or network |
| the station never answered its gateway-info prompt | station or network |
| the account holds no material for the cipher the station named | the account |
| the cloud cipher lookup failed | the cloud, transiently |
| material for a different cipher was used, and derived to nothing | a lookup answering off-target |
| the station's reply would not derive | firmware or material |
| the session closed under the call | timing |

A call that failed during resolution sent no media command either, so nothing recorded the topology it was resolved under.

## What

Typed refusals: `StationUnreachableError` carrying `waitedMs`, and `StationKeyUnavailableError`. Both name the station, which for an attached camera is the parent, so a caller cannot recover it from the serial it passed in.

Phases, each emitted where the fact is known rather than where it is missed:

- `session-connect-wait`, then `session-connected` or `session-unreachable`. Emitted only where a wait happens, so absence states the session was already up. The leading one is the only record a caller gets when its own deadline expires inside the wait.
- `station-resolved`: topology, channel, and whether the signed-in account administers the station. `unstated` is a record naming no administrator, which is not the same as naming another.
- `level2-negotiating`: the station answered its prompt and a derivation began under the cipher it named. Separates a silent station from one that answered and produced nothing usable.
- `level2-unavailable` with `no-cipher-key`, `derivation-failed`, `not-negotiating`, `session-closed` or `grace-elapsed`. Every ending of the wait carries one, including the ordinary ending: a station prompted on connect that never replies. Replaces `level2-absent`, which named the same fact.
- `cipher-fallback`: the lookup answered with a different cipher's material, which was used and derives to nothing. The only lookup outcome no other phase already accounts for.

`P2P_STATION_WAITS` publishes `connect`, `level2Grace` and `level2Settle`. A caller bounding one of these calls is racing them, and a bound below them reports its own expiry in place of the reason. Otherwise that bound is a literal copied into caller source that goes stale silently.

## Breaking, hence 0.2.0 rather than a patch

`LiveTrace` gains seven phases and loses `level2-absent`. A caller with an exhaustive phase table gets a compile error, which is the point of publishing the union.

`P2P session for ... did not connect` and `level-2 key not ready for ...` become typed errors. No shim, per this repo's rule below 1.0.

## Verified

`npm run verify` green, about 3500 tests over 186 files. No new wire is sent: this names what already happened, so nothing here needs a device to confirm.

Contracts worth naming:

- A disconnected session driven through a real media acquisition. Wait traced first, outcome last, `StationUnreachableError` raised, no media command attempted. An already-connected session emits none of the three connect phases.
- A level-2 wait that runs out yields exactly one `level2-unavailable` with `grace-elapsed` and its `waitedMs`, the ending that previously produced nothing at all.
- `cipher-fallback` fires on an off-target answer and stays silent on an exact one.
- `station-resolved` pinned on all three administrator answers, including a record naming none.
- Traces are collected off the logger, so the specs see what a host sees, including the handle that groups one attempt.


src/transport/p2p/p2p-session.ts
```diff
@@ -475,9 +480,15 @@ export class P2PSession extends EventEmitter {
...
   async awaitLevel2Key(graceMs: number, graceFrom: "call" | "session" = "call"): Promise<boolean> {
-    if (this.closed) return false;
+    if (this.closed) {
+      this.trace({ phase: "level2-unavailable", reason: "session-closed" });
+      return false;
+    }
     if (this.level2Key) return true;
-    if (!this.level2Pending) return false;
+    if (!this.level2Pending) {
+      this.trace({ phase: "level2-unavailable", reason: "not-negotiating" });
```
lenoxys, line 488: **The headline row of the table still emits nothing.**

Only the two pre-wait guards trace. A station that never answers its prompt goes `level2-wait`, 25s, `false`, `StationKeyUnavailableError`, with no terminal phase. Twice, with the reprompt.

That is the common case and the one still unaccounted for. Trace the outcome where the wait resolves, not only where it is refused before waiting: `timeout` and `closed` both need a reason there.

src/transport/p2p/live-trace.ts
```diff
@@ -36,18 +36,86 @@ export type LiveTrace =
...
+   * negotiated from one whose cloud cipher lookup answered nothing, from one whose reply would not derive.
+   * Each calls for a different next step, and only `no-cipher-key` and `derivation-failed` are about this
+   * account's own cipher material.
+   *
+   * `cipherId` states which cipher the station asked for, where a station asked at all: it is what says whether
+   * the account was resolved for the cipher in question. The two reasons answered without a negotiation have
+   * none to name.
+   */
+  | {
+      phase: "level2-unavailable";
```
lenoxys, line 78: **Two vocabularies for one fact.**

`level2-absent` and this both mean "no key". Fold `absent` in as a reason (`grace-elapsed`) instead of leaving the old shape beside the new. Same edit as tracing the resolved wait.

src/transport/p2p/command-router.ts
```diff
@@ -390,13 +409,19 @@ export class P2PCommandRouter {
...
           const ciphers = await this.deps.mega.getCiphers([cipherId], adminUserId, stationSn);
-          ecc = ciphers.find((c) => Number(c.cipher_id) === cipherId)?.ecc_private_key ?? ciphers[0]?.ecc_private_key;
+          const named = ciphers.find((c) => Number(c.cipher_id) === cipherId)?.ecc_private_key;
+          ecc = named ?? ciphers[0]?.ecc_private_key;
+          outcome = named !== undefined ? "exact" : ecc !== undefined ? "fallback" : "none";
         } catch (e) {
+          outcome = "failed";
           this.deps.onError(e instanceof Error ? e : new Error(String(e)));
         }
+        traceLiveStart(logger, { phase: "cipher-lookup", outcome, cipherId });
```
lenoxys: **Emitted without the session's `traceId`.**

The only phase not under the station's handle, and the one you would line up against `level2-negotiating`. `traceId` is already public, so pass it as the source.

src/transport/p2p/live-trace.ts
```diff
@@ -36,18 +36,86 @@ export type LiveTrace =
...
+  /**
+   * The cloud lookup for a station's cipher material answered, and what it answered with.
+   *
+   * A key that never arrives has three causes on this side and they are not each other's: the account holds no
+   * material for the cipher the station named, the lookup itself failed, or material for a DIFFERENT cipher was
+   * used because the named one was absent from the answer — which then derives to nothing and looks like a
+   * station fault. `cipherId` is the cipher the station asked for, so this pairs with the negotiation that
+   * asked for it.
+   */
+  | { phase: "cipher-lookup"; outcome: "exact" | "fallback" | "none" | "failed"; cipherId: number }
```
lenoxys: **Three of the four outcomes are redundant.**

`exact` duplicates `level2-ready`, `none` duplicates `no-cipher-key`, `failed` already reaches `onError`. Only `fallback` is new information. Fire on that alone: one trace, three fewer literals to keep exhaustive.

The cache-hit path traces nothing either, so today the outcome disappears on every negotiation after the first.

lenoxys (commented):
Solid direction, and the six phases mostly pay for themselves. Four things before merge, marked inline: the common level-2 outcome still emits no phase, `level2-absent` and `level2-unavailable` are two vocabularies for one fact, `cipher-lookup` ships without the session's `traceId`, and three of its four outcomes are redundant. The rest are small.

src/transport/p2p/p2p-session.ts
```diff
@@ -435,8 +435,13 @@ export class P2PSession extends EventEmitter {
...
 
-  /** Emit a live trace under this session's handle. */
-  private trace(trace: LiveTrace): void {
+  /**
+   * Emit a live trace under this session's handle.
+   *
+   * Public so work staged on a session outside it — reaching the station before anything can be addressed to
+   * it — is recorded under the same handle as everything that follows, which is what groups one attempt.
+   */
+  trace(trace: LiveTrace): void {
```
lenoxys: **A permanent write handle on an exported class, for one caller.**

The router already imports `traceLiveStart` and `traceId` is already public. `traceLiveStart(logger, trace, session.traceId)` gets the same grouping with no surface change.

src/transport/p2p/command-router.ts
```diff
@@ -1255,12 +1280,32 @@ export class P2PCommandRouter {
...
+      session.trace(
+        session.isConnected
+          ? { phase: "session-connected", waitedMs: Date.now() - t0 }
+          : { phase: "session-unreachable", waitedMs: Date.now() - t0 },
+      );
     }
     opts.signal?.throwIfAborted();
-    if (!session.isConnected) throw new Error(`P2P session for ${parentSn} did not connect`);
+    if (!session.isConnected) throw new StationUnreachableError(Date.now() - t0);
+    const stationAdmin =
```
lenoxys: `admin_user_id` is re-derived here with three casts, two lines under where `accountId` already derived it. One local, both uses.

`Date.now() - t0` is computed three times as well, so the traced `waitedMs` and the one carried by the error can disagree. Hoist it.

src/transport/p2p/command-router.ts
```diff
@@ -106,6 +107,24 @@ const LEVEL2_GRACE_MS = 25_000;
...
+ * A caller that bounds one of these calls itself races these waits, and a bound below them reports the
+ * caller's own expiry in place of the reason this SDK was about to give — the two are indistinguishable to
+ * whoever reads the outcome, and they call for different next steps. Published so that bound can be derived
+ * rather than copied: a literal in a caller's source is a second source of truth that goes stale silently
+ * when these change.
+ *
+ * `connect` applies to every call on a station, because nothing can be addressed to one before its session is
+ * up. `level2Grace` applies twice where the key is required: the negotiation is re-prompted once.
+ */
+export const P2P_STATION_WAITS = Object.freeze({
```
lenoxys: `as const` covers callers. The `Object.isFrozen` assertion beside it tests V8, not this SDK.

src/core/contracts.ts
```diff
@@ -106,6 +106,54 @@ export class CameraDisabledError extends Error {
...
+ * A station reached over its HomeBase encrypts what it is sent under a key negotiated once per connection, and
+ * a media start for an attached camera has no unencrypted form at all — so without that key there is nothing
+ * to send, however reachable the station is. Naming this apart from a source that failed is what separates an
+ * account whose cipher material could not be resolved from a camera that is off, a station that is busy, or a
+ * stream that produced nothing: they share no next step.
+ *
+ * The `level2-unavailable` trace states WHY the key is not coming. This states only that it is not, because
+ * that is what the refusal itself knows.
+ */
+export class StationKeyUnavailableError extends Error {
```
lenoxys, line 125: Both new errors name no station, where the strings they replace did, and the traces carry `station-N` rather than the serial. Deliberate?

src/transport/p2p/__tests__/session-fixtures.ts
```diff
@@ -48,6 +51,18 @@ export function connectedSession(hasLevel2Key = true): FakeP2PSession {
   return session;
 }
 
+/**
+ * A session that has not connected, and does not while a spec waits on it.
+ *
+ * Every call on a station holds for this before anything is sent, so it is the state that separates a station
+ * that could not be reached from one that answered and refused.
+ */
+export function disconnectedSession(): FakeP2PSession {
```
lenoxys, line 62: This is `connectedSession(false)` with a flag flipped, and the already-connected spec flips it back, so it reads as a key-less session by accident. Use `connectedSession()` in that spec.

src/core/contracts.ts
lenoxys: Not deliberate, fixed in b30673f. `stationSn` is the first argument of both and appears in the message. `ResolvedSession` already carried `parentSn`, so both throw sites had it.

What decided it: for an attached camera the station is the parent, so a caller cannot recover it from the serial it passed in. Several cameras refused together are one station's outcome.

max246 (approved):

PR #164, opened 2026-09-13
feat(vacuum): send a scene, a room set and a zone set

martijnpoppen:
## What changed

`startScene`, `cleanRooms` and `cleanZones` on `vacuum_clean`, gated on `isAiotVacuum` like every other DP 152 verb. `VacuumRoomTarget` and `VacuumZoneTarget` join the public type exports. `docs/vacuums.md` gains a Controls section.

## Why

The three area-selecting `ModeCtrlRequest` verbs were encoders with no way to reach them: `encodeSceneClean`, `encodeSelectRoomsClean` and `encodeSelectZonesClean` built byte-proven frames, no member named one, and a spec asserted the setters did not exist.

The note above `ModeCtrlParamMethod` gave the reason — each needs an argument "this SDK cannot yet answer: a room or zone id comes from map data, which is not decodable here". **For two of the three that is no longer true.** A scene id is `VacuumScene.id` off the `SceneResponse` on DP 180, and a map id is on that same report and on a scheduled rooms-clean. So each verb's argument now comes from a read this SDK already publishes rather than from something a caller has to invent. `GOTO` keeps no encoder, because a goto point is a coordinate no read here supplies.

## Evidence

**Frames: proven.** The outer `ModeCtrlRequest` is byte-verified against a live T2351 capture, and the existing byte-level specs hold the writer to that wire. The `Param` payload shapes each have their own spec — room entries repeating with the map id beside them, zone corners ZigZagged so a negative one stays one byte and stays negative, a scene carried by its id alone.

**Method numbers: run on a T2351.** All three verbs were exercised on a live T2351 and each did what it names — a scene started, the named rooms were cleaned, the given rectangles were cleaned. So methods 24, 1 and 2 rest on observed behaviour, not on the vendor's `ModeCtrlRequest.Method` definition they were first read from. Same model as the frame capture, so one robot stands behind both halves of every mode-control write here.

That check was the one that mattered, and no amount of frame verification substitutes for it: an AIoT data-point write is fire-and-forget, so a wrong number would have been a *different* command arriving and looking exactly like success. `ModeCtrlParamMethod`'s type doc owns that argument, since it owns the numbers.

The clean parameters are untouched and stay read-only: they decode out of one reported payload and no capture pins a write for any of them, so there is no setter to offer.

## Verified how

- [x] `npm run verify` is green locally — 183 files, 3504 tests, all guards, `check:esm`, examples
- [x] **Ran against a real device — yes: a T2351, all three verbs, each performing the action it names**
- [x] Docs updated

The spec that asserted these setters did not exist is inverted: it now checks all three bind on an AIoT robot, none on a Tuya one, and that no `setCleanRooms` property-setter spelling appears beside them. A new spec drives `startScene(7)` through the bound verb and decodes the dispatched frame, which is what ties the published read (`VacuumScene.id`) to the verb that takes it.

`docs/vacuums.md` ended on *"Controls … are not implemented; the robot surface is read-only today"*, which was already false for start, pause, resume, return-to-dock, suction and volume before this change. It now documents what the line can be told to do, grouped by what a verb does — whole-floor, then part of a floor — and says which claim rests on what: the mode-control verbs were watched on a T2351, while the two suction setters rest on a data-point write the SKU's own catalog confirms.

## Unverified or deferred

- `GOTO` (method 4, param 7) is still listed and still has no encoder. Its gap is a different one from the others': its argument is a coordinate, and no read published here supplies one, so there is nothing for a caller to pass.
- Clean-parameter writes, as above — read-only, with no capture pinning a write.
- The nine parameterless `unverified` mode-control verbs are untouched. Promoting those on evidence is its own PR, and the same kind of run would settle them.

## Review

Ten threads from @lenoxys across two passes, all addressed.

**First pass — the blocker.** Shipping three uncaptured method numbers callable contradicted the rule its `unverified` siblings follow, and that was correct on the evidence as it stood. It is answered by the hardware run rather than by an exemption: a verb run on real hardware doing what it names is the promotion condition the rule states, so no precedent splits. Also fixed: two guide snippets that did not compile (`setLevel` → `setSuctionLevel`; `scenes` is a method, so `scenes?.()`), and an orphaned JSDoc block — two of them, in fact, the second being `encodeModeCtrl`'s stranded ~170 lines from its function.

**Second pass — how the claim is written**, and each of the four was right:

- **"A live robot" named no model**, alone among the claims in that file, and was the one form nobody could check later. It is a T2351 now, in all three members. I asked rather than inferred it.
- **`cleanRooms` was replaying the review inside the shipped doc.** Cut to the fact. That also removed a sentence that was wrong as well as surplus: its list of confirmed verbs started at 6 and dropped method 0, which is live-verified too.
- **The fire-and-forget argument had three homes in one PR.** One now, in the type doc that owns the numbers.
- **The guide's lead over-claimed** — "every verb below" swept in the suction setters, a loose claim inside a paragraph about claim precision. Narrowed, with the suction setters' different basis stated.


lenoxys (commented):
The frame work is solid and the map-id reasoning is right. The central call is what I would hold: shipping three uncaptured method numbers callable contradicts the rule and the sibling members in this same table. Two of the doc snippets also do not compile. Details inline.

src/model/capabilities/vacuum-clean.ts
```diff
@@ -2906,6 +2915,61 @@ export const VACUUM_CLEAN_MEMBERS = {
...
+   * The id is the device's own, as {@link VACUUM_CLEAN_MEMBERS.scenes} reports it — `VacuumScene.id`
+   * off the `SceneResponse` on DP 180. A scene the device reports invalid stays reportable and running
+   * it is still a well-formed request; `VacuumScene.invalidReason` says why the device will refuse.
+   *
+   * Frame shape is byte-proven against the shared outer `ModeCtrlRequest`. The method NUMBER is the
+   * vendor's own `ModeCtrlRequest.Method` value and is not capture-confirmed; on a fire-and-forget DP
+   * write a wrong number is a different command, so it is a maintainer decision that this ships
+   * callable rather than as an unverified write.
+   */
+  startScene: method(
```
lenoxys, line 2928: **This is the one I would hold.** The reasoning is written out honestly, but the same table already answers it the other way.

`stopCleaning`, `startSpotClean`, `goWash`, `startGlobalCruise`, `startFastMapping`: every verb whose number is the vendor's enum and nothing more carries `unverified: true`, under a JSDoc that says what this one says. "What is unconfirmed is the number alone. That still keeps them `unverified`: a wrong number is a different command arriving at real hardware, and an AIoT DP write is fire-and-forget, so a mistake looks exactly like success."

The evidence position of these three is identical and the consequence is worse. 1 and 2 are low numbers in a space where the captured verbs sit at 6, 13 and 14, and these carry an argument, so a mis-numbered frame is a real command with a room set attached.

Making it a per-PR judgement is how the rule stops being a rule: the next contributor reads two precedents and picks. Mark all three `unverified` and this becomes the same PR minus the risk. The encoders, the types, the docs and the pairing with `scenes()` all still land, and promotion is one word per verb once a capture arrives.

If the argument is that these three are worth more callable than the parameterless nine, that is a case for promoting all twelve on evidence, not for exempting three by decision.

src/model/capabilities/vacuum-clean.ts
```diff
@@ -2906,6 +2915,61 @@ export const VACUUM_CLEAN_MEMBERS = {
...
+  cleanZones: method(
+    ({ sink }) =>
+      (mapId: number, zones: readonly VacuumZoneTarget[]): Promise<void> =>
+        sink.dispatch(aiotDp(VACUUM_DP.MODE_CTRL, encodeSelectZonesClean(mapId, zones))),
+    "Clean the given rectangles of a named map (ModeCtrlRequest method 2 over DP 152).",
+    isAiotVacuum,
+  ),
 } as const satisfies Members;
 
 /** `vacuum_clean` — core RoboVac scalar state + decoded activity: power, activity, volume, battery. */
```
lenoxys, line 2974: Consequence of the above: an `unverified` member has no setter and its intent path throws, so the spec inversion goes the other way too. Assert the three do not bind on AIoT, and that the intent route throws.

The `startScene(7)` frame spec stays as written. It tests the encoder, which is the proven part.

src/model/capabilities/vacuum-clean.ts
```diff
@@ -266,11 +266,16 @@ export const ModeCtrlMethod = {
...
- * coordinate is signed centimetres in a frame no capture has pinned. Listing their numbers beside the
- * parameterless ones would invite a caller to send one with an empty payload, which is a valid frame
- * meaning something nobody intended.
+ * Kept out of {@link ModeCtrlMethod} because each is meaningless without an argument, and listing its
+ * number beside the parameterless verbs would invite a caller to send it with an empty payload — a
+ * well-formed frame meaning something nobody intended. Each verb built on one of these therefore takes
+ * its argument in the signature: {@link VACUUM_CLEAN_MEMBERS.startScene},
+ * {@link VACUUM_CLEAN_MEMBERS.cleanRooms} and {@link VACUUM_CLEAN_MEMBERS.cleanZones}.
+ *
+ * The numbers are the vendor's own `ModeCtrlRequest.Method` values and are NOT capture-confirmed; the
```
lenoxys: This block is orphaned. It sits between `encodeModeCtrlRequest`'s JSDoc and a second JSDoc, attached to no declaration, so nothing renders it and nobody editing `ModeCtrlParamMethod` sees it.

Pre-existing, but the PR is adding load-bearing evidence prose to it. Fold the two paragraphs into the `ModeCtrlParamMethod` doc below and drop the block.

docs/vacuums.md
```diff
@@ -121,8 +121,63 @@ advance throughout a run: since the realtime feed is the only place a robot's st
...
+
+```ts
+const clean = dev.vacuumClean();
+
+await clean?.startCleaning?.(); // whole-floor auto clean
+await clean?.pauseCleaning?.();
+await clean?.resumeCleaning?.(); // resumes where it stopped, unlike a fresh start
+await clean?.returnToDock?.();
+
+await dev.suction()?.setLevel?.(2); // raw level, see above
```
lenoxys: `setLevel` does not exist. The member declares `writeAs: "setSuctionLevel"`, so this snippet does not compile.

docs/vacuums.md
```diff
@@ -121,8 +121,63 @@ advance throughout a run: since the realtime feed is the only place a robot's st
...
+Each verb is optional on the surface, because whether a device has it is a runtime fact. The `?.` is
+not defensive style — it is the type telling you to check.
+
+### Cleaning part of a floor
+
+Three verbs take an argument, and each argument comes from a read the robot already publishes rather
+than from anything you have to invent:
+
+```ts
+const scenes = clean?.scenes; // decoded off the robot's own scene report
```
lenoxys: `scenes` is a method, not a property, so this assigns the function and the `.find` below it fails. Wants `clean?.scenes?.()`.

Worth a compile check over the guide's snippets generally. Both bugs here only a reader hits.

docs/vacuums.md
```diff
@@ -121,8 +121,63 @@ advance throughout a run: since the realtime feed is the only place a robot's st
 also the only way to follow a clean in progress. They are payload-backed on the AIoT line, so those
 announcements name the property with no value — re-read it through the accessor.
 
-**Controls** — start / pause / return-to-dock, suction, volume — are not implemented; the robot surface
-is read-only today.
+## Controls
```
lenoxys, line 124: Whatever the outcome above, this section should split by evidence instead of listing verbs flat. The four captured ones read as equal to three uncaptured ones.

A caller reading this cannot tell which writes are proven, and if the three stay `unverified` they are absent from the bound object, so the section documents an API that is not there.

src/model/capabilities/vacuum-clean.ts
martijnpoppen: You are right that the precedent cannot be per-PR judgement, and the evidence has since moved — so the split is now by evidence, which is the rule you are asking for rather than an exemption from it.

**`startScene` has been run on hardware.** The maintainer exercised method 24 against a live robot and the named scene started. That is the capture-equivalent you asked for on the one verb the read side could actually drive end to end (a scene id comes from `scenes()`; a room or zone id still has to be supplied by hand, which is why those two are harder to exercise). Its JSDoc now records the observation instead of the vendor's enum, and the guide places it with `startCleaning`, `pauseCleaning`, `returnToDock` and the suction setters.

**`cleanRooms` and `cleanZones` keep the vendor's numbers**, and both the member docs and the guide now say so in your terms — 1 and 2 are low numbers in a space whose confirmed verbs sit at 6, 13, 14 and 24, both carry a payload, and an AIoT DP write is fire-and-forget, so a mis-numbered frame is another command arriving with a room set attached and looking exactly like success. They stay callable as the author's decision with that written down, not as an oversight and not as a claim of equal evidence.

So this is not "two precedents, pick one": watched number → callable, unwatched number → callable only where the author has signed for it, and the doc says which is which at the member and in the guide. On your alternative — promoting all twelve on evidence — that is the right shape and worth its own PR; a capture of the vendor app running a room clean promotes these two with it.

---

src/model/capabilities/vacuum-clean.ts
martijnpoppen: Fixed, and there were two orphans rather than one. The block you flagged duplicated what `ModeCtrlParamMethod`'s own doc says, so its substance (the member links, and now the per-number evidence) is folded into that doc and the block is gone. The comment directly below it is `encodeModeCtrl`'s, stranded ~170 lines above the function it describes — same defect, same region, so it is moved onto the declaration rather than left for the next reader to trip over.

---

docs/vacuums.md
martijnpoppen: Fixed — `setSuctionLevel`, per the member's `writeAs`.

---

docs/vacuums.md
martijnpoppen: Fixed — `clean?.scenes?.()`.

On compile-checking the guides generally: agreed, and both bugs here are the argument for it — neither is reachable by any test in the repo, and a reader is the one who finds them. It is a bigger change than this PR should carry (extracting every fenced `ts` block across `docs/` and typechecking it needs a harness and a way to mark the deliberately-partial snippets), so I have not bundled it in. Worth its own issue.

---

docs/vacuums.md
martijnpoppen: Done — Controls now splits into **Confirmed on hardware** (the four parameterless verbs, the two suction setters, and `startScene`, whose method 24 was run on a live robot) and **Number taken from the vendor's enum** (`cleanRooms`, `cleanZones`, with what that costs stated where the snippet is). The lead paragraph says outright what the grouping is for: every frame is byte-proven against the same captured `ModeCtrlRequest`, so what separates the groups is whether the number inside has been watched doing what it claims.

Your second point stands whichever way the members had gone, and is why the section is structured this way rather than annotated. `GOTO` also gets the line it was missing, since its gap is a different one — no read here supplies a coordinate, so it is not awaiting a capture the way the other two are.

---

src/model/capabilities/vacuum-clean.ts
martijnpoppen: Following from the thread above: the three still bind, so the spec inversion stays as written — it checks all three bind on an AIoT robot, none on a Tuya one, and that no `setCleanRooms` property-setter spelling appears beside them. The `startScene(7)` frame spec stays too, and is now the tighter of the two claims: it drives the verb the hardware run confirmed.

If a capture ever contradicts method 1 or 2, flipping either to `unverified` takes the member's own word plus that assertion, which is the shape you are describing and is cheap to reach from here.

---

src/model/capabilities/vacuum-clean.ts
martijnpoppen: Update, and it resolves this rather than arguing it: **all three verbs have now been run against a live robot** and each did what it names — a scene started, the named rooms were cleaned, the given rectangles were cleaned. So methods 24, 1 and 2 rest on observed behaviour, not on the vendor's enum, and the rule you were defending is satisfied on its own terms. Nothing is exempted and no precedent splits.

Your objection was the right one on the evidence as it stood, and it is worth saying that the reasoning you spelled out is what the code now records. Each member's note keeps the per-verb observation rather than deferring to one line on the table, precisely because your point about 1 and 2 outlives the check: they are low numbers in a space whose other confirmed verbs sit at 6, 13, 14 and 20, and both carry a payload, so a mis-numbered frame would have been another command arriving with a room set attached — invisible on a fire-and-forget write, and not something any amount of frame verification would have caught.

The guide's evidence split has collapsed with it: Controls now groups by what a verb does (whole-floor, then part of a floor) and states the evidence once. PR description updated with the hardware result.

Your wider point stands for the nine parameterless `unverified` verbs, which this branch does not touch — promoting those on evidence is its own PR, and the same kind of run would settle them.

---

src/model/capabilities/vacuum-clean.ts
martijnpoppen: Moot now, per the thread above: all three were run on a live robot, so none becomes `unverified` and the members keep their setters. The spec inversion stays as written — all three bind on an AIoT robot, none on a Tuya one, no `setCleanRooms` property-setter spelling beside them — and the `startScene(7)` frame spec stays too.

---

src/model/capabilities/vacuum-clean.ts
```diff
@@ -2906,6 +2911,64 @@ export const VACUUM_CLEAN_MEMBERS = {
...
+
+  /**
+   * Run a saved cleaning scene by its id (ModeCtrlRequest method 24 over DP 152).
+   *
+   * The id is the device's own, as {@link VACUUM_CLEAN_MEMBERS.scenes} reports it — `VacuumScene.id`
+   * off the `SceneResponse` on DP 180. A scene the device reports invalid stays reportable and running
+   * it is still a well-formed request; `VacuumScene.invalidReason` says why the device will refuse.
+   *
+   * Frame shape is byte-proven against the shared outer `ModeCtrlRequest`, and method 24 has been
+   * WATCHED: run against a live robot, it started the named scene.
```
lenoxys: Name the robot. Every other claim in this file names the model it was seen on, and the outer frame's own claim is "a live T2351". "A live robot" is the one form of the sentence that cannot be checked later.

If the run was on a different model than the capture, that difference is the interesting part of the claim, not a detail to smooth over.

src/model/capabilities/vacuum-clean.ts
```diff
@@ -2906,6 +2911,64 @@ export const VACUUM_CLEAN_MEMBERS = {
...
+   *
+   * `mapId` has no default and that is deliberate: room ids are per map, so assuming the map a
+   * single-floor home would have sends a two-floor home's ids against the wrong floor. `SceneInfo.mapid`
+   * on DP 180 and a scheduled rooms-clean's `map_id` are the two real map ids the device reports.
+   *
+   * `cleanTimes` is how many passes to make over the set; rooms with no `order` are visited in the
+   * order given.
+   *
+   * Frame shape is byte-proven against the shared outer `ModeCtrlRequest`, and method 1 has been
+   * WATCHED: run against a live robot, it cleaned the rooms named. Worth recording per verb rather than
```
lenoxys: From "Worth recording per verb" onward, this is commentary about where documentation should live, and it replays the review inside the shipped doc. The reader of this member needs the fact: method 1, run on <model>, cleaned the rooms named.

The list also drops method 0, which is live-verified too, so the sentence is wrong as well as surplus.

src/model/capabilities/vacuum-clean.ts
```diff
@@ -263,36 +263,24 @@ export const ModeCtrlMethod = {
...
  * Kept apart from {@link ModeCtrlMethod} because these are a different kind of thing: a parameterless
  * verb is complete on its own, whereas each of these is meaningless without an argument the caller has
  * to supply. Sending one with an empty payload is a well-formed frame that means something nobody
- * intended, which is exactly why the numbers do not sit beside the others.
+ * intended, which is exactly why the numbers do not sit beside the others. Each verb built on one takes
+ * its argument in the signature: {@link VACUUM_CLEAN_MEMBERS.startScene},
+ * {@link VACUUM_CLEAN_MEMBERS.cleanRooms} and {@link VACUUM_CLEAN_MEMBERS.cleanZones}.
+ *
+ * The outer frame all four ride in is byte-verified on a live T2351, and `SCENE`, `SELECT_ROOMS` and
+ * `SELECT_ZONES` have each since been RUN against a live robot and did what they name — so their numbers
```
lenoxys: The fire-and-forget argument now lands three times in one PR: here, in `cleanRooms`, and in the guide's lead paragraph. One home. This type doc is the right one, since it owns the numbers.

Also "all four" reads oddly when the sentence goes on to exclude `GOTO` two paragraphs later.

docs/vacuums.md
```diff
@@ -121,8 +121,84 @@ advance throughout a run: since the realtime feed is the only place a robot's st
...
+
+Every write below is **AIoT-only**. The legacy Tuya clean line dispatches nothing at all — that
+direction has no live capture behind it — so a Tuya robot binds the reads and none of the verbs, and
+the missing method is the signal.
+
+Each verb is also optional on the surface, because whether a device has it is a runtime fact. The `?.`
+is not defensive style — it is the type telling you to check.
+
+**Every verb below has been run against a live robot and did what it says.** That is worth stating
```
lenoxys: "Every verb below" includes the two suction setters, which rest on a schema-confirmed data-point write rather than on a watched run. Narrow it to the mode-control verbs, or the paragraph about claim precision makes a loose claim.

Minor, same section: "anything a caller has to invent" sits four lines from "the type telling you to check". Pick one voice.

lenoxys (commented):
The blocker is answered on its own terms. A verb run on real hardware doing what it names is the promotion condition the rule states, so no exemption is needed and the precedent holds. Doc bugs fixed, orphan folded, and the second orphan you found beside it was a good catch.

What is left is how the claim is written, not whether it holds.

src/model/capabilities/vacuum-clean.ts
martijnpoppen: Named: **T2351**, in all three members. I asked rather than inferred it — a model code inside an evidence claim is the last thing to guess at.

You were right that it was the one form of the sentence nobody could check later, and right that the answer mattered either way. It happens to be the same robot as the frame capture, which makes the claim simpler rather than more interesting: one model behind both halves of every mode-control write in this PR.

---

src/model/capabilities/vacuum-clean.ts
martijnpoppen: Cut to the fact: method 1, run on a T2351, cleaned the rooms named.

You are right about what it was — I was replaying this review inside the shipped doc, and a reader of that member has no stake in where documentation should live.

And thank you for the method 0 catch. It was wrong as well as surplus: `START_AUTO_CLEAN` is live-verified on the same T2351 and is simply omitted from the wire when zero, so a list of confirmed verbs that started at 6 was missing the first one. The whole sentence is gone, so the error goes with it rather than getting a correction — but it is worth saying that the same list is what I would have reached for again.

---

src/model/capabilities/vacuum-clean.ts
martijnpoppen: Agreed on both, and this type doc is the home — it owns the numbers, so the argument about what a wrong number costs belongs with them. The copies in `cleanRooms` and in the guide's lead are gone; the three remaining mentions in this file are pre-existing and about different members, so I left them alone.

"All four" was also just wrong to write in a paragraph that excludes `GOTO` two paragraphs down. It names the three that have an encoder.

For what it is worth, this is the second review in a row to find the same habit in my writing — one argument restated at every site it touches. The cost is not only length: a claim in four places is four places to be inconsistent, which is exactly what the count error on #165 turned out to be.

---

docs/vacuums.md
martijnpoppen: Narrowed, and you picked the right paragraph to hold to its own standard. It now reads: every **mode-control** verb was run on a T2351 — the whole-floor four and the three that take an argument — while the two suction setters rest on a data-point write the SKU's own catalog confirms rather than on a watched run.

I checked "the whole-floor four" rather than trusting it, given the method 0 slip in the other thread: `START_AUTO_CLEAN` (0), `START_GOHOME` (6), `PAUSE_TASK` (13) and `RESUME_TASK` (14) are the live-verified mode-control methods, and those are exactly the four verbs that section shows. So the count holds.

Voice fixed too — second person, like the rest of the guide.

---

lenoxys (approved):
Approving. The three numbers rest on a run against a T2351 rather than on the vendor's enum, so the unverified rule is satisfied on its own terms and the precedent it sets is the one the table already had. Docs bugs gone, the argument lives in one place, and the guide separates the mode-control verbs from the two suction setters.

Checked on a trial merge with current beta-0.2.0 rather than on the branch alone: 187 files, 3522 tests, green.

PR #165, opened 2026-09-13
feat(display): give the Smart Display its own line, namespace and reads

martijnpoppen:
## What changed

`display` becomes its own **product line** and its own **param namespace**, and a `display` capability publishes one typed read: `battery`. Three further params are reachable without one. Read-only; no write is offered. New guide at `docs/smart-display.md`.

## Why

A T87A0 resolved to exactly one capability — `info` — whose values come from the cloud record rather than from the device, so `describe()` reported a unit that had told us nothing. Its param ids were being read against another line's dictionary.

Two defects behind that, both previously flagged in the source as open doors:

- **The param namespace.** The display reports ids 8001-8006 and those were looked up in `SECURITY_PARAMS`, which has never assigned anything in the 8000s. Nothing misdecoded today, but a future security param in that range would have been read off a Smart Display as whatever it means on a camera. `display` now owns a dictionary, the way `print` already did.
- **The product line.** This was the live half. The codec sat in `security`, detection evidence is OR-ed, and several security capabilities match on NAME alone — so any capability whose `modelHints` matched this device's reported text attached to it. Measured, not hypothetical: with an adversarial name, **six attached** (light, doorbell, leak, smoke, CO, lock), none of them answerable, because this device connects over secure MQTT with no `p2p_did` and speaks no P2P at all.

## Evidence

One live T87A0, captured 2026-09-04 (redacted), reporting six params. Four are named; what each rests on differs, and that is what decides how far it reaches:

| Param | Value | Named as | Provenance, and why | Reaches |
| --- | --- | --- | --- | --- |
| 8001 | `"100"` | `battery` | **`verified`** — identified by the maintainer. The value could not have said so (`100` fits brightness, volume or charge equally), so the name rests on someone's knowledge of the hardware: our own observation, not the cloud data-point list. | **typed getter** — the one read a caller cannot get another way |
| 8005 | `"Smart Display E10"` | `modelName` | `mega` — self-evidencing: the value is the model's retail name, a fact the registry already carried from another source. | dictionary only — `info` already answers it |
| 8006 | `"T87A0"` | `modelCode` | `mega` — same: the value is the device's own model code. | dictionary only — `info` already answers it |
| 8003 | `"2.9.05"` | `softwareVersion` | `guessed` — its shape, and nothing else. | **`unexposed`** — in the schema with its type and label, `getProperty` answers, no typed getter |

`unexposed` rather than dictionary-only for 8003 (see Review): dropping the member would lose its `PropertySpec`, so the schema would no longer carry the type or the `guessed` label, and the same fact would be answered differently depending on which side a caller asked from. `cameraInfo` (1103), `batteryPowerStats` (3100) and the keypad's battery pair are the precedent — each carries a dictionary entry **and** an `unexposed` member.

A dictionary entry is what makes a param readable by name: `applyParams` names a reported param from `paramDef` when no `PropertySpec` claims it, so all four arrive on `getProperties()` (and in `toJSON()`'s `properties`). Note `describe()` is a different surface — it returns a `DeviceManifest` keyed on `details`, and the three without getters do not appear there.

**8002 (`"1"`) and 8004 (serial-shaped) are deliberately unnamed.** `1` fits any enum or flag; a serial-shaped value could be the display's own or the station's it is bound to. They arrive as `unknown_8002` / `unknown_8004`, which is what let 8001 be identified in the first place.

**The line change is not a wire claim** — it narrows what may attach to this codec. The `POISONED` case in `line-partition.spec.ts` measured the old behaviour at six leaked capabilities and now asserts none.

**Nothing is writable, and nothing is guessed into the typed surface.** No capture pins a display write, and an AIoT data-point write is fire-and-forget: a wrong frame to a device that acknowledges nothing looks exactly like success.

## Verified how

- [x] `npm run verify` is green locally — 184 files, 3514 tests, all guards, `check:esm`, examples
- [ ] Ran against a real device — **no, and nothing here could be.** This PR adds no write, so there is nothing to drive. The reads are grounded in the capture above; `battery`'s meaning comes from the maintainer.
- [x] Added or updated specs — `display.spec.ts`, plus four existing specs whose assertions this deliberately moves
- [x] Docs updated — `docs/smart-display.md`, in the Control nav

The four moved assertions each pinned the behaviour being changed: `model.spec.ts`'s display case (`["info"]` → `["display", "info"]`), `line-partition.spec.ts`'s `POISONED` case (six security capabilities → none) and its declared-line list, `action-accessors.spec.ts`'s accessor set, and `property-id-integrity.spec.ts`, which now knows the display dictionary so the new ids count as grounded rather than as debt.

### The namespace switch is now tested by its consequence

Worth calling out because the first version of this PR got it wrong. The only assertion guarding the change was `namespaceForCodec("display") === "display"` — the table agreeing with itself. Reverting the mapping did fail it, but with `expected 'security' to be 'display'`, which says a constant changed rather than that anything broke.

`display.spec.ts` now runs the captured record through `Device.fromRecord` and asserts what `getProperties()` yields: `modelName` and `modelCode` named (neither has a capability spec, so the dictionary is the only thing that can name them, and only the `display` dictionary holds them), `battery` coerced to the number `100`, and 8002/8004 arriving as `unknown_8002` / `unknown_8004`. The consequence asserts first and the constant last, so a revert now fails with `expected undefined to be 'Smart Display E10'` — confirmed by performing the revert both before and after.

## Unverified or deferred

- **`softwareVersion` is an inference, not a mapping.** One dotted value on one device, `provenance: "guessed"`, no typed getter. `dev.info()?.firmwareVersion` is the field to trust where the cloud record carries one — this display's did not, which is the only reason 8003 is named at all.
- **The battery's scale is claimed, and here is the claim.** A full charge reads `255` on a 0-255 scale and `1000` on a 0-1000 one, so `"100"` on a charged unit is positive evidence for a percentage rather than merely consistent with one. (An earlier revision of this PR listed the scale as unproven while the code said `kind: "percent"` — having it both ways. The label keeps it and the prose now defends it.) What nobody has done is watch the value **move**, so a reading frozen at 100 is not yet distinguishable from a healthy one.
- **8002 and 8004 stay undecoded.** What would settle them: the vendor app's own display settings screen, one control at a time, params diffed after each.
- **The screen, the volume and the assistant are not modelled** because no captured param carries them. That points at a channel nobody has looked at yet rather than at ids left unnamed.
- **A display's charge reads through `dev.display?.()`, not `dev.battery?.()`.** Same meaning, different wires — 1101 in the security space, 8001 here — and one capability reading both would be a claim that the ecosystems share a param space, which is the door this PR closes. A deliberate trade, not an oversight.

## Review

Two passes from @lenoxys, eleven threads, and the PR is smaller and more honest for both.

**First pass — the capability had grown past its defect.** The defect is the line and the namespace, about twenty lines; the rest was a capability whose three string reads did not earn a typed getter. `modelName`/`modelCode` restated `info`; `softwareVersion` advertised a `guessed` id in the typed surface, which a hard rule names outright — and its own doc saying "prefer `info.firmwareVersion`" was the tell. Also caught: **the count disagreed with itself four times**, because battery arrived in a later commit and the surrounding sentences did not move with it, including the module `description` and the `DeviceActionMap` entry. A fifth place was worse — the guide's opening snippet still showed the three strings and no battery, because that commit's guide edit asserted after its replace and never wrote.

**Second pass — two labels and one doc claimed more than the capture supports.**

- `provenance: "mega"` on `battery` rated the wrong axis. The tier means confirmed against the vendor's data-point list; this name came from a person, which is `verified`. The dictionary doc had also quietly redefined `mega` as "the device sent it" — whether an id is real, not whether its name is trusted — so one word carried two meanings and the weakest claim got the strongest label.
- `kind: "percent"` in code while the body deferred "the scale is not proven". Resolved by defending the claim rather than hedging it, reasoning above.
- **`Codec`'s JSDoc still said the opposite of this diff** — display's namespace and line "grouped into `security` by maintainer decision", and "no capability module targets it yet". Four copies of that paragraph existed; three were rewritten and this was the one a reader reaches first.
- Four JSDocs narrated the change rather than stating the fact ("did not always", "used to be read from", "would have re-opened the door"). A JSDoc reads as present tense, and four sites pinned to a prior state are four things to edit next time — which is exactly how the `Codec` paragraph went stale.
- 8003 back to a member as `unexposed`, correcting the first pass's call.


src/model/capabilities/display.ts
```diff
@@ -0,0 +1,143 @@
...
+  /**
+   * The model's retail name as the DEVICE reports it.
+   *
+   * Worth having beside `info.name` rather than folded into it, because the two answer different
+   * questions: `info` is sourced from the cloud record and the curated registry, this is the device's
+   * own answer. Where they agree — they did on the capture — that agreement is the evidence the id means
+   * what it is named. Where a future unit disagrees, the disagreement is the interesting part and would
+   * be lost if this were quietly merged into identity metadata.
+   */
+  modelName: {
```
lenoxys: `modelName` and `modelCode` answer questions a caller already has answers for. The registry row and the cloud record carry both, `info` publishes them, and the PR's own evidence for these two ids is that the values matched what we already knew.

Keep them in the dictionary, where they name the id and carry that agreement, and `describe()` still reports them. Drop the members: two typed getters, two constants and two paragraphs defending their place beside `info` buy a caller nothing.

The "where a future unit disagrees, the disagreement is the interesting part" argument is real, and it is a diagnostic need. A raw param that decodes to a name covers it.

src/model/capabilities/display.ts
```diff
@@ -0,0 +1,143 @@
...
+  },
+  /**
+   * Something version-shaped, named for what it looks like.
+   *
+   * The weakest claim in this module and labelled as such: `provenance: "guessed"`. One dotted value on
+   * one device is a reason to expect a version and not a mapping anyone confirmed, so a consumer showing
+   * this as a firmware version is showing an inference. `info.firmwareVersion` remains the field to
+   * trust when the cloud record carries one — this display's record did not.
+   */
+  softwareVersion: {
```
lenoxys: `provenance: "guessed"` on a typed getter is the one thing the rule names outright: do not advertise a guess in the typed surface. A caller reading `softwareVersion` off a bound object has no way to see the label, and the member doc telling them to prefer `info.firmwareVersion` is the tell.

Dictionary entry yes, member no. That leaves `display` with `battery`, which is the only read here that tells a caller something it could not get another way, and the module stops needing a paragraph to justify itself.

src/model/capabilities/display.ts
```diff
@@ -0,0 +1,143 @@
+import type { CapabilityModule } from "./types.js";
+import { propertiesOf, type Members, type Surface } from "./members.js";
+
+/**
+ * `display` — what a eufy Smart Display (T87Ax) actually reports.
+ *
+ * A small capability, and its shape IS the finding. The Smart Display shares a cloud
+ * account with the security line and nothing else: the one captured unit (a T87A0, 2026-09-04) connects
+ * over secure MQTT with no `p2p_did`, so it speaks no P2P at all, and it reported exactly six params in
+ * an id range (8001-8006) that no other line uses. Four of those six are named here; two are not — see
```
lenoxys: The count disagrees with itself four times. Here it is four named, at `DISPLAY_PARAM` it is "only the mapped three", at `detection` it is "the three ids", and `param-namespace.ts` says "three of them mapped".

Battery arrived last and three sentences did not move with it. Worse, the module `description` and the `DeviceActionMap` entry both list model code, retail name and version and leave battery out, and `description` is what a caller reads to decide what to offer.

src/model/param-dictionary.ts
```diff
@@ -1710,3 +1710,63 @@ export const CLEAN_PARAMS: Record<number, ParamDef> = {
...
+ * 8001 was the third of those until the maintainer identified it as the BATTERY. Worth recording how
+ * that went, because the value alone could not have said so: `"100"` fits a percentage of brightness,
+ * volume or charge equally well, and picking one would have been a coin toss shown to users as a fact.
+ * It took someone who knows the device; it did not take another capture.
+ */
+export const DISPLAY_PARAMS: Record<number, ParamDef> = {
+  8001: {
+    name: "battery",
+    type: "number",
+    // Identified by the maintainer, not inferred from the capture — `"100"` on a mains-powered screen is
```
lenoxys: Body comments inside the entries. Two exist in the other seventeen hundred lines of this file, and the rule puts rationale in a JSDoc above the declaration.

They are also the fourth telling of the same story. The 8001 identification is now in the table doc, in this comment, in `DISPLAY_PARAM`, in the member JSDoc and in the guide. One home.

src/model/capabilities/index.ts
```diff
@@ -1032,6 +1037,7 @@ export { RTSP_MEMBERS } from "./rtsp.js";
 export { SIREN_MEMBERS } from "./siren.js";
 export { SMART_LIGHT_MEMBERS } from "./smart-light.js";
 export { SMOKE_MEMBERS } from "./smoke.js";
+export { DISPLAY_MEMBERS, DISPLAY_PARAM, type DisplayActions } from "./display.js";
```
lenoxys: `DISPLAY_PARAM` is published, where no other capability's wire ids are. Nothing outside the module reads it, and a published id table is permanent. Drop it from the barrel.

lenoxys (commented):
Ponytail pass. The defect is real and the fix for it is about twenty lines: a line, a namespace, a dictionary. The other four hundred are the capability, and three of its four published reads either restate what `info` already answers or ship a labelled guess as a typed getter. Cut those and the module is one honest member, the same defect stays fixed, and the evidence stays where evidence belongs.

src/model/capabilities/display.ts
martijnpoppen: Agreed, and dropped. Both are dictionary-only now.

Your reading of the evidence is the one that settles it: what those two ids rest on is that their values matched what the registry and the cloud record already carried, and `info` publishes exactly that — so a typed getter beside it offered a caller a second spelling of what it had just read.

I checked the premise before cutting, because the diagnostic argument only survives if the value really is still reachable: `applyParams` names a reported param from `paramDef` when no `PropertySpec` claims it, so `getProperties()` returns `modelName` and `modelCode` by name and `describe()` carries them in `properties`. A future unit that disagrees with the registry still shows it. The guide now documents that route explicitly rather than leaving it to be discovered.

---

src/model/capabilities/display.ts
martijnpoppen: You are right, and the tell you point at is the decisive part: a member whose own doc says "prefer `info.firmwareVersion`" is a member arguing against its own existence. Dropped to a dictionary entry.

I did weigh one counter-argument and it does not hold. The downstream consumer I was building for keeps this as its only identity row *because* this device's cloud record carries no `main_sw_version`, so `info.firmwareVersion` is empty on it — which felt like a reason the getter earned its place. But that just moves the guess one layer out: a row labelled as a version, shown to an end user, sourced from an id named for its shape. The rule reads "don't advertise a guess in the typed surface", and a consumer surface is the typed surface's consequence. The consumer loses the row, which is the correct outcome.

`display` is now `battery` alone.

---

src/model/capabilities/display.ts
martijnpoppen: Caught fairly — and the fifth place was worse than the four you found. The guide's opening snippet still showed the three strings and no battery at all, because that commit's guide edit asserted after the replace and never wrote; nothing failed, so it shipped silently.

Cause was exactly as you read it: battery arrived in a later commit and the surrounding sentences did not move with it. The `description` and `DeviceActionMap` omission is the one that actually mattered — `description` is what a caller reads to decide what to offer, and it advertised three reads the device would not hand over a getter for.

With the members cut there is one count left to be wrong about, and it now says `battery` in all of: the module doc, `description`, the `DeviceActionMap` entry, `DISPLAY_PARAM`, the `detection` note, `param-namespace.ts` and the guide.

---

src/model/param-dictionary.ts
martijnpoppen: Both fixed. The entries are bare data now — name, type, provenance — and the rationale sits in the table's JSDoc where the other three dictionaries put theirs.

The repetition was worse than untidy: five tellings of the 8001 story meant five things to keep in step, and the count inconsistency you found in the next thread is what that costs. Told once, in the table doc.

---

src/model/capabilities/index.ts
martijnpoppen: Out of the barrel. `DISPLAY_MEMBERS` and `DisplayActions` still export, matching every other capability; the id table does not, also matching every other capability.

It is down to a single id now that the other three members are gone, so there was even less to publish than when you flagged it.

---

src/model/types.ts
```diff
@@ -74,6 +74,9 @@ export type Capability =
   // eufy_life smart lighting (own secure-MQTT DP namespace) — the T8L0x line's on/off/brightness/
   // effect control. Distinct from the camera-floodlight `light` capability above.
   | "smart_light"
+  // --- Smart Display line (own 8001-8006 param namespace, secure MQTT, never P2P) ---
+  // What a T87Ax Smart Display reports about itself. Read-only: no display write is captured.
```
lenoxys, line 78: The `Codec` doc forty lines above this still says the display's param namespace and product line are "grouped into `security` by maintainer decision, not wire evidence" and that "no capability module targets it yet". Both are false as of this diff.

The same paragraph existed in four places and three were rewritten. This is the copy that got missed, and it is the one a reader hits first, since `Codec` is where the line is declared.

src/model/param-namespace.ts
```diff
@@ -43,17 +46,15 @@ export function paramDef(ns: ParamNamespace, paramType: number): ParamDef | unde
...
- * (8001-8006) don't presently exist in `SECURITY_PARAMS`. Grouping it here means (a) a future security
- * param assigned in that id range would silently misdecode against this device, and (b) any security
- * capability whose `modelHints` regex matches this device's reported name/model text becomes
- * inference-attachable — the exact `T8L20`-as-"Outdoor Spotlights" collision this partition otherwise
- * guards against, now possible for `display` too. Measured, not hypothetical: with an adversarial name
- * (see `line-partition.spec.ts`'s `POISONED` case) six security capabilities currently attach this way;
- * the real device's actual reported name doesn't trigger any of them today (see `model.spec.ts`'s
- * display test), so this isn't a live problem, only an open door — flagged here so a future reader
- * doesn't mistake either the grouping or its current quiet outcome for evidence that it's safe.
+ * `display` (the T87Ax Smart Display line) reads its OWN dictionary, and did not always. It used to be
```
lenoxys: "reads its OWN dictionary, and did not always" narrates the change rather than stating the fact. Same in `ProductLine` ("It was grouped into `security` while it had no capability to partition"), in `DISPLAY_PARAMS` ("where this line's ids used to be read from", "8001 was a third until"), and in `battery` ("would have re-opened the door").

The rule is explicit that a JSDoc does not narrate its own history, and the text you replaced was present tense. Four sites now need editing the next time this moves. The commit message is where the before and after belongs.

src/model/capabilities/display.ts
```diff
@@ -0,0 +1,79 @@
...
+   * **On this capability rather than on `battery`, and that is the line partition doing its job.** The
+   * security-line `battery` capability reads param 1101, and a Smart Display's charge is 8001 in its own
+   * id space — two wires that happen to mean the same thing. Putting 8001 on the security module would
+   * have re-opened the door this line was split to close, so a consumer reads a display's charge through
+   * `dev.display()` and a camera's through `dev.battery()`.
+   *
+   * `percent` follows the line's own convention and one consistent reading, not a proven scale: every
+   * eufy battery this SDK models is 0-100, and the capture said `"100"`.
+   */
+  battery: {
```
lenoxys, line 56: `provenance: "mega"` is defined in `types.ts` as a name confirmed against the cloud data-point list. This name came from a person, and the new dictionary doc redefines the label as "the device sent it", which is a different axis: whether the id is real, not whether the name is trusted. One label, two meanings, and the weakest of the three claims gets the strongest word.

`kind: "percent"` has the same problem two lines down, on a scale the doc itself calls unproven. Either the label gives or the getter does.

src/model/param-dictionary.ts
```diff
@@ -1710,3 +1710,50 @@ export const CLEAN_PARAMS: Record<number, ParamDef> = {
...
   },
 };
+
+/**
+ * eufy Smart Display (T87Ax) param space — ids 8001-8006.
+ *
+ * Its own table rather than a corner of {@link SECURITY_PARAMS}, which is where this line's ids used to
+ * be read from by virtue of the codec's namespace grouping. Nothing in the 8000s was ever assigned a
+ * security meaning, so that grouping decoded nothing today — but it left the door open for a future
+ * security param in this range to be read off a Smart Display as something it is not.
```
lenoxys: Correcting my earlier comment. Dictionary-only is not the shape the repo has for this. The capability README's third rule is `unexposed: true`: in the schema, reachable through `getProperty`, no typed getter, with precedent in `motion`'s sensitivities, `smart-light` and `battery`'s 1103.

As dictionary-only, 8003 loses its `PropertySpec`, so the schema no longer carries its type or its `guessed` label. Same fact answered two ways depending on which side a caller asks.

src/model/capabilities/__tests__/display.spec.ts
```diff
@@ -0,0 +1,97 @@
+import { DISPLAY, DISPLAY_MEMBERS, DISPLAY_PARAM } from "../display.js";
```
lenoxys, line 1: The namespace switch is asserted nowhere. This spec checks `DISPLAY_PARAMS[id]` is defined, and `model.spec.ts` feeds all six real params through `resolveDevice` and asserts nothing about the properties that come out.

Revert `NAMESPACE_BY_CODEC.display` to `security` and the suite still passes, which means the change this PR is named for has no test. One assertion closes it: `getProperties()` names battery and the three strings, and 8002 and 8004 fall through as raw ids.

While there: the PR body says all four "arrive on `getProperties()` and in `describe()`'s `properties`". `describe()` has no `properties` key, that is `toJSON()`, and the dictionary-only three never reach `details` at all.

docs/smart-display.md
```diff
@@ -0,0 +1,101 @@
+# Smart Display
```
lenoxys, line 1: The guide states the scale as fact (`// 0-100 — the screen's charge`) while the PR lists "the battery's SCALE is not proven" as deferred. The callout hedges `softwareVersion` and not this, which is the one a consumer will render as a gauge.

Smaller: `{@link DISPLAY_PARAMS}` in `display.ts` points at a symbol that module does not import, so it dangles on the reference site, and `DISPLAY_PARAM` is now a one-entry const whose only use is the member ten lines below. Inline 8001.

lenoxys (commented):
Recheck after your two commits, this time with the standards and spec passes run properly rather than by eye. All five cuts landed and the module is honest now.

What the two passes turned up is heavier than what I sent first: one shipped doc still states the opposite of what this PR does, four JSDocs narrate the change rather than the fact, the trust label on 8001 says more than the evidence does, and the namespace switch the PR exists for is asserted nowhere. Also correcting myself on 8003.

src/model/types.ts
martijnpoppen: Fixed, and your diagnosis of how it survived is exactly right: four copies, three rewritten, and the one on `Codec` is where the line is declared, so it is the copy a reader reaches first and the one most likely to be believed.

It now states what is true — own param namespace, own product line, `display` capability targets it — and says the screen/audio/assistant params are *absent* rather than *deferred*, since nothing captured carries them.

Worth noting what the near-miss was: a stale doc that contradicts the diff is worse than no doc, because it reads as a deliberate statement about the current design. `guard:docrefs` checks that referenced symbols exist, not that prose is still true, and I do not have a way to check the latter automatically.

---

src/model/param-namespace.ts
martijnpoppen: All four rewritten to state the fact. The reason each one is now a *reason* rather than a *change*: the namespace exists because nothing in the 8000s carries a security meaning; the line exists because a NAME-regex capability is otherwise attachable; `battery` is on `display` because one capability reading both 1101 and 8001 would claim the ecosystems share a param space.

Your "four sites to edit the next time this moves" is the part that made it worth doing properly rather than trimming the worst offender. Each of those sentences was pinned to a specific prior state, so any future change to the partition would have had to update all four or leave a contradiction — which is precisely what happened to the `Codec` paragraph in the thread above, on this very PR.

The before-and-after is in the commit message, where it can go stale without misleading anyone.

---

src/model/capabilities/display.ts
martijnpoppen: You are right about the label, and the "one label, two meanings" reading is the sharper half of it. `provenance` rates how far a NAME is trusted; for a name, "confirmed against the live mega API" means the cloud data-point list. 8001's name came from a person. That is our own observation, so it is `verified` — and the dictionary doc no longer redefines `mega` as "the device sent it", which was rating a different axis (whether the id is real) under the same word. It now says which axis it rates, and that every id in the table is there *because* the device sent it, which is the entry price rather than the label.

On `kind: "percent"` I went the other way from what you may have expected — the label keeps it and the doc stops hedging — because the hedge was the part that was wrong. I was claiming `percent` in code and "not proven" in prose, which is having it both ways, and one of the two had to give.

The evidence is better than the convention I was citing: a full charge reads `255` on a 0-255 scale and `1000` on a 0-1000 one, so `"100"` on a charged unit is positive evidence for a percentage, not merely consistent with one. Stated that way in the member doc and the guide.

What is genuinely unproven is narrower than "the scale", and now says so: nobody has watched the value **move**, so a reading frozen at 100 is not yet distinguishable from a healthy one. If you would still rather the getter gave instead, say so and I will drop `kind`/`unit` — but I would rather ship the claim I can defend than a number a consumer cannot render.

---

src/model/param-dictionary.ts
martijnpoppen: Taken, and thank you for correcting it rather than letting the earlier call stand — "same fact answered two ways depending on which side a caller asks" is the defect, and it was mine to see and I did not.

8003 is now a member with `unexposed: true`, `provenance: "guessed"`, no typed getter. I checked the precedent before writing it, and it is stronger than one case: `cameraInfo` (1103), `batteryPowerStats` (3100), the keypad's battery pair and `smart-light`'s all carry **both** a dictionary entry and an `unexposed` member — so keeping the dictionary row alongside is the convention, not duplication.

The concrete thing it buys, which dictionary-only lost: `DISPLAY.properties` now carries 8003 with `type: "string"` and `provenance: "guessed"`, so the schema and `getProperty` agree about the label. There is a spec assertion on exactly that, because it is the property the shape exists to preserve.

---

src/model/capabilities/__tests__/display.spec.ts
martijnpoppen: The substance is right and I have fixed it. One correction to the mechanics, because I tested it rather than taking it on faith: reverting `NAMESPACE_BY_CODEC.display` to `security` does **not** leave the suite green — `display.spec.ts`'s "reads its own id space" case fails on its `namespaceForCodec("display")` assertion. I ran that revert to check.

But that only makes your point in a more embarrassing way: the single assertion guarding the change was the table agreeing with itself, and it failed with `expected 'security' to be 'display'` — which tells a future reader that a constant changed, not that anything broke. The consequence was untested, exactly as you said.

So the case now runs the captured record through `Device.fromRecord` and asserts what comes out of `getProperties()`: `modelName` and `modelCode` named (neither has a capability spec, so the dictionary is the only thing that can name them, and only the `display` dictionary holds them), `battery` coerced to the **number** `100` rather than the string, `softwareVersion` present, and 8002/8004 arriving as `unknown_8002` / `unknown_8004`. The consequence asserts first and the mapping constant last, deliberately — so a revert now fails with `expected undefined to be 'Smart Display E10'`. I re-ran the revert to confirm that is the message you get.

On the PR body: you are right, and it was wrong in both halves. `describe()` returns a `DeviceManifest` whose key is `details`; `properties` is `toJSON()`. Corrected there.

---

docs/smart-display.md
martijnpoppen: The guide/PR contradiction is fixed by making the code's claim the true one rather than by hedging the guide — reasoning in the `kind: "percent"` thread. `// 0-100 — the screen's charge` stays, and the paragraph below it now says why 0-100 is evidenced (255 and 1000 are what a full charge reads on the other two candidate scales) and what is still missing (nobody has watched it move). You are right that this is the one a consumer renders as a gauge, which is why it needed a decision rather than a caveat.

Both smaller ones fixed too. `{@link DISPLAY_PARAMS}` is gone — the module doc now refers to the display param dictionary in prose, since `display.ts` does not import the symbol and a dangling link on the reference site is worse than no link.

`DISPLAY_PARAM` I have kept, and it is no longer a one-entry const: 8003 coming back as an `unexposed` member (your thread above) gives it `BATTERY` and `SOFTWARE_VERSION`, which matches how every sibling capability names its wire ids. If you would still rather see them inlined I will, but the convention argument now points the other way.

---

max246 (approved):

src/model/capabilities/display.ts
lenoxys: Fixed. `verified` is the honest label, and taking the redefinition out of the table doc is the half that stops the drift.

The wider problem is not yours to carry, so I am filing it rather than holding this: `PropertySource` in `types.ts` and the header of `param-dictionary.ts` still order and gloss these tiers differently, and this PR had to add a third reading to reconcile them.

docs/smart-display.md
lenoxys: Fixed, and making the code claim the true thing rather than hedging the guide was the better direction.

lenoxys (approved):
Approving. Three rounds in, the substance is all answered: the display owns its line and its namespace, the reads are down to what the device evidences, 8003 sits where the README says an unevidenced meaning goes, and the namespace switch now fails a test when reverted, which was the gap that mattered most.

Checked on a trial merge with current beta-0.2.0 rather than on the branch alone: 188 files, 3539 tests, green.

Two things I am filing rather than holding this for. `PropertySource` and the `param-dictionary.ts` header still gloss the provenance tiers differently, which is what forced a third reading of them here, and that predates this PR. And `getterNames()` in the new spec filters the members table rather than binding through `bindMembers`, so it asserts a copy of the rule instead of the binder honouring it.

PR #166, opened 2026-09-13
build(docs): typecheck the guides' TypeScript snippets in CI

martijnpoppen:
## What changed

`scripts/ci/check-doc-snippets.mjs` compiles every ` ```ts ` block that ships against the library's real types, and `check:snippets` is now part of `npm run verify`. Three places carry snippets and all three are scanned — the guides under `docs/`, the `README` (the npm landing page), and the fences inside `src/` JSDoc (which generate the published reference). **89 snippets typecheck**; the 54 problems found on arrival are fixed, plus 15 the base branch brought in while this was open and 12 more in the README and `src/` JSDoc.

Closes @lenoxys' suggestion on #164 — *"worth a compile check over the guide's snippets generally"*.

## Why

Two snippets in `docs/vacuums.md` did not compile: one named `setLevel` where the member's `writeAs` is `setSuctionLevel`, the other read `scenes` as a property and called `.find` on the function. A reviewer found them, not CI. Nothing in the repo compiled the guides, so a rename in `src/` could silently make the documentation wrong — which is the failure mode a guide has, because it is believed.

## Evidence

n/a for the tooling. The doc fixes are grounded in the type system itself: each was a compiler error against this commit's `src/`, and the two headline ones are pinned as fixtures so the checker can be shown to catch them — at the right line, which took two attempts.

## Verified how

- [x] `npm run verify` is green locally — 188 files, 3524 tests, `check:snippets` reporting 89 snippets
- [x] Ran against a real device — n/a, this touches no wire
- [x] Added or updated specs — `src/__tests__/check-doc-snippets.spec.ts`, running the checker against fixture guides and asserting the exact lines it reports
- [x] Docs updated — fifteen guides, the README and four `src/` files corrected; `CONTRIBUTING` documents the gate
- [x] Re-verified against the **merged** tree, not just this branch — see below

### How a fragment is made compilable

A guide snippet says `await clean?.startScene?.(first.id)` without ever declaring `clean`, because the prose declared it. Three moves make that checkable without rewriting a single snippet for the checker's benefit:

- **A prelude of ambient declarations** for the shared vocabulary (`dev`, `eufy`, `cam`, `light`, `clean`, `s`, …), type-only and namespace-imported so it binds exactly one name and cannot collide with a snippet's own imports.
- **The body wrapped in a function**, which is what makes the prelude safe: a snippet declaring its own `const dev = …` shadows the ambient one, where a prelude in the same scope would be a duplicate identifier. It also gives `await` somewhere to live. Imports are hoisted back out, since they cannot sit in a function — and several snippets are nothing but an import, which is itself a claim worth checking.
- **`…` treated as elision.** `eufy.on("motion", (e) => …)` is the clearest way to write "then do something with it"; rewriting the guides to `(e) => {}` would make them worse to read in order to make them checkable. Every `…` becomes an expression of type `never` — including inside strings and comments, where the substitution provably cannot change an answer (`"…"` stays a string literal of the same type).

Typing the prelude correctly mattered as much as having one: `cam` is the camera capability surface, not a `Device`, and typing it as a Device made every `cam.live()` in the media guide look like a guide error when the guide was right.

### What the 54 actually were

**A real API error in three places.** `region: "eu"` is not a `RegionShard` — the type is `"eu-pr" | "us-pr"` — so anyone copying the logging setup out of `getting-started` or `troubleshooting` got a type error on their first line. Nothing about the prose looked wrong. This is the class the checker exists for.

**One systematic mistake, seventeen times.** Every capability accessor on `Device` is optional, so a call site needs `?.` twice — once for "does this device have it", once for "is it bound yet". `devices.md` writes it that way and says so; `vacuums.md`, `events.md` and `capability-manifest.md` wrote `dev.vacuumClean()?.activity`, which throws on any device without the capability. The optional media methods had the same hole: `snapshotStored?.()` and `snapshotLive()` sat one line apart in two guides.

Also real: `suctionLevelName(level)` was handed `number | undefined`, and the sensitivity-slider snippet read `m.sensitivityStep()` off a possibly-absent capability — narrowed through `!m`, which reads better as prose too, since "not drivable" does cover a device with no motion capability at all.

**The rest was the harness being too strict about whose code it was checking.** Several guides illustrate wiring the SDK into something the reader already has — an event bus, an audio encoder, a UI slider, a function registering an accessory. Those names have no type this library could supply, and inventing one for each would put fiction in the checker. So a snippet names them with a `typecheck: host` marker, declaring them `any` — which types nothing about them and leaves every SDK call around them checked exactly as before.

### Then the base moved, and the gate would have landed red

#164 merged into `beta-0.2.0` while this was open, bringing a whole new Controls section into `docs/vacuums.md` — the very file this PR's gate compiles. GitHub reported this PR mergeable, and it is; its CI was green. But the merged **content** had never been compiled together, and it had **15 problems**. Found by merging the base in locally and running the gate, rather than by trusting a green tick from before the merge.

Two causes, both ordinary: the accessor sweep happened on this branch, so #164's new snippets never got it; and `clean` is declared in one fence and used in two later ones, which is correct prose and impossible for a checker that compiles each fence alone (it joins the prelude beside `cam` and `light`).

The lesson is the reusable part: **a doc gate is only as green as the content it last saw, and a merge changes that content without touching either side's diff.**

### The second review found the gate lying about where errors were

Worth its own section because it is the most serious thing here, and it is mine.

**Every reported line was one too high, and two too high in a hoisted import.** `PRELUDE_LINES` counted the prelude's trailing newline and the `+ 2` beside it counted the same line again. This PR's own description called the markdown coordinate "the part that makes the output usable", and the fixture spec asserted `guide\.md:\d+:\d+`, which passes for any integer — so the guard was loose in exactly the place it claimed to bite. A wrong line in the right file is worse than no coordinate: it sends a reader to innocent code.

Fixed by removing the arithmetic rather than correcting it. Every generated line now carries the snippet line index it came from, so a tsc coordinate resolves by lookup. The spec asserts `:7` and `:9` outright.

That change paid for itself twice. A generated line with no snippet index behind it is a line the *checker* wrote, so those errors are now reported **once**, against the checker, instead of once per snippet: renaming a prelude name went from 80 problems (79 of them echoes at column 330 of a collapsed line) to 3, with the real guide error first.

**`host` was hiding one of our own types.** `docs/connectivity.md` declared `s` a host name, but `s` is `eufy.deviceState(sn)` — an SDK type. `any` is an opt-out from the guard, so the one member that guide teaches was the one thing unchecked: renaming `lastSeenMs` across `src/` reported nothing. `s` is in the prelude with its real type now, and the rule — a `host` name is never ours — is written into the marker's doc and `CONTRIBUTING`, because this failure wears the look of a deliberate choice.

**And the gate did not cover the copies that outrank a guide.** `region: "eu"` was this PR's best find and was still live in `eufy-mega.ts`'s JSDoc, which generates the published reference. Scanning the README and `src/` JSDoc (the same extractor, plus a `strip` for JSDoc's left margin) found **12 more problems** in eight places, including the seventeen-times accessor bug in `getDevice`'s own example and two `new EufyMega({ logger })` snippets missing their credentials.

### The guard's own guard

`--docs <dir>` points the checker at a fixture tree, and a spec runs it against a guide carrying the two bugs that motivated it — a setter that does not exist, and a method read as a property — plus one that compiles, exercising the `skip` and `host` markers. It asserts the failures are reported against the **markdown** at the **exact lines**, which is what the first version got wrong.

## Unverified or deferred

- **Two snippets are skipped**, each with a reason in its marker: a sketch of a returned shape, and one importing `tslog` — a package the reader has and this repo does not depend on. The reason is no longer *enforced*: three call sites all carry one, and policing them cost a capture group, a refusal branch, a fixture guide and a spec case. A reasonless skip would appear in a diff someone reads.
- **`host` names are `any`.** Deliberate — they are the reader's code, and a precise type here would be invented. A mistake in the *host half* of such a snippet is not caught; the SDK half still is.
- **The checker types, it does not run.** Whether a snippet works needs a device and belongs in `examples/`.
- **Bare indented code in JSDoc is still outside the gate.** `eufy-mega.ts`'s module header shows its example as indented lines, not a ```ts fence — which is why the `region: "eu"` there survived until it was read by eye. Fencing those would bring them in, and would change how the module headers render; not done here.
- **It does not check the merged tree by itself.** This PR found that case by hand. A CI job that merges the base before running `verify` would catch it generally, and is not in this diff.
- **The spec lives in `src/__tests__` while its fixtures live in `scripts/ci/`.** `vitest.config.ts` has `include: ["src/**/*.spec.ts"]`, so a spec under `scripts/` would not run. Documented in the spec rather than left looking accidental.
- `lib: ["DOM"]` came out of the generated tsconfig along the way — it re-typed global `fetch` and made `src/transport/push/fcm.ts` fail on a `BodyInit` overload the library's own config accepts. Flagging it because it is the one change here that touches how `src/` is compiled, even though only in this tool's throwaway project. `typeRoots` was added for the opposite reason: with the output now in `os.tmpdir()`, type resolution no longer walks up into this repo's `node_modules`.

## Review

Two passes from @lenoxys, eleven threads, all taken. The first pass's four (a 55-line elision lexer that only changed an answer in the case it got backwards, generating into `docs/`, enforcing the skip reason, fixtures in `src/`) I had trimmed around rather than acted on, which was wrong; they are done now. The second pass found the off-by-one, the hosted SDK type, the error explosion, the unchecked README and JSDoc, `cam.live!()` five times in the guide people read for it, a duplicated path expression whose two spellings disagreed (one matching two literal backslashes, so it replaced nothing on the platform it existed for), and a `—` escape in the package description that — contrary to the review's guess and my own assumption — came from a JSON round-trip in **this** PR.

`check:docs` is renamed `check:snippets`: one character from `guard:docs`, which guards the reference's publication and is a different thing entirely.


scripts/ci/check-doc-snippets.mjs
```diff
@@ -0,0 +1,465 @@
...
+ * `(e) => {}` to satisfy a checker would make them worse to read in order to make them checkable, so
+ * the checker gives way instead: a bare `…` becomes an expression of type `never`, which is assignable
+ * wherever a value is expected and never masks a type error of its own.
+ *
+ * Only OUTSIDE strings, template literals and comments — `description: "…"` is already valid and must
+ * stay a string, and an ellipsis in prose inside a comment is not code at all. Hence a scan rather
+ * than a replace: the state machine is small, and getting this wrong would either break valid snippets
+ * or silently edit their meaning.
+ */
+function expandElisions(code) {
```
lenoxys: This is a hand-rolled lexer guarding a case that cannot change the answer. Inside a string, `"…"` becomes `"(undefined as never)"`, which is still a string literal and typechecks identically. Inside a comment it is still a comment. The substitution is invisible to tsc in both places, and the temp copy is thrown away.

Where the carve-out does change an answer is the one it gets backwards: a bare `…` inside `${}` needs expanding, and the scanner treats a backtick as a quote and declines, so tsc rejects a snippet that was fine.

`code.replaceAll("\u2026", "(undefined as never)")`. Fifty-five lines out, and one fewer thing to be subtly wrong.

scripts/ci/check-doc-snippets.mjs
```diff
@@ -0,0 +1,465 @@
...
+ * Which tree to scan, so the checker can be pointed at a FIXTURE and thereby tested itself.
+ *
+ * A guard nobody checks is a guard nobody can trust: this one exists because two broken snippets shipped,
+ * and it would be the same mistake to take on faith that it catches them. `--docs <dir>` lets a spec hand
+ * it a guide with a known bug and assert on what comes back. Defaults to the real guides.
+ */
+const docsArg = process.argv.indexOf("--docs");
+const DOCS = docsArg === -1 ? join(ROOT, "docs") : resolve(process.argv[docsArg + 1]);
+// Generated files live beside the tree being scanned, so a fixture run cannot disturb the real one.
+const OUT = join(DOCS, ".snippets");
```
lenoxys: Generating into `docs/` is what forces the two `.gitignore` lines and the second fixture path. `os.tmpdir()` costs nothing, keeps CI out of the published tree, and `--keep` still has somewhere to point.

scripts/ci/check-doc-snippets.mjs
```diff
@@ -0,0 +1,465 @@
...
+    if (entry.name.startsWith(".") || entry.name === "node_modules") continue;
+    const path = join(dir, entry.name);
+    if (entry.isDirectory()) out.push(...markdownFiles(path));
+    else if (entry.name.endsWith(".md")) out.push(path);
+  }
+  return out;
+}
+
+/** `<!-- typecheck: skip — why -->`, which must carry a reason. */
+const SKIP = /^<!--\s*typecheck:\s*skip\s*(?:[—-]\s*(?<reason>.+?))?\s*-->$/;
```
lenoxys: The reason is mandatory to police three call sites, all of which fit on one screen and all of which already carry one. That rule costs the optional group here, the refusal path, the `unreasoned` fixture guide and a spec case.

Keep the marker, drop the enforcement. If a reasonless skip ever appears, it appears in a diff someone reads.

scripts/ci/__fixtures__/doc-snippets/broken/guide.md
```diff
@@ -0,0 +1,11 @@
+# A guide with the two bugs this checker was built for
```
lenoxys, line 1: New top-level directory in `src/` for markdown that is neither library code nor a spec. Fixtures in this repo live beside the spec that reads them, and this one belongs to a CI script, so `scripts/ci/` fits it better than the shipped source tree.

Also deletes the second `.gitignore` line once the temp dir moves out.

lenoxys (commented):
Right call, and it already earned itself: `region: "eu"` in the first snippet a reader copies is exactly the class of bug no reviewer catches twice. Ponytail pass on the size. Five hundred lines of tooling for seventy-eight snippets, and about a hundred come off without losing a single check.

scripts/ci/check-doc-snippets.mjs
```diff
@@ -0,0 +1,377 @@
...
+      }
+      const source = generated.find((g) => g.name === m.groups.name.split(/[/\\]/).pop());
+      if (!source) {
+        problems.push(line.trimEnd());
+        continue;
+      }
+      // A generated line sits after: the prelude, the hoisted declarations (which came from the body and
+      // so are counted back), the shadow block, and the wrapper's own opening line.
+      // prelude + the hoisted lines + the wrapper's opening line + the one-line shadow block.
+      const preamble = PRELUDE_LINES + source.hoistedCount + 2;
```
lenoxys: Every reported line number is one too high, and hoisted imports are two. Injected an error on `smart-lights.md:23` and got `:24`; the fixture's `setLevel` sits on line 7 and reports `:8`; `smart-lights.md:55` reports `:53`.

`PRELUDE_LINES` already counts the trailing blank line and the `+ 2` adds it again.

This is the claim the PR singles out as "the part that makes the output usable", and the fixture spec asserts `guide\.md:\d+:\d+`, which passes for any integer. The guard's own guard is loose exactly where it claims to bite. Assert the two known line numbers.

docs/connectivity.md
```diff
@@ -175,6 +175,8 @@ differs per device: a mains camera reports constantly, while a battery contact s
 days by design and be perfectly healthy. If the SDK picked one number it would be wrong for somebody —
 so you decide:
 
+<!-- typecheck: host s, myThresholdFor, deviceKind -->
```
lenoxys: `s` is not the reader's code, it is `eufy.deviceState(sn)`, an exported SDK type. Declaring it a host makes it `any`, so I renamed `lastSeenMs` across `src/` and this guide reported zero errors. The one member the snippet teaches is the one thing not checked.

`host` is sound for an event bus or a UI slider. Used on an SDK value it is an opt-out from the guard on our own surface, which is the failure mode the PR exists to close.

scripts/ci/check-doc-snippets.mjs
```diff
@@ -0,0 +1,377 @@
...
+
+/**
+ * Names the guides use as VALUES that a snippet may equally well import for itself.
+ *
+ * These cannot go in the prelude: `declare const EufyMega` at module scope collides with a snippet's own
+ * `import { EufyMega } from "@mega-yfue/eufy-sdk"`, which TypeScript calls a duplicate identifier. Emitted
+ * INSIDE the wrapper instead, where a snippet's import is shadowed rather than duplicated — same type, and
+ * an import of a name the barrel does not export still fails at module scope.
+ */
+const AMBIENT_VALUES = {
```
lenoxys: Rename anything the prelude names and the output becomes unusable: renaming `listLightEffects` produced 80 problems, the one real one plus 78 copies at column 330 of the collapsed shadow line, attributed to the first snippet line of fifteen guides.

That is the most likely trigger for this whole tool, so it is worth handling: collapse duplicates by message, or report prelude-level errors once against the prelude rather than per snippet.

Same file, two cheap cuts: `--keep` has no caller anywhere, and `AMBIENT_VALUES`'s keys are never read, only its values, so an array says it.

docs/live-media.md
```diff
@@ -52,8 +52,10 @@ Worth knowing:
 
 The direct escape hatch — raw frames as they arrive.
 
+<!-- typecheck: host consumeAudio -->
+
 ```ts
-const stream = await cam.live();
+const stream = await cam.live!();
```
lenoxys: `cam.live!()` is the idiom this PR calls the bug. The description says the systematic mistake was a missing `?.` on optional accessors, seventeen times, and `architecture.md` and `troubleshooting.md` got `cam?.live?.()` in this same diff. The media guide, which is the one people read for this, got the assertion that throws instead. Five of them here.

One idiom, and it is the one the other guides now use.

scripts/ci/check-doc-snippets.mjs
```diff
@@ -0,0 +1,377 @@
...
+ * Which tree to scan, so the checker can be pointed at a FIXTURE and thereby tested itself.
+ *
+ * A guard nobody checks is a guard nobody can trust: this one exists because two broken snippets shipped,
+ * and it would be the same mistake to take on faith that it catches them. `--docs <dir>` lets a spec hand
+ * it a guide with a known bug and assert on what comes back. Defaults to the real guides.
+ */
+const docsArg = process.argv.indexOf("--docs");
+const DOCS = docsArg === -1 ? join(ROOT, "docs") : resolve(process.argv[docsArg + 1]);
+// Generated files live beside the tree being scanned, so a fixture run cannot disturb the real one.
+const OUT = join(DOCS, ".snippets");
```
lenoxys: Still generating into `docs/`, which is what forces both `.gitignore` lines and the second fixture path. `os.tmpdir()`.

And `relative(OUT, src/index.ts).replace(...)` appears twice, with `/\\\\/g` in the prelude and `/\\/g` in the tsconfig. One fact, two spellings, one of them wrong. Hoist it.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@mega-yfue/eufy-sdk",
   "version": "0.1.1",
-  "description": "One typed TypeScript client for the Anker eufy v6 cloud — capability-driven devices, realtime events over P2P/MQTT/push, and live media",
+  "description": "One typed TypeScript client for the Anker eufy v6 cloud \u2014 capability-driven devices, realtime events over P2P/MQTT/push, and live media",
```
lenoxys: The published description's em dash became a `\u2014` escape. Unrelated to this PR and unmentioned in it.

Related to what belongs here rather than what does not: `CONTRIBUTING.md` documents the gate's chain and its individual pieces, and `check:docs` is in neither. The marker vocabulary is documented only in this script's header, where no guide author will look. `check:docs` also sits one character from the existing `guard:docs`, which guards something else entirely.

docs/troubleshooting.md
```diff
@@ -13,9 +13,9 @@ MQTT, P2P, push, live media), each line prefixed by subsystem (`[mega]` / `[smqt
 ```ts
 import { EufyMega, ConsoleLogger } from "@mega-yfue/eufy-sdk";
```
lenoxys, line 15: `region: "eu"` was the best find in this PR, and it is still live in `src/client/eufy-mega.ts`'s own JSDoc example, which generates the published reference. The checker scans `docs/*.md`, so the more authoritative copy of the same error survives.

`README.md`'s snippet is outside `docs/` too, and it is the npm landing page.

lenoxys (commented):
My first pass was against a stale head, before the trim commit. The elision scanner is already a two-line regex, so ignore that comment. `docs/.snippets`, the mandatory skip reason, `--keep` and the fixtures under `src/` are all still there.

Ran the two passes properly this time, including driving the checker against renames in `src/` rather than trusting the description. Three findings say the guard is weaker than it claims, and one guide now teaches the bug this PR says it fixed.

scripts/ci/check-doc-snippets.mjs
martijnpoppen: This is the finding of the review. You are right on all three numbers, and right that the fixture's `\d+:\d+` made the guard loose exactly where it claimed to bite — a wrong line in the right file is worse than no coordinate, because it sends a reader to innocent code that looks fine.

I fixed it by deleting the arithmetic rather than correcting it. Each generated line now carries the snippet line index it came from, and a tsc coordinate resolves by lookup:

```js
const index = source.mapTo[Number(m.groups.line) - 1];
```

`index === undefined` means the checker wrote that line, which also gave the scaffold-dedupe in your other thread for free. There is no preamble height to be wrong about any more, so the hoisted-import case stops being special.

The spec asserts the numbers outright now, as you asked:

```ts
expect(output).toMatch(/broken[/\\]guide\.md:7:\d+ .*'setLevel'/);
expect(output).toMatch(/broken[/\\]guide\.md:9:\d+ .*'find'/);
```

And I re-ran your `lastSeenMs` experiment against the real guides as an independent check: `docs/connectivity.md:181`, which is the line that snippet is on.

---

docs/connectivity.md
martijnpoppen: Correct, and the way you demonstrated it is the argument: you renamed `lastSeenMs` across `src/` and the guide teaching `lastSeenMs` reported nothing. A gate that reads green on a rename of the member its own example is about is worse than no gate, because it is *believed*.

`s` is in the prelude now with its real type:

```ts
declare const s: NonNullable<ReturnType<Sdk.EufyMega["deviceState"]>>;
```

The marker keeps `myThresholdFor` and `deviceKind`, which genuinely are the reader's. Re-running your rename now reports exactly one problem, at `connectivity.md:181`.

I also wrote the rule into `HOST_NAMES`'s own doc and into `CONTRIBUTING`, because your framing names why this one was hard to see: `host` is an opt-out from the guard, so using it on one of our types disables the check on our own surface *in the one place it looks deliberate*. It isn't a slip that shows up as a slip. If the shared vocabulary is missing a name, the fix is to add it to the prelude with its real type.

---

scripts/ci/check-doc-snippets.mjs
martijnpoppen: Handled, and your reason for it being worth handling is the right one — a rename of a prelude name is the most likely thing to trigger this tool, so the output it produces on that path is the output that matters most.

A scaffold error is now reported once, attributed to the checker rather than to a guide. The line map from the off-by-one thread gave it for free: a generated line with no snippet index behind it is a line the checker wrote, so it is by definition the same error in every snippet.

Measured with your own case — renaming `listLightEffects`:

```
docs/smart-lights.md:55:10 error TS2305: Module '…' has no exported member 'listLightEffects'.
docs/smart-lights.md:60:32 error TS7006: Parameter 'e' implicitly has an 'any' type.
scripts/ci/check-doc-snippets.mjs (the generated prelude, not a snippet): error TS2339: …
```

80 problems down to 3, and the real guide error is now first rather than buried. (`:55` is also the right line — it was reporting `:53`.)

Both cuts taken: `AMBIENT_VALUES` is an array, since only its values were ever read, and `--keep` is gone — no caller, and with the output in `os.tmpdir()` there was nothing left for it to point at.

---

docs/live-media.md
martijnpoppen: No defence for this one — the diff argued that a missing `?.` on an optional accessor is the systematic mistake, fixed it seventeen times, gave `architecture.md` and `troubleshooting.md` the safe idiom, and left the guide people actually read for live media on the assertion that throws. Five of them, as you counted, plus a sixth on `openReadable!`.

All six now use `cam?.live?.()`. Where the snippet goes on to use the handle several times I narrowed once instead of scattering `?.`:

```ts
const stream = await cam?.live?.();
if (!stream) return; // this device has no camera, or no live path bound yet
```

which reads better as prose too — nine `stream?.` lines would be noise, and "this device has no camera" is a real case a reader should see handled. Same for `talkback`. The one-shot calls just take `?.`, and the stored-snapshot example's annotation is honest about it now (`Buffer | undefined`).

---

docs/troubleshooting.md
martijnpoppen: Both are in the gate now, and you were right that this is the copy that matters more: the same error in the published reference outranks the same error in a guide.

The extractor already did the work — a fence is a fence — so it needed one function to un-prefix JSDoc's left margin (`stripJsdoc`) and the file list widened to the README and `src/**/*.ts`. Turning it on found **12 problems in eight places**, all real:

- `dev.camera()?.snapshotStored()` in `getDevice`'s example — the seventeen-times bug, in the reference.
- `new EufyMega({ logger })` in two `logger.ts` examples, missing `email`/`password`, so neither compiled as written.
- `cam.live()` in `contracts.ts` — the example on `live()`'s own declaration, where `live` is optional.
- `ask`, `log` and `write` — the reader's own names, in a place with no marker vocabulary; switched to the prelude's `promptUser`, `myLog` and `sink`, so the reference and the guides share one vocabulary.
- `PtzDirection` in the README, which is a real export and just needed the prelude to know it.

`region: "eu"` itself is fixed (`"eu-pr"`), though I should be precise about why the gate did not catch that one: it is in `eufy-mega.ts`'s module header as bare indented code, not a ```ts fence, so it is still outside the scan. Fencing those would bring them in, and I have not done it here — it changes how the module headers render, which is a separate call.

89 snippets, up from 78.

---

package.json
martijnpoppen: One correction: the `—` escape **is** this PR's. `git log -S` puts it in my own commit — almost certainly a JSON round-trip I did on `package.json`, which re-serialised the file with ASCII escaping. So thank you for catching it; it was not pre-existing as we both would have assumed.

Restored to the em dash, and edited as text this time rather than through a parse-and-dump, since that is what broke it.

The other two taken as well:

- Renamed `check:snippets`. `check:docs` one character from `guard:docs` was a genuine trap, given that the two guard unrelated things — snippet compilation versus the reference's publication.
- `CONTRIBUTING` now documents it: in the `verify` chain, in the individual-pieces list, and a short section on what it scans and the two markers, including that a `host` name must never be one of ours. You are right that the script's own header is where no guide author will ever look.

---

scripts/ci/check-doc-snippets.mjs
martijnpoppen: Taken — it is `code.replaceAll("…", "(undefined as never)")` now, and the 55 lines are gone.

I owe you an apology for the shape of the previous round: I replaced the character-scanning loop with a token regex and left the carve-out standing, which addressed the length and not the argument. Your argument was the right one and I did not engage with it.

It is right on both halves. The carve-out cannot change an answer where it fires — `"…"` becomes `"(undefined as never)"`, still a string literal of the same type, and an ellipsis in a comment is still a comment. And it *did* change an answer where it misfired, on the `${…}` case, by reading a backtick as a quote. A guard whose only observable effect is the case it gets wrong is not a guard.

The reasoning is now in the function's doc, so the next person to feel the urge to be clever about strings finds out why not.

---

scripts/ci/check-doc-snippets.mjs
martijnpoppen: Done — `mkdtempSync(join(tmpdir(), "eufy-doc-snippets-"))`. You raised this twice and were right twice; it should not have taken a second pass.

Both `.gitignore` lines are gone with it, and the second fixture path stopped needing to exist, exactly as you said it would.

One wrinkle worth recording, since it is the only cost: type resolution walks up from the tsconfig, so from `/tmp` it never reaches this repo's `node_modules` and `types: ["node"]` failed with `Cannot find type definition file for 'node'`. One `typeRoots` entry fixes it, and it is documented as being there only because the output moved out of the tree.

`--keep` is dropped rather than repointed, per your other thread — no caller, and nothing to point at now.

---

scripts/ci/check-doc-snippets.mjs
martijnpoppen: Taken. The enforcement is gone; the marker and the convention stay.

Your accounting was the persuasive part — a mandatory reason cost the optional capture group, the refusal branch, a whole fixture guide and a spec case, to police three call sites that all already carry one. That is a lot of machinery to defend a habit, and "if a reasonless skip ever appears, it appears in a diff someone reads" is true: three skips in the entire repo is not a population that needs automated policing.

`SKIP` is now `/^<!--\s*typecheck:\s*skip\b[^>]*-->$/` — anything after `skip` is for the human reading the diff. The `unreasoned` fixture and its case are deleted. `CONTRIBUTING` asks for the reason, which is the right altitude for it.

The orphaned-marker check stays, since that one catches a marker silently applying to the wrong snippet, which no diff reader would notice.

---

scripts/ci/__fixtures__/doc-snippets/broken/guide.md
martijnpoppen: Moved to `scripts/ci/__fixtures__/doc-snippets/`, beside the script that reads them, and out of the shipped source tree. The `unreasoned` one is deleted outright (skip-reason thread), so it is two fixtures now, not three.

One deviation from the letter of it: the **spec** stays at `src/__tests__/check-doc-snippets.spec.ts`, because `vitest.config.ts` has `include: ["src/**/*.spec.ts"]` — a spec under `scripts/` simply would not run. So the fixtures are beside the script and the spec is where vitest can find it, and the spec's doc says so rather than leaving the split looking accidental. Say the word if you would rather widen the vitest include instead.

And you were right about the knock-on: the second `.gitignore` line is gone, along with the first, now that nothing is generated into the tree at all.

---

scripts/ci/check-doc-snippets.mjs
martijnpoppen: Both done, and the second one is the better catch of the two.

`os.tmpdir()` — covered in the first-pass thread; it should not have needed asking twice.

The duplicated expression is hoisted to one `SDK_ENTRY`, and you are right that one of the two spellings was wrong. `/\\\\/g` in the prelude matches **two** literal backslashes, so it replaced nothing on the only platform it exists for; the tsconfig's `/\\/g` is the correct one. A Windows run would have had a prelude importing a path with backslashes in it while the tsconfig beside it resolved fine — the kind of divergence that only shows up on the machine nobody ran it on.

Its doc names that as the reason it is hoisted, rather than just "DRY":

> Hoisted because it is needed twice — the prelude's `import type` and the tsconfig's `paths` — and spelling it twice is how the two came to disagree.

`TYPE_ROOTS` is a second such constant, with the same treatment, needed only because the output moved to a temp dir.

---

scripts/ci/check-doc-snippets.mjs
```diff
@@ -0,0 +1,451 @@
...
+    body.push({ text, i });
+  });
+  return { hoisted, body };
+}
+
+/** Markdown files to scan, deepest-first so a nested guide is found too. */
+function markdownFiles(dir) {
+  const out = [];
+  for (const entry of readdirSync(dir, { withFileTypes: true })) {
+    if (entry.name.startsWith(".") || entry.name === "node_modules") continue;
```
max246, line 234: `markdownFiles` recurses all of `docs/` skipping only dot-dirs and `node_modules`, so it also walks the **gitignored, generated `docs/api/`** TypeDoc output — whose ```ts fences are type-signature fragments that don't compile. Reproduced on this branch with `docs/api/` present (it's generated locally and by the docs-site build): `node scripts/ci/check-doc-snippets.mjs` reports **2289 spurious problems** and errors out. README/CONTRIBUTING tell contributors to run `npm run verify` (now → `check:docs`) green before a PR, so anyone who has built the API reference hits thousands of false failures; CI only stays green because it never generates `docs/api` before verify. Fix: skip `docs/api/` here the same way dot-dirs are skipped (e.g. `entry.name === "api"` under `docs/`, or exclude the configured build-output dir).

scripts/ci/check-doc-snippets.mjs
```diff
@@ -0,0 +1,451 @@
...
+        .filter(Boolean);
+      continue;
+    }
+    if (line !== "```ts") {
+      // A blank line between the marker and its fence is fine; anything else means the marker was
+      // orphaned, which is worth reporting rather than silently applying to a later snippet.
+      if (line !== "" && skip) found.push({ orphanedSkip: skip });
+      if (line !== "") {
+        skip = undefined;
+        host = [];
```
max246: Asymmetric with the `skip` marker two lines up: an orphaned `skip` (prose between the marker and its fence) is reported as `orphanedSkip`, but an orphaned `host` marker is silently reset here with no diagnostic — which the comment just above says is exactly what shouldn't happen ("worth reporting rather than silently applying to a later snippet"). Consequence: a misplaced `<!-- typecheck: host bus -->` means the intended snippet is typechecked WITHOUT its host `any` declarations, so the reader's own names (bus, encoder, …) surface as false 'does not exist' errors and CI fails on correct documentation, with no hint the marker was orphaned. Report an orphaned `host` the same way as `skip`.

scripts/ci/check-doc-snippets.mjs
martijnpoppen: Confirmed and fixed. I reproduced it rather than taking the count on faith: one synthetic TypeDoc page — `getProperty(name): undefined | PropertyValue` in a fence under `docs/api/` — is already enough to fail the run, so a real reference reaching thousands is exactly right.

Your diagnosis of why it stayed hidden is the part worth keeping: CI never generates `docs/api` before `verify`, so the gate was green for everyone who had not built the docs site and broken for everyone who had — and README and CONTRIBUTING tell contributors to do precisely that. A gate that fails for people who followed the instructions is worse than one that does not exist.

I took your second suggestion rather than the first: the directory comes from `docs/typedoc.json`'s own `out`, with `"api"` as the fallback if the config is unreadable. Where the reference is generated is a fact that file owns, and this PR has already been bitten once by one fact spelled in two places — @lenoxys caught `SDK_ENTRY` written two ways with one of them wrong. Hardcoding `"api"` here would have been the same trade.

Skipped only at the scan ROOT, so a hand-written `docs/something/api/` is still checked; it is the generated tree at the top that is not ours to hold to this.

There is a spec for it now, confirmed to fail with the skip reverted.

---

scripts/ci/check-doc-snippets.mjs
martijnpoppen: Right, and the asymmetry is the whole of it — the comment two lines up states the rule and the very next line breaks it for the other marker. Both are reported the same way now.

Your reasoning about the consequence is what makes this worse than the `skip` case, and it is now in the code so the next reader gets it: a dropped `skip` fails loudly on the snippet it meant to exempt, but a dropped `host` leaves the snippet compiling **without** the reader's names, so it fails with `Cannot find name 'bus'` — which reads as a mistake in correct documentation and says nothing about a marker missing its fence. The fixture asserts the two lines arrive together, so they explain each other:

```
guide.md:1 typecheck:host marker is not above a ```ts fence
guide.md:6:1 error TS2304: Cannot find name 'bus'.
```

**Building that fixture turned up a third bug, mine.** I put the fixture in `os.tmpdir()` and the run reported `No inputs were found in config file` — zero snippets compiled. The generated filename came from `rel`, which is relative to the REPO, so a tree above it produced names starting `..`, and tsc's `*.ts` include treats a leading dot as a hidden file and matched none of them. So `--docs` — a documented interface this PR added, and the one the guard's own guard rides on — silently checked nothing whenever it was pointed outside the repo. It never showed because the spec's fixtures all live inside it.

The name is sanitised now (`rel` is what gets reported, and is unaffected), and both new specs run from `os.tmpdir()` deliberately, since that is the case the bug hid in. Each of the three fixes was confirmed to fail with only itself reverted.

`npm run verify` green: 188 files, 3526 tests, 89 snippets.

---

max246 (approved):

PR #167, opened 2026-09-13
fix(p2p): let a still give a station back, and state when a stream re-asserts its channel

lenoxys:
## Why

A station serves one camera at a time and refuses a second channel on a busy one. A still that keeps its pull attached therefore denies that station to every other camera behind it — and it was keeping it in two places it did not need to.

Measured on a real base fanning three cameras:

- a live request refused **370 ms after** the still it was waiting on had already collected everything it needed
- a live request refused **160 ms after** that still had been asked to yield, which it did **1.06 s later**

## What

**The abort reaches the capture, not only the wait for it.** `abortable` raced the promise, so an abandoned still rejected its caller while the capture went on collecting and held the channel to its own 20 s keyframe timeout. It rejects with the caller's own reason, because the abandonment is the caller's fact rather than a failure of the source.

**The consumer is released before the decode, not after.** A decode works on bytes already collected and needs no station, so a still was holding the channel for the length of an FFmpeg run past the point it had what it came for.

**`channel-silent` states a re-assert.** A station that switches to a sibling leaves a stream with no frames, no error and no stop; the recovery existed only as free text at debug level. `reasserted` is the repair, `declined` is a pull nothing is attached to leaving the station with the camera someone is watching. Its **absence** is evidence too — it is what ruled the station out of a frozen picture that turned out to belong to the caller's own output.

## What breaks

`LiveTrace` gains a phase, so a caller with an exhaustive table typed against the union fails to compile until it handles it. That is the compatibility claim working: the table is meant to fail rather than silently discard a phase.

## Verified

- `npm run verify`: **3502 tests / 183 files**
- three new specs, each failing without the change: an abandoned capture detaches at once, one already abandoned attaches nothing, and the station is free while a deliberately slow decode still runs
- **on a live fleet**, side-loaded as a packed tarball against the consuming host: seven consecutive sessions, every cession honoured in **2–7 ms** where it had taken 1.06 s, and **zero station refusals** across 166 retained records — the refusal had been reproducible on demand for the whole afternoon


src/transport/p2p/live-trace.ts
```diff
@@ -129,7 +129,19 @@ export type LiveTrace =
...
+   *
+   * A station that switches to a sibling leaves the stream it was serving with no frames, no error and no
+   * stop, so this silence is the only statement that it happened. `reasserted` re-issued the media start,
+   * which is the repair; `declined` left the channel alone because nothing is attached to this pull and
+   * taking the station back would take it from a camera someone is watching.
+   *
+   * Media still arriving means this never fires, so a picture that stopped advancing while this is silent
+   * stopped for a reason upstream of the station's attention.
+   */
+  | { phase: "channel-silent"; silentMs: number; outcome: "reasserted" | "declined" };
```
martijnpoppen, line 144: Non-blocking, but `silentMs` means something different here than it does one member up.

`path-stale` reports **measured** silence — `Date.now() - this.lastPongAt` (`p2p-session.ts:416`) — and traces once per connection behind `pathStaleTraced`. This one passes `stallMs`, the configured window, and the `declined` branch re-arms (`live-stream.ts:235-239`), so it re-emits every window for as long as nothing is attached. Same field name, same union, two readings: one is how long it has actually been quiet, the other is a constant from the options.

That matters most in the case you're adding it for. A stream nobody is watching produces an unbroken run of byte-identical records — `{phase: "channel-silent", silentMs: 50, outcome: "declined"}` over and over — while the real silence is 50, 100, 150… Someone reading a retained window to place an event against it takes the last record as 50ms of quiet when the channel has been quiet for seconds. The debug line has the same shape, but it's a line in a log; these are the records the module doc says are meant to be kept and matched on.

Measuring it fixes both halves at once — the repeats stop being duplicates and start being the thing worth reading, the same way `path-stale` is. `armStallWatch` already replaces the timer on every own-channel frame, so the stamp is right there:

```ts
// set where the timer is (re-)armed
const armedAt = Date.now();
…
this.trace({ phase: "channel-silent", silentMs: Date.now() - armedAt, outcome: "declined" });
```

With a `lastOwnFrameAt` stamp it reads as true elapsed silence across re-arms rather than per-window, which is the `path-stale` semantics exactly.

If you'd rather keep the constant, then `windowMs` (or `stallMs`) says what it is and stops colliding with the measured sense — but I'd take the measurement, since `declined` repeating forever is the case that needs to be readable.

---

src/transport/p2p/__tests__/shared_snapshot.spec.ts
```diff
@@ -45,4 +50,61 @@ describe("captureSnapshotFromShared (V6 snapshot as consumer)", () => {
...
+    const held = source.consumerCount;
+
+    const capture = captureSnapshotFromShared(source, { timeoutMs: 1000, ffmpegPath: executable });
+    expect(source.consumerCount).toBe(held + 1);
+    await settle(300);
+
+    expect(source.consumerCount, "the station is free while the decode is still running").toBe(held);
+    await expect(capture).rejects.toBeInstanceOf(Error);
+    watcher.detach();
+    rmSync(decoder, { force: true, recursive: true });
```
martijnpoppen, line 97: Optional, test hygiene only — the assertion this spec exists for is the right one and it genuinely fails without the change (I ran it against the base `media.ts`).

`rmSync` sits after two `await`s with no `try`/`finally`, so the temp dir survives any failure — and this spec fails loudly by design when the behaviour regresses, which is exactly when it leaks. A `try { … } finally { rmSync(…) }` around the body, or `onTestFinished(() => rmSync(decoder, { force: true, recursive: true }))`, keeps it to one line.

Also the first spec in the repo to write and exec a `#!/bin/sh` stub. Fine against ubuntu-only CI, and I can't see a neater way to get a deliberately slow decoder, so this is a note rather than a request — it just means `npm test` stops being runnable on Windows from here.

---

martijnpoppen (approved):
Approving — the fix is right, and it's right for a reason the code already stated.

`abortable`'s own doc says the underlying wait is deliberately left running because "these are shared negotiations whose result other callers are also waiting on". That holds for the `awaitLevel2Key` calls it still wraps, and it was never true of a snapshot capture: that pull belongs to one caller and nobody else is waiting on it, so racing the promise abandoned the *waiting* while the capture kept the channel to its own 20s keyframe timeout. Dropping it from `snapshotLive` and pushing the signal into the collection is the version that matches the rationale.

Checked the mechanics rather than the description. `throwIfAborted()` runs before `attach()`, so an already-abandoned call takes no consumer. `onAbandoned` goes through the same `cleanup()` as every other settle path, so the listener set and both timers come off together and the abort listener is removed on the success path too. The `let burst` + `finally { detach() }` restructure puts the release between collection and decode without changing what a failed decode does. Rejection value is unchanged — `abortable` rejected with `signal.reason` and so does this — so the `AbortableCall` contract reads the same to a caller. No exhaustive `switch` over `LiveTrace["phase"]` anywhere in `src/`, so the new member only lands on the surface spec, as the PR says.

Verified locally rather than taking the numbers: `npm run verify` green (3515 tests / 187 files — a little above the PR body, the base has moved since). More usefully, I reverted `media.ts` + `command-router.ts` to the base and re-ran the three new specs: all three fail, including the slow-decoder one at 307ms, so each is really load-bearing and not just passing alongside the change.

One non-blocking thing inline: `channel-silent`'s `silentMs` is the configured window, while `path-stale`'s is measured elapsed time — and since the `declined` branch re-arms, it emits byte-identical records for as long as nothing is attached. Worth measuring it so the repeats carry the information they look like they carry. Plus a test-hygiene nit on the temp dir. Neither needs to hold this up.

---

PR #168, opened 2026-09-13
feat(p2p)!: serve several cameras on a station at once, a connection each

martijnpoppen:
## What changed

A camera asked for while its station's **own session** is already carrying a live pull now gets a P2P session of its own, instead of being refused with `StationBusyError`.

- `SessionManager` keys by session rather than by station. Each entry records the station it connects to, so the power tier and idle window come from the hardware.
- The router opens a **media session** — key `<stationSn>#live:<channel>` — for a camera that finds the shared connection claimed. It carries media alone: wired for errors and its own teardown, nothing else.
- It waits for its level-2 key before being handed out, and is closed when its pull stops.
- Only the continuous-pull egresses may open one. A still must not: a socket and a key negotiation per thumbnail is not worth one frame.
- `StationBusyError` is removed — it was the refusal this replaces.

**BREAKING:** `StationBusyError` is gone; a caller catching it can drop that branch. `getP2pSessions()` also lists media sessions, keyed `<stationSn>#live:<channel>` — a caller iterating that map now sees entries that are not station serials.

## Why

A station serves one camera **per session**, not per station. It answers the most recent media start on a session, so two cameras sharing one take it from each other in turn — which is what the measurement behind `StationBusyError` recorded. That measurement was sound; the enforcement generalised it from the session to the station, and the contract's own wording ("over one session") said so.

## Evidence

A Smart Display E10 was captured holding **three concurrent streams off one HomeBase at three different geometries** — 2304×1296, 1600×1200 and 3840×2160 — by opening one PPCS session per camera. Different sizes at once rules out a composed stream. That capture now lives in `openMediaSession`'s JSDoc rather than only here.

Reproduced through this SDK from a **single client**, two attached cameras off one HomeBase:

```
ch 0   429 frames  15.1 fps   3840×2160
ch 4   395 frames  15.3 fps   2048×1536 → 1600×1200
max concurrent: 2 of 2
per channel: ch0 429, ch4 395 — matching each handle exactly
```

The same command returned a refusal and one stream before this change. Rates match a client-per-camera arrangement, so the extra session costs nothing measurable, and the per-channel totals matching the handles exactly verifies `LiveStream`'s per-camera filter on hardware.

## Verified how

- [x] `npm run verify` green — 3532 tests, 189 files, all five guards, typecheck, build, examples
- [x] Ran against a real device — a HomeBase with two attached cameras (one 4K), 30 s, both at full rate concurrently; previously a refusal
- [x] Added or updated specs — `station-serves-one-camera.spec.ts` → `station-serves-several-cameras.spec.ts`, rewritten and extended; new `media-session-opens.spec.ts` fakes `P2PSession` so the factory really runs; `session-manager.spec.ts` gains the power-tier cases
- [x] Docs updated — `live()` and `snapshotLive` in `core/contracts.ts`, `getP2pSessions()`, and the `docs/live-media.md` section on several cameras behind one station

Every new assertion was mutation-tested. **Three of the first attempts survived their mutation and were rewritten** rather than kept.

## Review round — what it found

Six findings, four of them bugs I introduced. All addressed; thread-by-thread replies inline.

- **`onAutoClose` half-wired.** The contract was changed to hand over a session *key* and the call site still routed it into `tearDownStation` — orphaning the live source and emitting a `p2pClose` for a non-station key. Now dispatched on the key.
- **`station = key` default.** `retain`/`hold` called `entry()` with one argument, so a hold on a media key filed that key as its own station, `poweredFor` answered `wired` for a battery HomeBase, and the session never idle-detached. Station is now required.
- **Media session never waited for level-2.** An attached camera's start has no level-1 form, so the second camera's first start was dropped. Visible in the live run above as 4.4s to first frame against the first camera's 2.0s — which I had recorded as variance.
- **Simultaneous starts collided.** Both read the station's session as free, since consumers attach after the call returns. The choice is now reserved synchronously. Writing that test found an **older** fault in the same window: the newcomer's linger sweep disposed the source its caller was still holding, because a never-attached source reads as a pull nobody wants.
- **Stills had silently started opening their own connection**, reversing "a live view outranks a tile". Restricted to continuous pulls.
- **A media session on a wired station outlived its camera** — the infinite idle window meant never reclaimed. `SharedLiveSource` now reports its own stop.

Plus ten stale "a *station* serves one camera at a time" claims (true of a session, false of a station) and a cleanup pass.

## Unverified or deferred

- **Two cameras, not four.** Nothing in the design caps it, but four concurrent sessions from one client is untested — and four is where a per-session cost would appear.
- **Battery stations.** A media session's socket and heartbeat now last exactly as long as its pull, but the drain of two concurrent heartbeats on a battery-backed base is unmeasured.
- **`liveSessionKeys` is still a third map** beside `liveSources` and `liveSourceOpts`. It changed meaning during review — it is now the reservation that makes simultaneous starts safe — and folding the three into one record touches every call site, so it is offered as a follow-up rather than buried in this push.
- **Deliberately narrower than keying every live session by station+channel.** Talkback, snapshot and record ride the station session; moving live off it unconditionally would split one camera's audio and video across two connections.

https://claude.ai/code/session_01SQWp63MFJewgUir74u4CKQ


src/transport/p2p/command-router.ts
```diff
@@ -830,21 +898,25 @@ export class P2PCommandRouter {
...
    *
    * A stopped source is skipped even when consumers are still attached to it. A failed start fails its
    * consumers without detaching them, so a caller still holding a dead handle leaves the count non-zero,
-   * and counting that as a viewer would refuse every later stream on the station until the client
-   * restarted. Only a source that can still deliver holds a place.
+   * and counting that as a viewer would put every later camera on a connection of its own until the
+   * client restarted. Only a source that can still deliver holds a place.
    */
-  private occupiedSiblingChannel(parentSn: string, key: string): number | undefined {
+  private stationSessionInUse(parentSn: string, key: string): boolean {
```
max246, line 968: **Concurrency gap — the feature's headline case can still collide.** Two sibling `live()` starts in flight at once (e.g. `Promise.all([live(camA), live(camB)])`) both see `stationSessionInUse()===false`: consumers only attach *after* `sharedLiveSourceFor` returns, so each other's source is still at `consumerCount===0` and is skipped by the `consumerCount === 0` guard here. Both then take `sessionKey = parentSn` (the shared own session) and contend — the exact stutter the PR removes, under 'open several at once'. Sequential opens are fine; simultaneous ones aren't. Worth either serialising the session-key decision + retain per station, or accepting it explicitly with a note.

src/transport/p2p/command-router.ts
```diff
@@ -881,13 +953,46 @@ export class P2PCommandRouter {
...
+  }
+
+  /**
+   * Drop the live source a media session was carrying, after that session closed on its own.
+   *
+   * The source holds the closed connection and never re-resolves it, so it can only answer its retained
+   * keyframe and then fail on its warm-up deadline. Its consumers get `stop`, and the next attach builds
+   * a fresh source — which, finding the station busy again, opens a fresh media session for it.
+   */
+  private tearDownMediaSession(sessionKey: string): void {
```
max246, line 1048: **Lower-severity: a media session on a WIRED station lingers open after its camera stops.** A clean stop takes the source to `state === 'stopped'` but does not drop it (`dropLiveSource` only runs on the next open for this key, `releaseLingeringSiblings`, or station teardown), and `onIdle → release` arms an **infinite** idle window on a wired tier — so `tearDownMediaSession` never fires here. The second socket + its 5 s PING keepalive stay open until the same camera is re-requested, a sibling opens, or the station tears down. Bounded and self-healing, but it's extra idle sockets/keepalives the pre-PR one-at-a-time design never held; worth confirming that's acceptable (esp. before pointing this at battery cameras).

src/transport/p2p/session-manager.ts
```diff
@@ -71,25 +74,36 @@ export interface SessionManagerOpts {
...
+   * only the one that closed is stale, so an owner told the station would tear down connections that are
+   * still serving.
    *
    * Those two are the only closes with no caller to follow up: everything riding the session is stale the
    * moment it goes, and only the owner knows what that is. A close a caller asked for is that caller's to
    * clean up after, which is why this does not fire for {@link SessionManager.close},
    * {@link SessionManager.closeAll}, or a superseded open.
    */
-  onAutoClose?: (parentSn: string) => void;
+  onAutoClose?: (key: string) => void;
```
max246, line 94: **Needs fixing before merge — the router doesn't honour this new key contract.** This signature now correctly hands `onAutoClose` a session **key** (a media key can be `<sn>#live:<ch>`), but the router still wires it as `onAutoClose: (parentSn) => this.tearDownStation(parentSn)` (`command-router.ts:260`) with no `isMediaSessionKey` dispatch. So when a media session idle-closes on a **battery-tier parent**, `tearDownStation('<sn>#live:<ch>')`:
1. leaves the owning live source (keyed `<sn>:<ch>`) **orphaned** — its prefix scan looks for `<sn>#live:<ch>:` and never matches `<sn>:<ch>`, so it serves a stale keyframe then fails its deadline; and
2. calls `deps.onClose('<sn>#live:<ch>')`, a spurious `p2pClose` for a non-station key — against the contract (in `getP2pSessions()` and this method's own JSDoc) that media sessions raise no connection events.

Fix at command-router.ts:260: `onAutoClose: (key) => P2PCommandRouter.isMediaSessionKey(key) ? this.tearDownMediaSession(key) : this.tearDownStation(key)`. Latent today (needs a battery HomeBase serving 2+ cams) and **untested** — the new spec never drives a media session through this idle path; please add that case.

(Minor, while here: the `releaseLingeringSiblings` debug string and the `sharedLiveSourceFor` docstring still say "one station serves one camera at a time", which this PR makes false — worth updating so the lingering-sibling drop isn't misread as 'concurrency unsupported'.)

src/transport/p2p/command-router.ts
```diff
@@ -368,24 +379,64 @@ export class P2PCommandRouter {
...
+        let dskKey: string | undefined;
+        try {
+          dskKey = (await this.deps.mega.getDskKeys([parentSn]))[parentSn]?.dskKey;
+        } catch (e) {
+          this.deps.onError(e instanceof Error ? e : new Error(String(e)));
+        }
+        const localAddress = this.deps.localAddresses?.[parentSn] ?? freshestLanIp(raw);
+        const session = this.makeSession(parentSn, did, raw, dskKey, localAddress, key);
+        register(session);
+        await session.connect();
```
lenoxys, line 457: The station's session is resolved with `requireLevel2ForAttached: true` and refuses when no key arrives. A media session gets `await session.connect()` and nothing else, and an attached start has no level-1 form, so `sendMediaPayloadLevel2` drops it with `media-command-unsent` and reason `level2-key`.

So the ordinary path for the second camera is: connect, start dropped, silence until a keepalive tick happens to find the key. No test covers either outcome. This is the one I would call a bug rather than a rough edge.

src/transport/p2p/session-manager.ts
lenoxys, line 94: `onAutoClose` was rewritten to hand over the key rather than the station, and this call site still routes it into `tearDownStation`. Probed on your head: a battery station's media session idling out emits `p2pClose "T8010P0000000000#live:2"`.

That handler can do nothing with the key either. Its `liveSources` sweep tests `startsWith("<key>:")` and matches nothing, and the `close` listener that would reach `tearDownMediaSession` never runs because the entry is discarded first. The only observable effect is a public event carrying something that is not a station serial.

src/transport/p2p/__tests__/station-serves-several-cameras.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+    onConnect: () => {},
+    onClose: () => {},
+    onError: () => {},
+    onLevel2Ready: () => {},
+    onFrame: () => {},
+  };
+  const built = new P2PCommandRouter(deps);
+  const manager = (built as unknown as { manager: Manager }).manager;
+  for (const sn of [STATION_SN, SOLO_A, SOLO_B]) manager.register(sn, session);
+  for (const channel of [0, 2]) manager.register(`${STATION_SN}#live:${channel}`, session, STATION_SN);
```
lenoxys, line 55: Both keys are pre-registered with the same `connectedSession`, so `acquire` returns on the existing entry and the factory is never called. `openSession`, `makeSession`'s `announces === false` branch, the media session's `close` handler and `tearDownMediaSession` are executed by no test in this PR.

Every new assertion reads `liveSessionKeys`, which is bookkeeping. Nothing here would notice if a second connection was never opened, which is the claim in the title. One case where the factory really runs and two distinct fake sessions come back closes it.

src/core/contracts.ts
```diff
@@ -810,12 +777,12 @@ export interface MediaProvider {
...
-   * Several cameras behind one station may stream at the same time only where the station serves them at
-   * the same time. Where it serves one camera at a time, a second viewer is refused with
-   * {@link StationBusyError} rather than admitted and degraded: accepting it does not make the station
-   * serve two, it makes both stutter. Which camera deserves the station is the caller's decision, so
-   * nothing is queued or pre-empted. Each handle receives only the frames the station tagged for ITS
-   * camera.
+   * Several cameras behind one station stream at the same time, each over its own connection to it. One
+   * connection serves one camera — a station answers the most recent start on a session, so two cameras
+   * sharing one take it from each other in turn — so a camera asked for while its station is already
+   * serving another gets a connection of its own. Measured on a base carrying two attached cameras, one
```
lenoxys, line 784: Ten lines above this, `snapshotLive`'s `retained` still says a live still is refused while a sibling is watched "because a station serves one camera at a time". Same interface, same file, opposite claims.

Seven more: `media.ts` says the SDK refuses a second channel on a busy station, which is load-bearing for why a still detaches before decoding; `command-router.ts` logs "one station serves one camera at a time" as the reason it drops a sibling, at runtime, including siblings that no longer contend; plus `live-stream.ts` in three places, `shared-live-source.ts`, `eufy-mega.ts`, and the section of `docs/live-media.md` about exactly this.

Also: the display-side capture that rules out a composed stream is the load-bearing evidence for the reversal, and it lives only in the PR body. The two-camera run is recorded here and in the commit. That one deserves the same treatment.

src/transport/p2p/session-manager.ts
```diff
@@ -139,10 +157,12 @@ export class SessionManager {
    * `connect()` and resolves the connected session.
    */
   async acquire(
-    parentSn: string,
+    key: string,
     factory: (register: (session: P2PSession) => void) => Promise<P2PSession>,
+    station: string = key,
   ): Promise<P2PSession> {
-    const e = this.entry(parentSn);
+    const parentSn = key;
```
lenoxys: `const parentSn = key` keeps the old log strings compiling and makes them lie: `P2P session start superseded for station T8010P0000000000#live:2` is not a station. The JSDoc above still documents a `parentSn` parameter that no longer exists, and every other method still names its key `parentSn`.

Rename once and let the strings break where they are wrong.

src/transport/p2p/session-manager.ts
```diff
@@ -106,31 +120,35 @@ export class SessionManager {
...
 
-  /** Get or create the lifecycle entry for a station. */
-  private entry(parentSn: string): SessionEntry {
-    let e = this.entries.get(parentSn);
+  /**
+   * Get or create the lifecycle entry under `key`, recording which station it connects to. An existing
+   * entry keeps the station it was opened with — the key owns one connection for its lifetime, and a
+   * later caller passing a different station would otherwise re-point a live entry's power tier.
+   */
+  private entry(key: string, station: string = key): SessionEntry {
```
lenoxys: `station = key` is a guess in the one place a wrong answer is invisible. `retain()` and `hold()` call `entry()` with one argument, so a retain on a media key with no entry files `station: "...#live:2"`, `poweredFor` then answers `wired` for a battery station, and that session never idle-detaches. Battery drain is the failure this class exists to prevent.

Make `station` required. Three call sites pass it already.

src/transport/p2p/command-router.ts
```diff
@@ -399,14 +450,22 @@ export class P2PCommandRouter {
...
    */
   private makeSession(
     stationSn: string,
     did: string,
     raw: Record<string, any>,
     dskKey: string | undefined,
     localAddress: string | undefined,
+    key: string = stationSn,
   ): P2PSession {
+    const announces = !P2PCommandRouter.isMediaSessionKey(key);
```
lenoxys: Ponytail pass, all in this file:

`announces` is recovered by string-sniffing the key it was just built from, and `makeSession`'s `key = stationSn` default is dead since the only call site always passes one. Pass the boolean, drop the default.

`liveSessionKeys` is the third map keyed on station and channel beside `liveSources` and `liveSourceOpts`. Two had precedent, three is a record wanting to exist.

Two key encodings now coexist, `:` for sources and `#live:` for sessions, each parsed by `startsWith` or `includes` at four sites, and one of them leaks into `no P2P session for X (known: ...)`.

`tearDownStation`'s `manager.keys()` sweep is unreachable: the `liveSources` loop above it already closed and discarded every media session synchronously. `openStation` is now a middle man for `openSession(sn, sn)`.

Net: about seventy lines, and one fewer thing to be subtly wrong.

Last: the `BREAKING:` footer names `StationBusyError` only. `getP2pSessions()` changing key shape breaks any caller iterating that map, and the body says so where the footer does not.

lenoxys (changes requested):
Requesting changes, which I have not done on your PRs before. The reversal itself I believe: two cameras, one at 4K, both at full frame rate is the right shape of evidence, and rereading the old measurement as "one session serves one camera" rather than "one station serves one camera" is the honest reading of it.

Three things stop it. The second camera's first start is silently unsent on the common path, because a media session never waits for the level-2 key the attached start requires. A public event now fires with a key that is not a station. And the claim in the title, a connection each, is executed by no test: the fixture files both keys against one session object, so the factory never runs.

Beyond that, the invariant you deleted is still asserted in eight places this diff does not touch, one of them a runtime log line giving it as the reason it drops a sibling.

src/transport/p2p/command-router.ts
martijnpoppen: Right, and fixed by serialising rather than by a note.

The session-key choice is now written to `liveSessionKeys` **synchronously, before the connection is opened**, so it is a reservation rather than a record; and `stationSessionInUse` counts an `idle` sibling as holding the session, since a source handed to its caller but not yet attached to is exactly what a simultaneous start looks like from the other side.

Writing the test for it found a second fault in the same window, older than this PR: the newcomer's `releaseLingeringSiblings` **disposed the source its caller was still holding**. A source that has never been attached to reads as a pull nobody wants, but it is not a linger — it has never warmed, so it has sent no media start and holds nothing. `idle` sources are now skipped there too. Without that, `Promise.all([live(a), live(b)])` failed with "SharedLiveSource is disposed".

Cost, taken knowingly and written into the JSDoc: a pull genuinely abandoned before its first attach keeps the station's own session, and the next camera pays for a connection instead of reclaiming it. One socket against destroying a start someone is waiting on.

Covered by `media-session-opens.spec.ts` → "gives simultaneous starts one connection each", and the two halves are split in `station-serves-several-cameras.spec.ts`.

---

src/transport/p2p/command-router.ts
martijnpoppen: Confirmed and fixed — and I'd call it more than lower-severity, because the wired tier is the common case, so the infinite window means "never reclaimed" rather than "reclaimed late".

`SharedLiveSource` now reports its own stop (`onStopped`), and the router closes the media session there. `onIdle` was the wrong hook: it fires when the last consumer leaves while the linger is still holding the pull open for a cheap re-attach, so closing there would pay for the linger with the connection shut.

Pinned by "closes its connection when the pull stops, which a wired station would never reclaim", which fails without the hook.

On the battery question you raise at the end: with this, a media session's socket and heartbeat last exactly as long as the pull. What I still cannot answer is the drain of two concurrent heartbeats on a battery-backed base — untested, and called out in the PR body.

---

src/transport/p2p/session-manager.ts
martijnpoppen: You're right, and it was mine: I changed the contract to hand over a key and never updated the call site. Fixed exactly as you wrote it:

```ts
onAutoClose: (key) =>
  P2PCommandRouter.isMediaSessionKey(key) ? this.tearDownMediaSession(key) : this.tearDownStation(key),
```

Both consequences you named were real — the orphaned source from the prefix scan that cannot match, and the `p2pClose` carrying something that is not a station serial, against the contract this PR's own JSDoc states.

Now tested: "tears down only its own source when the manager idle-closes it, and reports no station close" drives a media key through the handler and asserts both the surviving sibling and the absence of any event. It fails if the dispatch is reverted.

The stale "one station serves one camera at a time" strings are gone too — ten of them, including the runtime debug line you point at. They now say *session*, which is what stayed true.

---

src/transport/p2p/command-router.ts
martijnpoppen: Agreed, this was the bug. `openMediaSession` now waits on the same terms the station's own session gets — `awaitLevel2Key`, one reprompt, then `StationKeyUnavailableError` — so it refuses rather than handing back a connection that cannot carry a start.

Worth adding that the live run in the PR body already showed it and I misread it: the second camera took **4.4s** from open to first frame against the first camera's **2.0s**. I recorded that as variance. It is the shape you describe — first start dropped, picture waiting for a keepalive tick to find the key.

Now covered by "refuses a media session that connects without a level-2 key, rather than returning a mute one". My first attempt at that test passed with the wait removed, because the fake always had a key; the fake now takes a per-session flag, and the test fails without the wait.

---

src/transport/p2p/__tests__/station-serves-several-cameras.spec.ts
martijnpoppen: Correct, and the sharpest of the comments — I noted the pre-registration when writing that fixture and did not follow the thought through to "so the claim in the title is untested".

`media-session-opens.spec.ts` closes it the way you describe: `P2PSession` itself is faked, so `acquire` finds no entry, the factory runs, and `openSession` / `makeSession` / the `announces === false` branch / the media `close` handler / `tearDownMediaSession` all execute. It asserts **two constructions and two distinct instances**, not a map entry.

It also pins what the second connection must not do — `connect`, `data` and `level2Ready` on it produce no events, while the same on the station's session produces one.

Related, from your other thread: I mutation-tested every new assertion this round. Two of my first attempts survived their mutation and were rewritten rather than kept.

---

src/core/contracts.ts
martijnpoppen: All ten corrected, plus the guide. They now say *session* rather than *station*, which is the half that stayed true — one session does serve one camera at a time; what this changes is that a station is no longer limited to one session.

Chasing them turned up something worse than a contradiction, in the `snapshotLive` line you point at. Stills go through the same `sharedLiveSourceFor`, so a **thumbnail refresh had silently started opening its own socket and level-2 negotiation** — the doc said "a live view outranks a tile" while the code had stopped doing that. Only the continuous-pull egresses may open a connection now (`mayOpenOwnSession`), so a still contends as it always did and the retained image answers it. That was your comment's real payload and I would not have found it from the code alone.

On the evidence: agreed, and moved. `openMediaSession`'s JSDoc now carries the display capture — three cameras in the same second at 2304×1296, 1600×1200 and 3840×2160, which no composed stream can be — beside the two-camera reproduction that was already there. It was the observation the whole change rests on and it was living in a PR body.

---

src/transport/p2p/session-manager.ts
martijnpoppen: Fixed — `onAutoClose` now dispatches on the key, so a media key reaches `tearDownMediaSession` and only a station serial reaches `tearDownStation`.

Thanks for probing it on the head rather than reasoning about it; the `p2pClose "…#live:2"` you observed is exactly what the new spec asserts is absent.

Your point about the `close` listener never running because the entry is discarded first is why the fix belongs at the dispatch and not on the listener.

---

src/transport/p2p/session-manager.ts
martijnpoppen: Fixed — `station` is required, and threaded from all four call sites (`retain`, `release`'s lookup, `hold`, `bumpCommand`).

Your reasoning about *where* it is invisible is what made the test worth writing. My first version registered the entry before retaining, so `entry()` found it already correctly filed and the mutation survived. Rewritten so `retain` creates the entry, which is the pre-warm shape you describe; it now fails if `retain` files under the key.

Two cases in `session-manager.spec.ts`: `poweredFor` is asked about the station and never the key, and a media session on a battery station closes on that station's window.

---

src/transport/p2p/session-manager.ts
martijnpoppen: Done — shim gone, renamed once, and the strings did break where they were wrong. `P2P session start superseded for station …` is now `for …`, since the subject may not be a station.

`poweredFor` deliberately kept its `parentSn` parameter name: it is the one thing here genuinely asked about a station rather than a key, and now that the entry records the station separately, that name is accurate instead of a leftover.

---

src/transport/p2p/command-router.ts
martijnpoppen: Most of it done; one deliberately not, and one I'd rather not do here.

**Done:** `announces` is passed rather than string-sniffed from the key it was built from, and `makeSession`'s dead `key = stationSn` default is gone. `tearDownStation`'s `manager.keys()` sweep was indeed unreachable — the `liveSources` loop above closes every media session on that station synchronously — removed. `openStation` is gone; the three station opens call `openSession(sn, sn)` directly. The `BREAKING:` footer now names the `getP2pSessions()` key-shape change beside `StationBusyError`.

**Not done — `liveSessionKeys` stays a third map.** It changed meaning while fixing the concurrency thread: it now records *every* source's session and is written before the connection opens, so it is the reservation that makes simultaneous starts safe, not bookkeeping beside the other two. Folding three maps into a record is right, but it touches every call site of `liveSources` and `liveSourceOpts`, and doing it in the same push as four behavioural fixes would bury them. Happy to do it as a follow-up if you'd rather it land now than later.

**Two key encodings:** also still true. `:` for sources and `#live:` for sessions are now genuinely different things — one keys a pull, the other keys a connection — so I'd argue they should not collapse into one; but the parsing being spread over `startsWith`/`includes` at several sites is fair, and that goes with the record refactor.

---

max246 (approved):

src/transport/p2p/__tests__/media-session-opens.spec.ts
```diff
@@ -0,0 +1,199 @@
...
+  (
+    r as unknown as {
+      sharedLiveSourceFor(s: string, o: object, m: boolean): Promise<{ attach(): unknown; dispose(): void }>;
+    }
+  ).sharedLiveSourceFor(sn, {}, true);
+
+const sessionKeys = (r: Router) => (r as unknown as { liveSessionKeys: Map<string, string> }).liveSessionKeys;
+const sources = (r: Router) => (r as unknown as { liveSources: Map<string, unknown> }).liveSources;
+
+describe("a second live camera on a claimed station session", () => {
```
lenoxys, line 95: This is the question. The sweep still takes every lingering sibling on the station, and its doc justifies that by contention over one session. A sibling lingering on its own media connection contends for nothing, so disposing it and closing its socket destroys exactly the cheap re-attach the linger exists for.

Nothing pins it either way. Whichever answer is right, it wants a test: either the sibling on its own connection survives, or the doc stops giving contention as the reason.

src/transport/p2p/shared-live-source.ts
```diff
@@ -129,6 +129,13 @@ export interface SharedLiveSourceOptions {
    * Not called by {@link SharedLiveSource.dispose}: the owner asked for that one, and it is the very thing
    * an owner does in response to this callback.
    */
+  /**
+   * Called once this source has stopped pulling — its linger elapsed, its budget ran out, or it was
+   * torn down. Distinct from {@link onIdle}, which fires when the last consumer leaves and the linger
+   * is still holding the pull open for a quick re-attach: a caller that owns a connection for this
+   * source alone must release it here, not there, or it pays for the linger with the connection closed.
+   */
+  onStopped?: () => void;
```
lenoxys, line 141: `onStopped`'s block was inserted between `onStartFailed`'s existing JSDoc and `onStartFailed` itself. The old block now dangles above the new field and `onStartFailed` ships undocumented.

The new doc is also written to its audience: "a caller that owns a connection must release it here" names who calls it and what they do with it, where the rule wants what the value is and what it guarantees.

src/transport/p2p/session-manager.ts
```diff
@@ -102,75 +116,80 @@ export class SessionManager {
...
-   * Ensure a session to `parentSn` is open, building it via `factory` if cold. Concurrent calls for the
+   * Ensure a session to `key` is open, building it via `factory` if cold. Concurrent calls for the
    * same cold station share ONE connect (the `connecting` promise); `factory` builds + wires + awaits
    * `connect()` and resolves the connected session.
    */
   async acquire(
-    parentSn: string,
+    key: string,
     factory: (register: (session: P2PSession) => void) => Promise<P2PSession>,
+    station: string = key,
```
lenoxys: `station = key` is still defaulted here and on `register`, so the footgun the commit says it removed is latent rather than gone: it survives only because both callers happen to pass explicitly. A media key filed as its own station reads `wired` and never idle-detaches, which is the drain this class exists to prevent.

While here: `retain` and `bumpCommand` take `station` only because they might create an entry, and only `acquire` and `hold` ever do. Dropping it there removes about thirty mechanical test edits and one duplicated fact.

One leftover from the old list: composite keys still reach a caller through `manager.keys()`, in `no P2P session for X (known: ...)`. Filter to station-shaped keys or say the shape.

src/transport/p2p/live-stream.ts
```diff
@@ -188,7 +188,7 @@ export class LiveStream extends EventEmitter {
    *
    * The nudge differs by topology and only one branch is a ping: an own-session camera sends a small
    * keepalive, while an attached camera has no such state and re-sends the FULL media start. On a station that
-   * serves one camera at a time that restart re-asserts this channel against whatever else is warm, so two
+   * serving one camera at a time, that restart re-asserts this channel against whatever else is warm, so two
```
lenoxys: "On a station that serving one camera at a time" is garbled, and still says station where the claim is now per session.

Six specs also still assert the station-level invariant in their prose: `stored-snapshot-lifecycle`, `attached-keepalive-stall`, `live-sibling-channel`, `live_stream` twice, and `shared_snapshot`. Specs do not ship, but they are where the next reader looks for what the SDK believes.

lenoxys (commented):
Unblocking. All three blockers are answered, and I checked the mechanism rather than the reply: the media session's key wait is real, because the new spec's fake connects with no key and the assertion only passes if something waits; `p2pClose` no longer carries a composite key; and the new spec mocks `P2PSession`, so two distinct connections are actually constructed rather than pre-registered. The display capture is in the source now.

Four things left, none of them a hold, and one question I would like pinned either way before merge.

src/transport/p2p/__tests__/media-session-opens.spec.ts
martijnpoppen: You're right, and the answer is that the sibling survives — the sweep now stops at the session boundary:

```ts
if (this.liveSessionKeys.get(key) !== parentSn) continue;
```

Its own doc already gave the reason and I hadn't followed it: the contention being relieved is contention over **one session**, and a sibling on a connection of its own is not on that one. Dropping it closed a socket and destroyed the cheap re-attach the linger exists for, to settle a competition that was not happening.

Pinned by "leaves a sibling lingering on its own connection alone, having nothing to contend with" — it drives two cameras onto separate connections, lets the second go to linger, reopens the first, and asserts the second's source and its connection are both still there. Fails without the guard.

---

src/transport/p2p/shared-live-source.ts
martijnpoppen: Both correct, both mine. `onStopped` now sits after `onStartFailed` with its own block, so `onStartFailed` has its documentation back.

The doc is rewritten to state the declaration: what has happened to the pull, that it will not resume, and how it differs from `onIdle` **by what is still running** — `onIdle` fires while the linger is still holding the pull open, this fires when it is not. No caller, no instruction.

Inserting a block between an existing doc and the field it documents is a bad failure mode, because the result still compiles and still reads as documented. Worth me grepping for rather than trusting.

---

src/transport/p2p/session-manager.ts
martijnpoppen: Fair on all three — the last commit claimed to remove a footgun and left it on the path that matters.

**`acquire`:** station is required now. That is the only path that opens a session under a key which is not a station serial, so it was the reachable hazard.

**`register`:** default kept, and it now says why it is safe there and nowhere else — every non-serial key goes through `acquire`, and the router never calls `register` at all (21 of its 22 call sites are specs seeding a station under its own serial). If you'd rather have it required regardless, say so and I'll take the churn; I judged 21 sites repeating a serial to be worse than the sentence.

**`retain`:** you're right that it took a station only because it might create an entry — so it no longer creates one. A retain names a session someone acquired, so it is now refused when nothing is open, mirroring `release`. `hold` remains a creating path and opens the entry itself before retaining it. The ~30 mechanical edits are reverted with it.

`bumpCommand` keeps its station: it is a thin wrapper over `hold`, which still creates.

**`manager.keys()`:** filtered to station-shaped keys in that message.

---

src/transport/p2p/live-stream.ts
martijnpoppen: Garble was mine, from the previous round's rewrite — fixed, and it says session now:

> the nudge … re-sends the FULL media start. Over one session, which serves one camera at a time, that restart re-asserts this channel against whatever else is warm on it

All six specs corrected too. Agreed on why they matter: they do not ship, but a spec's prose is what the next reader takes as the SDK's belief, and leaving them saying *station* would have meant the claim survived in the place people actually read.

A repo-wide grep for the old phrasing now returns nothing, across `src/`, the specs and `docs/`.

---

lenoxys (approved):
Approving. My earlier "Unblocking" went out as a COMMENT, which does not clear a CHANGES_REQUESTED — this is that, done properly.

I checked this round at the mechanism rather than at the replies. `announces` is a parameter passed as `key === parentSn`, not recovered from the key it was built from. `makeSession` has no `key` default left. `tearDownStation` is down to its two real loops, the unreachable `manager.keys()` sweep gone, and `openStation` returns zero occurrences. The `BREAKING:` footer names the `getP2pSessions()` key-shape change beside `StationBusyError`, with the consequence for a caller iterating that map. `acquire` requires its station, `retain` neither creates nor takes one, and composite keys no longer reach a caller through `known: ...`.

The invariant I said was still asserted in eight untouched places is now consistent everywhere, including the log line that gives it as the reason it drops a sibling — and that line agrees with the code, because the loop stops at the session boundary above it. The one place still saying "station" is the new distinction between one camera and several, not the old claim. The two RTSP mentions are a different fact and correctly left alone.

Three things I am accepting rather than conceding, so the record is clear:

`register` keeps its `station = key` default. `acquire` closes the reachable hazard, and the declaration now says why the default is safe there and nowhere else. Twenty-one sites repeating a serial to defend a path no production code takes costs more than the sentence, and I could not find a guard that makes it mechanical without teaching `SessionManager` the media-key shape, which it should not know.

`bumpCommand` keeps its station. I asked for both and got one; the reason holds — it is a thin wrapper over `hold`, which still creates, so the duplicated fact I was aiming at is not there.

The maps and the key parsing are #176, and I was wrong to call the parsing separable from the record — see the thread. Nothing in it is behavioural.

What I would still watch after merge: the reservation in `liveSessionKeys` is the property the whole simultaneous-start case rests on, and it is written before the connection exists. #176 is the change most likely to break it quietly.

PR #170, opened 2026-09-13
fix(battery): a mains camera has no cell, so it reports none of one

martijnpoppen:
## What changed

A mains camera serves **no cell reading under any public name**. Two halves, and only the first is this PR's:

1. **Here.** `T8410` joins `MAINS_CAMERA_MODELS`, and the cell gate covers every read whose subject is the physical cell — `level`, `charging`, `temperature`, `health`, `solarIntensity`, `solarConnected24h`, `batteryPowerStats` — where it previously covered the first two. Which reads those are is `CELL_PARAMS`, declared once, and `cellGated` applies `notMainsCamera` from it as the table is built.
2. **#175.** A param a resolved capability's gate withheld no longer falls through to the param dictionary's name. That half was mine in `8d0c494` and is @lenoxys' now — see **Why `device.ts` is no longer here**.

#175 is merged into this branch so CI runs against the behaviour the specs below assert. Those commits leave this diff the moment #175 lands on `beta-0.2.0`.

## Why

A live T8410 showed a battery level, a cell temperature, a state of health and both solar reads. It is an Indoor Cam Pan & Tilt: mains-only, no cell, nothing for a panel to charge.

**T8410 was not listed.** T8419, the Indoor Cam beside it, is. Confirmed mains-only by the maintainer.

**The guard only ever covered two of the seven cell reads**, so `T8425` — mains, confirmed on owned hardware, listed since the guard was written — still published a cell temperature, an ageing figure and a solar harvest. That also explains a separate report of **30 °C on every device at once**: `batteryTemperature` is 1138, a mains camera reports it as a fixed sentinel, and nothing stopped it becoming a temperature reading.

**`batteryPowerStats` (3100) was a seventh**, and `unexposed: true` looked like it covered it (@max246). It doesn't: `unexposed` suppresses the fluent GETTER, while `propertiesOf` filters `writeOnly` and `available` and deliberately not `unexposed`.

**And gating a member did not actually withhold the reading** (@lenoxys). This is the finding of the review, and the first revision of this PR did not fix the bug it was written for. `applyParams` names by precedence — curated `PropertySpec` → param dictionary → `unknown_<paramType>` — so removing a member's spec let the **dictionary** name the value instead. On the head that was reviewed:

```
getProperty("battery")        = 88      ← a camera with no cell
getProperty("solarIntensity") = 7
getProperties() → battery, batteryTemp, devBatteryHealthyV2,
                  solarIntensity, solarPanelConnect24h, batteryPowerDatas
```

For 1101 and 1309 the dictionary spells it **exactly as the member does**, so the gate changed nothing a caller could see. The rest came back renamed rather than withheld — `batteryPowerStats` as `batteryPowerDatas`, which undid the premise the seventh gate was argued on. The gate was *demoting* a reading, not removing one. **#175 fixes that**; this PR's guard is only true once it lands.

## Evidence

Measured on `getProperty`/`getProperties` — the surface a host reads. The first revision's table was measured with `propertiesOf`, the schema projection, which is how the gap above went unnoticed in it.

| Model | Cell readings served, before | After |
| --- | --- | --- |
| T8410 (Indoor Cam Pan & Tilt) | `battery`, `batteryTemp`, `devBatteryHealthyV2`, `solarIntensity`, `solarPanelConnect24h`, `batteryPowerDatas` | **none** |
| T8419 (Indoor Cam) | same six | **none** |
| T8425 (Floodlight Cam) | same six | **none** |
| T8114 (eufyCam) | all seven, under the member's own names | all seven |

Every model still serves `workingMode` and `powerSource`, which is the split that keeps the capability attached. On a mains camera the seven cell params remain in state as `unknown_1101`, `unknown_1138`, … — the graceful-unknown rule says never drop data, and an `unknown_` name is the one form that carries a value without claiming it means anything. `inspectParams` still reaches them.

T8410's evidence bar is the maintainer's, matching how T8419 was added rather than the owned-hardware bar T8425 met.

## Why `device.ts` is no longer here

@lenoxys wrote that fix an hour before I did — `bbf8028` at **13:31 UTC** against my `8d0c494` at **14:29 UTC** — having found the bug reviewing this PR at 07:04 and then gone and fixed it himself. Two implementations of one change, conflicting in `device.ts`. His is the better one, and the difference is not stylistic.

`withheldParams` was built from `mergeProperties(resolved.capabilities)`, which de-duplicates **by name, first-wins**. A param claimed under a name an earlier module already claimed never reached the set, so the dictionary could still name it. That collision is in the tree today:

```
battery  → battery:1101, lock:1101, vacuum_clean:163, display:8001
charging → battery:2111, keypad:1102
rssi     → contact:1141, lock:1141, keypad:1141, vacuum_clean:134
```

163 and 8001 are both `battery` in their own dictionaries. On a device resolving `battery` and `display`, `MODULES` order keeps 1101 and drops 8001 — so a gated display battery read leaks back under `battery`. His `claimedParams` collects per module, *before* the dedup, and has no such hole. Mine was silently sensitive to the order of the `MODULES` list, which is a latent bug and not a preference.

**The specs that commit added stay.** `getProperty`/`getProperties` under both the member's and the dictionary's spelling is the regression cover for what this PR claims, and it is worth having however the naming precedence is fixed. Confirmed green on his implementation with mine removed before removing it: 190 files, 3538 tests.

One behavioural difference, raised on #175 rather than argued here: 3100 carries `encoding: "json"`, and his skips `paramDef` entirely, so a withheld encoded param lands as the raw string rather than the decoded object. His is the consistent choice — every other `unknown_` is raw.

## What is deliberately NOT gated

`workingMode`, `recordDuration`, `recordInterval`, `recordAutoStop` — a mains camera really does have its working mode and recording settings.

`cameraInfo` (1103) — its meaning is unevidenced, so gating it would assert it *is* a battery fact, the kind of claim this guard exists to stop making.

**`powerSource` (1293) is out, and is the one open question** (@lenoxys). My first answer for it was wrong: I wrote that it says which supply is in use and a mains camera genuinely answers it. It does not — the values are `0 = Battery, 1 = External Solar Panel`, and *both* describe how a cell is fed, so by this list's own rule it arguably belongs in `CELL_PARAMS`. It stays out for two reasons that survive scrutiny: every param in the list is one a live mains camera was **observed** to publish and 1293 was not among them, and it is `requires`-gated on its own param so it appears only where the device reports it. Neither is evidence about what a mains camera does with 1293, and gating it would withhold a described **write** rather than a read. **Unresolved rather than settled** — a reviewer's call.

## Verified how

- [x] `npm run verify` green — 190 files, 3537 tests, 91 snippets, all guards, `check:esm`, examples
- [x] **The docs build too** (`npm -C docs ci && npm -C docs run build`), which is not part of `verify` by design
- [x] #175 merged in and re-verified, as was the base after #172 landed
- [x] Ran against a real device — the maintainer's T8410 reported the symptom; the fix withholds reads and drives no wire
- [x] Added or updated specs — see below
- [x] Docs updated — the guard's doc states what it covers, that it fails open, and that it is a static model fact

Each half of this PR confirmed to fail alone: dropping T8410 fails with `T8410P00 still publishes 1101`; bypassing `cellGated` fails across the family; dropping 3100 from `CELL_PARAMS` fails the pinned-contents check. Reverting #175's dictionary skip fails with `a mains camera served battery: expected { name: 'battery', … } to be undefined` — which is the dependency, stated as a test.

### One declaration, not two

The first revision declared "this read describes the cell" twice — `available: notMainsCamera` on the member and again as a `CELL_PARAMS` entry — with a spec holding the two equal. @lenoxys called that a table per concern with a test to keep the copies in step; it was also the weaker half of what @max246 originally asked for. `cellGated` applies the gate from the list instead, composing with a member's own `available` rather than replacing it.

The join by param id that remains is argued in the thread rather than changed: the alternative that removes it is a `cell?: true` tag, and `Members` is a closed `ValueMember` interface, so that means a battery-domain flag on the type every capability module is written against. Open to being overruled on that.

### What the specs pin

- the published schema, driven off `CELL_PARAMS` rather than a second list of names;
- **`getProperty`/`getProperties`** — both the member's spelling and the dictionary's, which is the half that made the gap invisible, and the half that depends on #175;
- **the binder** — `bindMembers` makes its own `available` decision and is what a caller holding `dev.battery()` touches. Pinned whole: a mains camera binds `powerSource`, `setPowerSource`, `workingMode`, `setWorkingMode` and nothing else;
- **the fail-open behaviour**, so it is a decision rather than a property of how the predicate is written;
- **`CELL_PARAMS`'s contents**, written out once — every other assertion iterates that list, so deleting an entry would delete its own check.

### Three stale fixtures, one load-bearing

`poll.spec.ts`'s `liveCamera` was model `T8410` with a reported level, captioned "a live battery camera" (@max246). Grepping the class found two more in `manifest.spec.ts`, one asserting `toContain("level")` on a `T8410` record — which passed only because `ctxWith` built a context with no `model`. `ctxWith` takes the model now; putting `T8410` back fails, which it could not do before.

Worth stating generally: the property schema is model-gated (through `fromRecord`) while a spec helper's bind context often carries no `model` at all.

## Unverified or deferred

- **The guard fails OPEN** (@lenoxys). A record with no `model` matches no prefix and publishes every cell read — the same hole this PR found in `ctxWith`, in the source rather than the specs. That is the safer direction (a real battery camera never silently loses its charge because its record came back thin), and it means the list can only ever be a **floor**. Pinned by a spec so it is chosen, not inherited.
- **It is a static fact about a model**, not a live reading of what is powering the device, so a unit whose supply can change is unhandled. Moot for these three, which have no cell at all.
- **The model list stays explicit** (@max246; not blocking). Composing `isFloodLight`/`isIndoorCamera` would over-reach, trading a phantom battery on three models for a *missing* battery on real ones.
- **The guard covers READS only.** `detection.evidenceParams` still names 1101 and the three `batteryAlert` events are ungated, so a mains camera still resolves as `battery` and can still deliver a cell alert. The detection half is load-bearing — detaching the capability would take the working-mode and recording settings with it — so both are pinned as deliberate.
- **`poweredOf` (camera.ts) still resolves these as `battery`**, so a live stream is budgeted as if cell-powered. An unnecessary power budget is cheap; refining it is a separate change.
- **Nothing here establishes what 1138 means on a device that DOES have a cell.** Still `apk` provenance, never confirmed against a moving value.


src/model/capabilities/battery.ts
```diff
@@ -368,6 +379,7 @@ export const BATTERY_MEMBERS = {
    */
   solarConnected24h: {
     param: BATTERY_PARAM.SOLAR_CONNECT_24H,
+    available: notMainsCamera,
```
max246, line 411: The gating added here doesn't cover `batteryPowerStats` (3100), a physical battery-cell read 11 lines down at line 393 — it's `unexposed: true` but has no `available: notMainsCamera`, and `propertiesOf` filters `writeOnly`+`available` but NOT `unexposed` (I reproduced this: the member stays in the schema and answers via `getProperty` on a mains camera). So the PR's "unexposed ⇒ no guard needed" reasoning covers the fluent-getter path but not schema/`getProperty` — same phantom-cell-read class this PR removes. Add `available: notMainsCamera` to it. More broadly: hand-adding the gate to each cell member is the exact omission that caused this bug (4 of 6 were missed); a structural marker (a `cell: true` tag the gate is applied from centrally, or one array of cell members) would make the omission impossible for the next cell read added. (Also, unrelated to this line: `test/poll.spec.ts`'s `liveCamera` fixture still uses model `T8410` captioned "battery camera" — now false; same stale-fixture cleanup this PR did in property-announcements/typed-reads but missed there.)

src/model/capabilities/battery.ts
```diff
@@ -121,25 +121,33 @@ function recordSetting(param: number, value: number, ctx: CommandContext): Comma
...
+ * (Indoor Cam Pan & Tilt) is confirmed mains-only by the maintainer, reported after a live unit showed
+ * a battery level, a cell temperature and both solar reads it cannot have.
  *
  * KNOWN, ACCEPTED trade: because the capability stays, `poweredOf` (camera.ts) still resolves these as
  * `battery`, so a live stream is budgeted as if cell-powered. An unnecessary power budget is cheap; a
  * phantom battery icon is a support ticket — so the visible entity is fixed here and the budget is
  * left as-is (a `poweredOf` refinement would be a separate change).
  */
-const MAINS_CAMERA_MODELS = ["T8425", "T8419"] as const;
+const MAINS_CAMERA_MODELS = ["T8425", "T8419", "T8410"] as const;
```
max246, line 150: Explicit prefix list, so the sentinel-read symptom persists for any mains-only camera model NOT enumerated here — each new one needs a manual edit + release. Known trade (not blocking), worth a line in the PR/issue so it's a conscious choice rather than a silent gap.

src/model/capabilities/battery.ts
martijnpoppen: Confirmed on all three points, and fixed in 46c12ac.

**The gate.** `available: notMainsCamera` is on `batteryPowerStats` now. Your diagnosis of why it was missed is the part I hadn't seen: I read `unexposed: true` as covering it. It doesn't — it suppresses the fluent getter, and `propertiesOf` filters `writeOnly` and `available` and deliberately *not* `unexposed`, because a schema entry reachable through `getProperty` is the whole point of that mark. So it was the same phantom cell read through the other door. That reasoning is now on the member itself, since the next reader will make the same assumption I did.

**The structural point, which is the real finding.** You're right that hand-adding the gate per member is what produced the bug — twice now, since the first pass covered 2 of 7 and this one hid behind a different flag. I went with your second suggestion rather than a `cell: true` tag: `CELL_PARAMS` names the params whose subject *is* the physical cell, once, beside the guard.

I made it a *checked* invariant rather than a centrally-applied one, and the reason is that `available` stays visible and greppable at each member, which is how the rest of this table reads. What makes it a gate rather than a reminder is that the spec holds the two against each other **both ways**:

- every param in `CELL_PARAMS` is published on a battery camera and withheld on a mains one — a cell read declared without the gate fails;
- every param withheld on a mains camera is in `CELL_PARAMS` — a gate on something that *isn't* a cell read fails too, since that would withhold a reading a mains camera genuinely answers.

Each direction confirmed to fail with only itself reverted:

```
AssertionError: 3100 is a cell read with no notMainsCamera gate: expected [ 1293, 1246, … ] to not include 3100
AssertionError: 1293 is gated but is not in CELL_PARAMS: expected [ 1246, 1249, 1250, 1251, 1103 ] to include 1293
```

One judgement call recorded on the list: `cameraInfo` (1103) is **not** in it. Its meaning is unevidenced, so gating it would assert it is a battery fact — the kind of claim this guard exists to stop making. `powerSource` and the `record*`/`workingMode` settings are out for the plainer reason that a mains camera genuinely answers them.

**The stale fixture — and two more of it.** `liveCamera` is a `T8114` now. Grepping for the class found the same thing twice in `manifest.spec.ts`, and one was load-bearing:

```ts
expect(battery.reads.map((r) => r.accessor)).toContain("level");   // on a T8410 record
```

That passed only because `ctxWith` built a context with no `model`, so `notMainsCamera` saw `""` and installed the read on everything. A fixture naming a model while the context it binds with knows none proves nothing about that model. `ctxWith` takes the model now, both records are `T8114`, and putting `T8410` back fails the assertion — which it could not do before.

`npm run verify` green: 188 files, 3540 tests.

---

src/model/capabilities/battery.ts
martijnpoppen: Agreed, and it is a conscious choice — it just wasn't written down where a reader would find it, which is a fair thing to call out.

It is already argued on `MAINS_CAMERA_MODELS` itself: a `device-family.ts` predicate (`isFloodLight`/`isIndoorCamera`) would over-reach, because not every floodlight or indoor cam is mains-only, and this guard must assert "mains" only for hardware actually checked. Asserting it by family would trade a phantom battery on three models for a *missing* battery on real battery cameras, which is the worse failure — a device that silently stops reporting its charge. `WORKING_MODE_DEFAULT_MODELS` in the same file takes the same shape for the same reason.

So the cost is exactly as you describe: a mains-only model not enumerated here keeps the sentinel reads until someone adds a prefix. I've added it to the PR body under "Unverified or deferred" rather than leaving it implicit, including that the three entries sit at different evidence bars (T8425 owned hardware, T8419 the app's own mains handling, T8410 maintainer-confirmed from a live unit).

Not treating it as blocking, per your note.

---

src/model/capabilities/battery.ts
```diff
@@ -121,30 +121,67 @@ function recordSetting(param: number, value: number, ctx: CommandContext): Comma
...
+ * read fails too. Neither list can drift from the other without CI saying so.
+ *
+ * What is NOT here matters as much. `powerSource` (1293) says which supply is in use, which a mains
+ * camera genuinely answers; `workingMode` and the three `record*` settings describe how hard the camera
+ * works, not what powers it; and `cameraInfo` (1103) is a number whose meaning is unevidenced — gating
+ * it would assert it is a battery fact, which is the kind of claim this guard exists to stop making.
+ * The two solar params ARE here: a panel exists to charge a cell, so a device without one has no solar
+ * harvest to report either.
+ */
+export const CELL_PARAMS: readonly number[] = [
```
lenoxys, line 191: "This read describes the cell" is now declared twice, as `available: notMainsCamera` on the member and again as a list entry here, with a spec holding the two equal. That is the table-per-concern shape the rule names, and max246's first suggestion, a flag on the member the gate is applied from, keeps it one declaration.

The doc also cites `battery.spec.ts`. Shipped source cites its peers only, and a spec never ships, so that pointer dangles in `dist/` the same way a `.md` link would. `guard:docrefs` only greps `.md`, so nothing catches it.

And the reason given for excluding 1293 does not hold: this doc calls `powerSource` the read that says which supply is in use, while `BATTERY_PARAM.POWER_CHARGE` calls it a charging flag and the member's own doc says `0 = Battery, 1 = External Solar Panel`. Neither is mains versus cell. On a cell-less camera that read, and its described write, is the same phantom this PR exists to remove. The exclusion may be right, but not for that reason.

src/model/capabilities/battery.ts
```diff
@@ -121,30 +121,67 @@ function recordSetting(param: number, value: number, ctx: CommandContext): Comma
...
+ * (Indoor Cam Pan & Tilt) is confirmed mains-only by the maintainer, reported after a live unit showed
+ * a battery level, a cell temperature and both solar reads it cannot have.
  *
  * KNOWN, ACCEPTED trade: because the capability stays, `poweredOf` (camera.ts) still resolves these as
  * `battery`, so a live stream is budgeted as if cell-powered. An unnecessary power budget is cheap; a
  * phantom battery icon is a support ticket — so the visible entity is fixed here and the budget is
  * left as-is (a `poweredOf` refinement would be a separate change).
  */
-const MAINS_CAMERA_MODELS = ["T8425", "T8419"] as const;
+const MAINS_CAMERA_MODELS = ["T8425", "T8419", "T8410"] as const;
```
lenoxys, line 157: The prefix list has precedent in this file, so the list is not my objection. The unstated part is that it fails open: a record with no `model` publishes all seven reads. That is the same hole the PR found in `ctxWith` and fixed only in the specs.

Nothing here handles a device whose supply can change either. For these three models that is moot, and worth one line saying so rather than leaving a reader to wonder.

src/model/capabilities/battery.ts
```diff
@@ -121,30 +121,67 @@ function recordSetting(param: number, value: number, ctx: CommandContext): Comma
...
+ * real cell. They keep the `battery` capability — it owns the working-mode and recording settings,
+ * which a mains camera genuinely has — but every read that describes a PHYSICAL CELL is withheld via
+ * {@link notMainsCamera}, so they don't sprout a bogus battery %, a cell temperature, a state of
+ * health or a solar harvest.
+ *
+ * The guard covers `level`, `charging`, `temperature`, `health`, `solarIntensity`, `solarConnected24h`
+ * and `batteryPowerStats`: a device with no cell has no charge, no cell temperature, no ageing figure,
+ * no power history and nothing for a panel to charge. The settings beside them are deliberately NOT
+ * gated, and that split is the whole reason the capability stays attached. Which reads are which is
+ * {@link CELL_PARAMS}, held against this table in both directions by the spec — the per-member gate is
```
lenoxys: Three JSDoc passages narrate the review: "the per-member gate is where this went wrong: the first pass covered two of the seven" here, "where this was first got wrong" on the model list, and "which is how this one came to be missed" on `batteryPowerStats`. The rule is explicit that a JSDoc does not narrate what a fix corrected, and there is no precedent for it in this file. All three need editing the next time a cell read lands, which is the test the rule states.

The `poll.spec.ts` fixture comment is the same thing in a spec.

src/model/capabilities/__tests__/battery.spec.ts
```diff
@@ -1,6 +1,7 @@
 import {
```
lenoxys, line 1: Two gaps in what is asserted.

The binder is untested for a mains model. Both new cases go through `propertiesOf`, which is the production projection, but `bindMembers` makes its own `available` decision and no case drives a mains camera through it. I checked by hand that it binds only `powerSource` and its setter, so the behaviour is right and unpinned.

And the claim is "gates every read that describes a physical cell", while `detection.evidenceParams: [1101]` and the three `batteryAlert` push events are ungated, so a mains camera still resolves as `battery` and can still emit a low or hot cell alert. Not a read, same claim.

Cut: `expect(cell.length).toBe(CELL_PARAMS.length)` plus the seven-name array is subsumed by the bidirectional check above it.

lenoxys (changes requested):
The bug is real and the direction is right, but the evidence table in the body does not describe what the code does. On this head a `T8410P00` record still answers `getProperty("battery")` with 88 and `getProperty("solarIntensity")` with 7, and `getProperties()` lists nine cell readings. `applyParams` names an unmatched param from the dictionary, so gating the member takes away the typed getter and leaves the value in live state under the same public name.

So a host still sees a phantom cell, which is the thing this PR is for. That also undoes the premise the seventh gate was justified on: `batteryPowerStats` is not stopped, it lands under `batteryPowerDatas`.

Second, `CELL_PARAMS` states the same fact the members table states, joined by param id, which is what the one-table rule is about.

src/model/capabilities/battery.ts
martijnpoppen: All three taken in 1b838db, and the first one is the correction that matters.

**The double declaration.** You're right, and so was @max246's first suggestion — I answered his ask with the weaker half. `cellGated` now walks the table as it is built and adds `notMainsCamera` to every member whose param is in `CELL_PARAMS`; no member states the gate at all. Applied, not checked, so there is one declaration and nothing for a spec to hold equal. Adding a cell read is adding its param to one list.

A member's own `available` is composed with rather than replaced. None declares one, and a future one would be about something else entirely — a family, a codec — so silently dropping it would leave a gate that reads as present and is not.

**The dangling citation.** Correct, and the reason it would have survived is the useful part: `guard:docrefs` only greps `.md`, so a `.ts` JSDoc pointing at a file that never ships has nothing checking it. Removed. Worth knowing that class exists beyond this one instance.

**1293 — you're right, and the exclusion needed a different argument.** `powerSource` is `0 = Battery, 1 = External Solar Panel`. Both values describe how a **cell** is fed, so "which supply is in use, which a mains camera genuinely answers" is simply wrong, and by this guard's own rule it arguably belongs in the list. On a cell-less camera it is the same phantom, and it carries a described write on top.

Two reasons that do hold, now written on the list instead:

1. Every param in `CELL_PARAMS` is one a **live mains camera was observed to publish** — the original symptom report named a level, a cell temperature, a state of health and both solar reads. 1293 was not among them.
2. It is `requires`-gated on its own param, so it appears only where the device reports it, unlike the sentinels which arrive whether or not they mean anything.

Neither is evidence that a mains camera handles 1293 sensibly — I have none either way — and gating it would withhold a described **write** rather than a read, which is a different class of change from the rest of this PR. So it is recorded as **unresolved** rather than decided, on the list and in the body. If you think it should go in, I'll add it; I did not want to widen the guard past what was measured on the strength of reasoning alone.

---

src/model/capabilities/battery.ts
martijnpoppen: Both stated in 1b838db, and the fail-open observation is the sharper one — it is the same hole this PR found in `ctxWith` and fixed only in the specs, which I did not notice was the *same* hole.

`notMainsCamera` matches a model prefix, so a record carrying no `model` matches nothing and publishes all seven cell reads. That is the direction I would choose — a real battery camera never silently loses its charge because its record came back thin — but it was inherited rather than chosen, and it means the list can only ever be a **floor**: a thin record gets the sentinels, exactly like a mains model nobody has enumerated. There is a spec for it now, pinning both `{}` and `{ model: "" }`, so the behaviour is a decision with a test rather than a property of how the predicate happens to be written.

The static-model point is in too. Nothing here reads what is *currently* powering the device — it is a fact about a model — so a unit whose supply can change is unhandled. Moot for these three, which have no cell to charge at all, and worth one line because the shape of the guard does not say so.

---

src/model/capabilities/battery.ts
martijnpoppen: All four deleted in 1b838db — the three JSDoc passages and the `poll.spec.ts` fixture comment.

The test you name is the one that settles it, and it is objective rather than a matter of taste: every one of those passages would have needed editing the next time a cell read landed, because each counted something ("two of the seven") or dated something ("where this was first got wrong"). A doc that has to be revised by an unrelated change is narrating the change that produced it.

What survives is the mechanism, which is the part a reader needs: `unexposed` suppresses the fluent getter and not the schema entry, because `propertiesOf` filters `writeOnly` and `available` and deliberately not `unexposed` — so a cell param needs `CELL_PARAMS` either way. That is a live constraint on the next member, where "which is how this one came to be missed" was only a story about this one.

You raised the same thing on #172 and it was the better catch there too. The two PRs had four and three of them respectively, which suggests it is a habit rather than a slip, so I've taken it as a general correction rather than seven local ones.

---

src/model/capabilities/__tests__/battery.spec.ts
martijnpoppen: Both gaps covered in 1b838db, and the cut is taken — with one addition that came out of making it.

**The binder.** You're right that it is the surface that matters: `propertiesOf` is one projection, and a caller holding `dev.battery()` goes through `bindMembers`, which makes its own `available` decision. It is also the surface the original symptom was reported from, so covering only the schema left exactly the thing the bug was seen on unpinned.

Pinned **whole** rather than as a list of absences, so a cell getter that starts installing shows up even if nobody thought to name it:

```ts
expect(bound("T8410P00")).toEqual(["powerSource", "setPowerSource", "setWorkingMode", "workingMode"]);
```

Which matches your hand-check exactly. (One wrinkle worth recording: the bound object is keyed by the table's **member** names, not property names — `level`, `temperature`, `health`, not `battery`, `batteryTemperature`, `batteryHealth`. My first attempt asserted the property names and passed vacuously for the absences.)

**The scope claim.** Also right, and the honest fix was to narrow the claim rather than widen the guard. `evidenceParams` still names 1101, so a mains camera reporting that sentinel still resolves as `battery` — and that is load-bearing, because detaching the capability would take the working-mode and recording settings with it, which these models genuinely have. The three `batteryAlert` events are the same shape: a station can still deliver a low- or hot-cell alert for one of them. Both pinned as deliberate, with the note that changing the push side needs evidence about what these models actually emit, which I do not have.

**The cut, and what it exposed.** `expect(cell.length).toBe(CELL_PARAMS.length)` and the seven-name array are gone — every assertion drives off `CELL_PARAMS` directly now, so the names are not restated.

That turned out to open a hole, which I would rather flag than let you find: with the gate applied *from* that list, deleting an entry deletes its own check too, so a deletion silently un-gates a read and stays green. I confirmed it — dropping 3100 passed. So the list's contents are pinned once, written out, because a test that takes its expectation from the thing under test proves nothing about it. Dropping 3100 now fails. That is the one place duplication earns its keep here, and it is in the spec rather than the source.

---

max246 (approved):

lenoxys (commented):
Six of the seven are properly gone now, and the one-declaration fix is the right shape: `cellGated` applies the gate from `CELL_PARAMS` as the table is built, so there is one list and the gate derives from it.

The headline read still answers, though. On `7b6a1c5`, a `T8410P00` record reporting 1101/1102/1105:

```
caps:            video, snapshot, motion, person_detection, camera, battery, info
getProperties(): battery, sdinfo, cameraInfo, unknown_1104, unknown_1105, unknown_1106, unknown_1107
getProperty("battery")        -> { name: "battery", paramType: 1101, value: 88 }
getProperty("batteryTemp")    -> undefined
getProperty("solarIntensity") -> undefined
getProperty("batteryStatus")  -> undefined
```

So the cell temperature, the solar reads and the status are honest raw ids now, and the sentinel this PR is named for is still served under a public name. The route is the param dictionary rather than the members table: 1101 is named `battery` there, so gating the member removes the typed getter and `applyParams` names the value anyway. 1105 has no dictionary entry, which is why it degraded correctly and this one did not.

Two ways out, and I do not mind which: skip a dictionary name for a param no installed member claims, or stop resolving `battery` for a mains model at all, which also settles `detection.evidenceParams: [1101]` and the three `batteryAlert` events still being live on a camera with no cell.

The evidence table in the body needs the same correction either way: `After | none` is true for six of seven.

lenoxys (commented):
Ran this against a live account, whole fleet, both builds. Fourteen devices, `beta-0.2.0` versus `7b6a1c5`, comparing for each one: which of the eight cell params it actually sends, which cell properties end up public, what `getProperty` answers, and what `dev.battery()` binds.

**The two builds are byte-identical on all fourteen.**

Good news first: no regression. Every real battery device keeps everything. Six models report 1101 and still publish it, three of them with `solarIntensity`, `temperature` and `health` bound exactly as before.

The part that matters for this PR: **not one device reports 1101 on a mains model.** The only unit in the `MAINS_CAMERA_MODELS` list is a T8410, and what it sends is `1103=1` and nothing else. No `battery` capability is resolved for it on either build, no cell property is public, `dev.battery()` is unbound. The evidence gate already withheld all seven reads, because the sentinel this PR is built around never arrives.

So the symptom is not reproducible on this hardware, and the model list has no observation behind it. Three possibilities, and the PR does not say which:

- it is another unit, a T8425 or T8419, neither of which is on this account;
- 1101 arrives on a path other than the cloud record, a P2P notification mid-session, in which case the gate does real work and the test for it is a notification rather than a record;
- it was inferred rather than observed, and the list is defensive.

Any of the three is fine as a claim. "Ground every wire claim in evidence you can point at" asks which one it is.

One more thing the sweep turned up, on `CELL_PARAMS` rather than on the gate: a HomeBase reports `1103 = -43`. That reads like dBm, not a cell health grade, and 1103 is in the list as `devBatteryHealthyV2`. Nothing publishes it today so nothing is wrong downstream, but the list's own doc says every entry describes a physical cell, and one of them is answering -43 on a mains-powered station.

My blocking finding stands and is now precisely bounded: `getProperty("battery")` still answers under a public name when 1101 is present, which no device here does, and which is exactly the device this PR exists for. Nothing on this fleet can tell you whether that path is hit in the field.

Probe kept as `scratch/an-170-fleet.mjs` so it can be rerun per build. No serial, name or token in the output.

martijnpoppen:
@lenoxys — you're right, and this was the finding. Fixed in 8d0c494.

First, an apology for the delay in answering this one: I replied to your four inline threads this morning and never read the review body they came with, so the most serious point sat unanswered for five hours while I reported the PR as green. Reading threads and not reviews is its own version of reading metadata and not threads.

**Reproduced before acting on it**, on the head you reviewed:

```
getProperty("battery")        = 88      ← a camera with no cell
getProperty("solarIntensity") = 7
getProperties() → battery, batteryTemp, devBatteryHealthyV2,
                  solarIntensity, solarPanelConnect24h, batteryPowerDatas
```

Six cell readings on a gated T8410. My evidence table was measured with `propertiesOf` — the schema projection — not the surface a host reads, so it described something true about a projection and nothing about the device.

**The mechanism, which is worse than "some names leak."** `applyParams` names by precedence: curated `PropertySpec` → param dictionary → `unknown_<paramType>`. Gating a member removes its spec, so the value falls to the **dictionary** — and for 1101 and 1309 the dictionary spells it exactly as the member does. So for the headline battery percentage and the solar intensity, the gate changed *nothing a caller could see*. The others were renamed rather than withheld. `batteryPowerStats` landing as `batteryPowerDatas` is the same thing, and you're right that it undoes the premise the seventh gate was argued on.

The gate was demoting a reading, not removing it.

**The fix is in `device.ts`, not `battery.ts`**, because the hole is in the naming precedence rather than in the guard. Any `available`-gated member on any capability behaved this way; fixing it in the battery module would have left it in place for the next one. `withheldParams` is the ungated merge minus the resolved one — no new declaration, and it follows any gate rather than this one:

```ts
const named = spec ? spec.name : def && !this.withheldParams.has(pt) ? def.name : undefined;
```

The value is still kept, as `unknown_<paramType>`. The graceful-unknown rule says never drop data, and `unknown_` is the one form that carries a value without claiming it means anything; `inspectParams` still reaches it. A mains camera now serves no cell reading under **any** public name, while `powerSource` and the recording settings stay. A real battery camera is untouched.

Pinned by a spec covering both spellings, confirmed to fail with only the skip reverted:

```
AssertionError: a mains camera served battery: expected { name: 'battery', … } to be undefined
```

---

**On `CELL_PARAMS` — I've thought about it and I'd like to push back, rather than move a third time.**

`cellGated` already applies the gate from that list, so the fact is stated once; what's left is the join by param id. The alternative that removes the join is @max246's original first suggestion, a tag on the member — and `Members` is `Record<string, Member>` over a closed `ValueMember` interface, so that means adding a battery-domain flag (`cell?: true`) to the type **every** capability module is written against. That seems a worse trade than a seven-entry list sitting ten lines from the only function that reads it, both private to the module.

I may be weighing your rule wrong, though — it's your rule. If you'd rather have the tag on the member, say so and I'll add it to `ValueMember` with a doc explaining why a domain concept lives on the generic type.

`npm run verify` green: 189 files, 3535 tests, 91 snippets. Docs build green.

---

martijnpoppen:
@lenoxys — you wrote this fix an hour before I did and I didn't see it. `bbf8028` at **13:31 UTC**, my `8d0c494` at **14:29 UTC**. You found the bug reviewing this PR at 07:04, went and fixed it, and I then pushed a second implementation of the same change onto the PR you'd found it in. Mine is reverted in `da0c8a2`; `device.ts` is yours.

Not just deferring to who was first — **yours is measurably more correct.** `withheldParams` built the set from `mergeProperties(resolved.capabilities)`, which de-duplicates **by name, first-wins**. A param claimed under a name an earlier module already claimed never reached the set, so the dictionary could still name it. That collision exists in the tree today:

```
battery  → battery:1101, lock:1101, vacuum_clean:163, display:8001
charging → battery:2111, keypad:1102
rssi     → contact:1141, lock:1141, keypad:1141, vacuum_clean:134
```

163 and 8001 are both `battery` in their own dictionaries. On a device resolving `battery` and `display`, `MODULES` order keeps 1101 and drops 8001 — so a gated display battery read leaks back under `battery` and the gate changes nothing a caller can see, which is the exact shape of the bug you reported. `claimedParams` collects per module, before the dedup, and has no such hole. Mine was silently sensitive to the order of the `MODULES` list. That's a latent bug, not a preference, and I'd have shipped it.

**This PR now carries only the half that needs device evidence** — T8410 in the model list, and `cellGated` over the seven cell params. #175 is merged into the branch so CI runs against the behaviour these specs assert; those commits leave the diff when #175 lands on `beta-0.2.0`.

The specs from the reverted commit stay: `getProperty`/`getProperties` under both the member's and the dictionary's spelling is the regression cover for what this PR claims, and it's worth having however the precedence is fixed. Verified green on your implementation with mine removed before removing it — 190 files, 3538 tests — then `npm run verify` on the merged branch: 190 files, 3537 tests, 91 snippets. Docs build green.

One difference worth a decision, which I've raised on #175 rather than argued here: 3100 carries `encoding: "json"`, and skipping `paramDef` outright means a withheld encoded param lands as the raw string where mine kept the decoded object. I think yours is right — every other `unknown_` is raw — but it's a visible change to `unknown_3100` and better named than discovered.

`CELL_PARAMS`' param-id join and `powerSource` (1293) are still open above, unchanged and still yours to call.

---

lenoxys (approved):
Lgtm

PR #171, opened 2026-09-13
docs(gallery): show the Smart Display, on a line of its own

martijnpoppen:
## What changed

The T87A0 product render, and the gallery figure pointing at it. `docs/devices-gallery.md` gains a top-level `## 🖥️ Display` group, and the page intro — which listed five lines — names the sixth.

## Why

#165 gave the Smart Display its own product line, codec and param namespace. The gallery had no render for it at all: 144 figures, and the only `T87*` entries are `T87B0` and `T87B2`, which are the SmartTrack Link and SmartTrack Card. So the one device on the newest line was the one device with no tile.

Downstream it showed up as a consumer drawing the display with a generic glyph while every other device class had a photograph — which is what prompted this.

## Evidence

n/a for artwork. What is checkable is that it builds and resolves, and that is checked below rather than eyeballed.

The render itself: 1200×1200, real transparency, trimming to 1118×747. That is the isolated-product shape the gallery is made of, not a product cut out of a scene — so it composites onto any ground without carrying one with it.

## Verified how

- [x] `npm run verify` green — 188 files, 3539 tests, all guards, `check:esm`, examples
- [x] Ran against a real device — n/a, this adds no wire and no code
- [ ] Added or updated specs — none; there is nothing here a spec can assert that the docs build does not
- [x] Docs updated — this is the docs change

**The docs build is the real gate for this one**, and it is outside `verify` by design, so it was run explicitly: `npm -C docs ci && npm -C docs run build` (TypeDoc → `guard:docs` → VitePress) completes, `dist/devices-gallery.html` references `security-T87A0.webp`, and the file is present at `dist/devices/security/security-T87A0.webp`. `guard:docrefs` passes.

### A top-level group, not a Security subsection

The gallery's top-level groups are product LINES — Security, Clean, Life — and #165 made `display` a sixth. Filing this tile under Security would undo in the gallery exactly what that PR established in the model: it connects over secure MQTT with no `p2p_did` and speaks no P2P at all, which is the whole reason the line was split out.

The group carries a `###` subsection like every other, and that is not cosmetic. The gallery's filter script collapses an empty grid by walking back to the nearest heading and hides it only when that heading is an `H3`. A group without one would leave a stray heading behind whenever a search matched nothing inside it.

## Unverified or deferred

- **The image PATH and the grouping now disagree.** The render is committed at `devices/security/security-T87A0.webp` while the figure sits in the Display group. The directories are per-family and this one is arguably `display/`. Left as committed rather than re-filing the asset in the same change that introduces it; nothing reads the directory for meaning, and moving it later costs one `git mv` plus one `src` attribute.
- **No render for any other T87Ax.** The registry has only `T87A0` observed so far, so the group has one tile. `classify.ts` routes the whole `T87A` prefix to the display codec, so a sibling model would resolve correctly and simply have no picture until one is added.
- **Nothing here is generated.** `scripts/device-artwork-to-webp.sh` converts renders that are dropped in by hand and rewrites references; it does not add figures. So a future render needs its own `<figure>` the same way this one did.

---
_Generated by [Claude Code](https://claude.ai/code/session_01W5fkFeirrJA4u6gwYZ5xq3)_


docs/devices-gallery.md
```diff
@@ -683,6 +683,25 @@ The eufy **Clean** line — robot **vacuums** (the classic **RoboVac** and newer
...
+The eufy **Display** line — a `display` codec and product line of its own, not a security device: it
+connects over secure MQTT with no `p2p_did` and speaks no P2P at all. The `display` capability reads the
+screen's charge (param 8001); its retail name, model code and a version-shaped string are readable by
+name without a typed getter. Read-only — no display write is captured. See the
+[Smart Display guide](/smart-display).
+
+<div class="device-gallery">
+
+<figure>
+  <img src="/devices/security/security-T87A0.webp" alt="Smart Display E10 T87A0" loading="lazy" />
```
max246, line 699: Nit (not blocking): Display-line device but the image lives at devices/security/security-T87A0.webp (security/ folder, security- prefix). clean/life/mower each have their own folder; devices/display/display-T87A0.webp would match the convention + the "line of its own" framing. Rest checks out: /smart-display guide exists, param 8001 = charge is accurate.

max246 (approved):

docs/devices-gallery.md
martijnpoppen: Agreed — and this PR's own "Unverified or deferred" section flagged the same thing, so you and it reach the same conclusion independently:

> **The image PATH and the grouping now disagree.** The render is committed at `devices/security/security-T87A0.webp` while the figure sits in the Display group. The directories are per-family and this one is arguably `display/`. Left as committed rather than re-filing the asset in the same change that introduces it; nothing reads the directory for meaning, and moving it later costs one `git mv` plus one `src` attribute.

You're right that "a line of its own" is exactly the framing that makes the `security/` folder and the `security-` prefix read wrong, so the deferral is weaker than I made it sound.

It merged before this landed, so it needs a follow-up rather than a push here. Not opening one unprompted — it is two lines against a merged PR — but it is recorded and I've raised it with the maintainer. Worth noting it is three moves, not two: the Homey app carries its own copy of the render (the SDK package ships `dist` only, so a consumer's generator cannot read `docs/`), and its `DEVICE_PHOTOS` entry names the file, so that side wants renaming in the same breath or the two repos disagree about what the file is called.

Thanks for checking the other two claims rather than taking them — the `/smart-display` guide and param 8001 are the parts a reader would be misled by if they were wrong.

---

PR #172, opened 2026-09-13
feat(arming): every guard mode the station reports, it also accepts

martijnpoppen:
> **This PR changed shape.** It opened as `qualifyMode` — an instrument for sending the five guard modes this SDK refused, to find out whether a station takes them. It was used, all five came back confirmed, and the instrument is gone again. The net diff is the **result**: nine settable modes. The commits keep the record of how they were qualified.

## ⚠️ Breaking

**`ArmingMode` widens from four members to nine.** Additive for a caller *passing* a value; a compile error on upgrade for one keying `Record<ArmingMode, …>` exhaustively, or switching over it with no default. Filed here rather than under "what changed", because a reader of the diff will not see it.

## What changed

`ArmingMode` is all nine modes. `schedule` 2, `custom2` 4, `custom3` 5, `off` 6 and `geo` 47 join the four that were already settable, so `mode` stops being the one member whose write domain is narrower than its read.

Removed with it: `qualifyMode`, `UNQUALIFIED_MODES`, and the `armingFrame`/`armingCommand` split that existed only to share a frame between the probe and the setter. The instrument's domain was the *complement* of `ArmingMode`, so confirming the five emptied it — what was left was a member that could only throw.

`SETTABLE_MODES` and the member's `args` entry are gone too (@lenoxys). With the two sides agreed, an `args[0].values` beside an `enumValues` was a second declaration of the same set; `writeDomain` falls back to `enumValues`, so enforcement and the generated refusal are byte-identical without it. `ARMING_MODE_WIRE` is now typed `Record<ArmingMode, number>`, so **tsc** enforces that the table's keys are exactly the union — the property two of this PR's assertions had been standing in for.

`AlarmDelayMode` is untouched and its doc now says so explicitly.

## Why

The refusal was right while it stood, and this is what it was waiting for rather than a reversal of it. "Unverified write wires throw, never guess" also says to promote a mode *as it is captured* — and nothing had captured one. The app names all nine, but a name is not a wire, and on a fire-and-forget write a mode the station drops looks exactly like one it honours. That left five modes permanently unknown, because the only thing that would settle one is sending it to real hardware and watching.

What was missing was never a different command. The frame was already proven on this hardware and only `mode_type` differed. What was missing was a **readback** to tell "accepted" from "silently dropped".

## Evidence

`armingModeChanged` is that readback, and it is why this is evidence rather than an inference from one: `commandSinkFor` in `eufy-mega.ts` emits the semantic event **only after** `refreshEventState` has converged on the `expected` value — so a station that ignored the write produces silence, not a false positive. I checked that in the client rather than assuming it.

Each of the five was sent individually and each brought MODE_SWITCH back. Recorded in `ARMING_MODE_WIRE` at the bar `custom1` set — **live confirmation, not byte-capture**, dated 2026-09-13, and labelled as such beside the three that are byte-exact:

- ✅ wire-captured (byte-exact, T8030 2026-07-23, re-confirmed live 2026-08-05): `away` 0, `home` 1, `disarmed` 63
- ✅ confirmed live (T8030 2026-09-12): `custom1` 3
- ✅ confirmed live (2026-09-13): `schedule` 2, `custom2` 4, `custom3` 5, `off` 6, `geo` 47

## Verified how

- [x] `npm run verify` is green locally — 188 files, 3538 tests
- [x] **The docs build too** (`npm -C docs ci && npm -C docs run build`). Not part of `verify` by design, which is how the first revision of this PR went red on a TypeDoc dangling-link warning; run explicitly since.
- [x] **Ran against a real device** — a T8030, by the maintainer. That is the whole basis of this PR: each of the five sent and each observed to report its own MODE_SWITCH within the convergence window.
- [x] Added or updated specs. The `it.each` block that asserted the five are *refused* now asserts they are *sent*, checking the built frame against the byte-captured one (only `mode_type` differs) through both the fluent setter and the intent path.
- [x] Docs updated — **six** places claimed three or four settable modes, not the four this body first said: `ARMING_CMD.SET_ARMING`, the `mode` member's JSDoc, its published `description`, the `ARMING` module doc, and — both of which **ship**, both found by @lenoxys — `members.ts`'s `writeDomain` doc (wrong on the count even before this PR) and `src/model/capabilities/README.md`'s walkthrough, which used this member as its only example of a narrowed write.

## Unverified or deferred

**A refused mode is now quiet, where it used to be a compile error.** `setMode` resolves at `acknowledge()`, before convergence. A convergence failure does not reject — it emits **`commandUnconfirmed`** (`sn`, `property`, `param`, `expected`, `observed`, `timeoutMs`). So on a station that refuses `geo`, a caller who simply awaits `setMode` sees it resolve; the information exists only for a caller subscribed to that event. That is the real price of widening a union that guards a fire-and-forget wire, and it is the trade this PR makes: the compile error was the cheapest possible failure, and the evidence that removed it is real.

**The modes were confirmed on an account where they are configured.** `schedule`, `custom2`, `custom3` and `geo` exist only once set up in the app. The confirmations establish that the frame carries the integer and that the station converges on it — not what an **unconfigured** station does with the same frame, which nothing here exercises. That is also precisely where the gap above is most likely to be felt. Not treated as a reason to hold the promotion: the alternative is refusing all five on every station, configured or not, which is worse for the confirmed case and no better for the other.

**`AlarmDelayMode` stays narrower, deliberately.** Cmd 1255 carries its own `mode_id` on its own wire, with no capture beyond the byte-captured three and no known GET to read one back. Evidence for 1224 is not evidence for it.

**Not byte-captured.** Six of the nine writes are live confirmations rather than frame captures. The frame is the captured one with a different integer, which is the strongest claim available without a fresh capture per mode, and it is labelled that way rather than rounded up to "wire-captured".

**No member narrows its write now**, so neither `members.ts` nor the README can point at a live example of the feature. Both state the rule and the shape to recognise, and say outright that there is no current instance — rather than inventing a hypothetical member, which would be a second thing to keep true.


martijnpoppen:
Reopened, because the reason this went red is fixed and the branch now carries something different from what it opened as. If it was closed deliberately rather than because of the red check, say so and I'll close it again.

**Why the `docs` check failed.** Not the library — TypeDoc, on the first commit:

```
[warning] The comment for UNQUALIFIED_MODES links to "SETTABLE_MODES" which was
          resolved but is not included in the documentation.
[warning] Found 0 errors and 1 warnings
##[error]Process completed with exit code 4
```

I had exported `UNQUALIFIED_MODES` with a `{@link SETTABLE_MODES}` in its doc, and `SETTABLE_MODES` is private. Warnings are errors in that build, so one dangling link failed it. `UNQUALIFIED_MODES` no longer exists — confirming the five modes emptied its domain, so it and `qualifyMode` came out in the second commit — and the docs build now passes on this head.

That gap is worth naming: **`npm run verify` does not cover the docs build**, deliberately (`guard:docs` needs the docs deps). So a green `verify` is not a green PR, and I was reporting `verify` as though it were. I ran `npm -C docs ci && npm -C docs run build` against this head directly rather than assuming the removal fixed it.

**What the PR is now.** It opened as `qualifyMode`, an instrument for sending the five refused guard modes to find out whether a station accepts them. It was used, all five came back confirmed, and the instrument is gone. The net diff is the result: nine settable modes. The body is rewritten to match; the commits keep the record of how the five were qualified.

Also merged `beta-0.2.0` in, since #171 landed while this was closed. On the merged tree: `verify` green (188 files, 3539 tests) and the docs build green.

One thing in the body that wants a reviewer's opinion rather than mine — `manifest.spec.ts`'s narrowing test lost its only example, because after checking every module, arming was the SDK's only narrowed member.

---

src/model/capabilities/arming.ts
```diff
@@ -152,8 +165,8 @@ const ARMING_MODE_LABELS: Record<number, string> = enumLabels(ARMING_MODE_WIRE);
  * labels come from, so the offered set cannot name a mode the write would build differently.
  *
  * Published as the `mode` member's argument `values`, which is what makes it the domain the check and the
- * generated refusal both use: a caller is offered four and held to four, while the read still reports
- * all nine.
+ * generated refusal both use: a caller is offered nine and held to nine. Still DERIVED rather than assumed
```
lenoxys: `members.ts` is normative on this: the only reason to state an argument's `values` beside an `enumValues` that already publishes the read's set is that the two differ. They agree now, and the README says a member whose two sides agree declares `enumValues` alone.

I deleted `SETTABLE_MODES` and the `args` entry to check: enforcement and the refusal message are byte identical, because `writeDomain` falls back to `enumValues`, and the only failures are the two assertions this PR added to pin the agreement. Keeping the list so that narrowing again is one edit is flexibility bought with a second declaration, a const and two tests.

Taking the intersection off `ARMING_MODE_WIRE` closes the same gap better: type it `Record<ArmingMode, number>` and tsc enforces that the table's keys are exactly the union, which is what the new `args[0].values === Object.keys(enumValues)` assertion is standing in for.

src/model/capabilities/arming.ts
```diff
@@ -108,36 +118,39 @@ export const ARMING_CMD = {
  * model-dependent) — the single source of truth {@link ARMING_MODE_LABELS} (the `armingMode`
  * PropertySpec's `enumValues`) derives from.
  *
- * The READ side is why all nine are here: a station set to a schedule reports 2, and a getter answering a
- * number nothing can name is worse than one naming a mode we cannot set. The WRITE side takes only the
- * four of {@link ArmingMode} — {@link SETTABLE_MODES} is that subset, derived from this table rather than
- * listed again. {@link AlarmDelayMode}, the narrower domain cmd 1255 takes, is the byte-captured three.
+ * Read and write are the same nine as of the confirmations below, so this table is the whole domain in
+ * both directions — {@link SETTABLE_MODES} and {@link ARMING_MODE_LABELS} are both derived from it rather
```
lenoxys, line 120: Two more places still claim three or four settable modes, and the body says four sites were updated where six needed it.

`members.ts`'s `writeDomain` doc says arming reports eight guard modes and can set only the three whose wire was captured, which was already wrong on the count before this PR, and it ships. The README walkthrough at 283-286 uses this member as its only example of a narrowed write, so if the `args` entry goes, that example needs replacing rather than editing.

Also: `ArmingMode` is a published type and widens from four to nine. Additive for a caller passing a value, breaking for one keying an exhaustive `Record<ArmingMode, ...>`. That belongs under a breaking heading, not under what changed.

src/model/capabilities/arming.ts
```diff
@@ -183,13 +196,22 @@ function armingModeOf(v: boolean | number | string): ArmingMode | undefined {
...
 }
 
+/**
+ * How long a mode write is given to show up in a readback before it counts as unconfirmed.
+ *
+ * Generous because the readback is a bounded cloud-list refresh rather than a local echo, and a station
+ * between P2P sessions can take most of it. It is also the window each of the nine modes was confirmed
+ * within, so shortening it would make a mode that works report as dead.
+ */
+const MODE_CONVERGENCE_MS = 20_000;
```
lenoxys: A named const with eight lines of doc, one use, no behaviour change, declared two hundred lines from where it is read, and not mentioned in the body. The doc also overstates it as the window each of the nine modes was confirmed within, where the recorded timings are around five seconds. Put `20_000` back.

Same file, smaller: `armingCommand`'s new `payload` local is churn, and four JSDoc passages narrate what the code used to be ("this union was four for as long as five modes had no confirmation", "these five spent a release refused", "which is what mattered while it WAS narrower"). Deleting the `NOT settable` body comments was the right direction; the prose went the other way.

src/model/capabilities/__tests__/arming.spec.ts
```diff
@@ -21,6 +21,7 @@ const ctx: CommandContext = {
   paramIds: new Set(),
   accountName: "someone+tag",
 };
+
```
lenoxys, line 24: Two costs of the promotion that the deferred list does not name.

`setMode` resolves at `acknowledge()`, before convergence, and a convergence failure goes to `reportUnacknowledged`. So on a model that refuses `geo`, a caller now gets silence where it used to get a compile error. That is the real price of widening the union and it should be stated.

And `schedule`, `custom2`, `custom3` and `geo` exist only once configured in the app. Evidence from an account where they were configured says nothing about a station where they are not, and nothing here exercises that axis.

lenoxys (commented):
The evidence holds, and I checked the mechanism rather than the claim. `refreshEventState` compares the refreshed param to `expected` by value and `emitSemantic` only fires after that returns true, so a mode arriving is evidence of that value read back rather than of any change at all. A dropped mode raises `StateConvergenceError` instead of a false pass. Cloud-list readback on one T8030 is not a byte capture, and you say so.

Three things before merge, and two of them are about the narrowing declaration that now narrows nothing.

src/model/capabilities/arming.ts
martijnpoppen: Taken, and the reasoning is right — `members.ts` says an `args[0].values` beside an `enumValues` means the two DIFFER, and they don't any more. Deleted in 64be836, along with `SETTABLE_MODES`.

I reproduced your check before removing it rather than taking the "byte identical" on trust. `writeDomain` falls back to `enumValues`, and the generated refusal is unchanged to the character:

```
mode: "not-a-mode" is not a valid value (must be one of 0/1/2/3/4/5/6/47/63)
```

"Flexibility bought with a second declaration, a const and two tests" is the fair accounting. Narrowing again is now one edit to `ArmingMode` plus one `args` entry, which is the same edit it always was.

And your replacement for what those tests were pinning is better than the tests. `ARMING_MODE_WIRE` is `Record<ArmingMode, number>` now, so tsc enforces that the table's keys are exactly the union — a compile error where the assertion was a runtime one, and one that cannot be left out of a future edit. The remaining spec line just records that no `args` key exists, so re-adding one is a deliberate act rather than a drift.

One knock-on worth naming, since it changes a published shape: without a stated `args`, the derived argument carries no `values` at all — `describeWrite` deliberately omits `enumValues` there so a second copy cannot drift from the read's. `manifest.spec.ts`'s assertion moved to hold the indirection instead: the action names the read it reflects, and the domain is published on that read, once.

---

src/model/capabilities/arming.ts
martijnpoppen: Both found and fixed in 64be836, and you're right that the body undercounted — four claimed, six needed it.

**`members.ts`'s `writeDomain` doc.** Worse than stale: "reports eight guard modes and can set only the three whose wire was captured" was wrong on *both* counts before this PR touched it (nine and four), and it ships. It states the rule now, with no live example, plus the sentence that keeps it honest — nothing declares a narrowed write today.

**The README walkthrough.** Same treatment, and your conditional was the right way round: the `args` entry went, so the example had to be replaced rather than edited. Since no member narrows its write any more, neither doc can point at a real one; both give the shape to recognise (a device reporting a value it will not accept back) and say outright there is no current instance. That seemed better than inventing a hypothetical member, which would be a second thing to keep true.

**The breaking half — you're right and I had it filed wrong.** Widening a published union is additive only for a caller *passing* a value. For one keying `Record<ArmingMode, …>` exhaustively, or switching with no default, it is a compile error on upgrade. That is a breaking change to a published type and does not belong under "what changed". I've moved it into the body under its own heading rather than leaving it implied by the diff — flagging it here too because it is the kind of thing a reader of the changelog needs and a reader of the diff will not see.

---

src/model/capabilities/arming.ts
martijnpoppen: All three taken in 64be836.

`20_000` is back at its single use. Everything you list about the const was true — one use, two hundred lines away, eight lines of doc, no behaviour change, unmentioned in the body — and the doc was worse than redundant: "the window each of the nine modes was confirmed within" reads as a measurement when the recorded timings are around five seconds. It was a budget dressed up as evidence. The `payload` local is gone too.

The prose is the part I'd defend least. You're right that it goes the other way from deleting the `NOT settable` body comments, and the rule's own test is the one that settles it: all three passages would have needed editing the next time a mode landed, which is the definition of a JSDoc narrating a fix rather than stating what is true. Replaced with the rule — a mode belongs in the union once its write is confirmed, and a live confirmation establishes that *this* command carries that integer rather than that another command exists. That last clause is the only thing from the narration worth keeping, because it is a live constraint on how the next mode gets promoted, not history.

---

src/model/capabilities/__tests__/arming.spec.ts
martijnpoppen: Both are real costs and both belong in the deferred list. Added to the body.

**The acknowledge/convergence gap.** Confirmed in `commandSinkFor` and `reportUnacknowledged`, with one refinement that makes it slightly better than silence and worth stating exactly: a convergence failure emits **`commandUnconfirmed`** (`sn`, `property`, `param`, `expected`, `observed`, `timeoutMs`) rather than rejecting. So the information exists — but only for a caller subscribed to that event. A caller who just awaits `setMode` sees it resolve, and resolve is what it used to be unable to do at all for `geo`.

That is the honest price of widening a union guarding a fire-and-forget wire, and it is the same trade in both directions: the compile error was the cheapest possible failure, and the evidence that removed it is real, so what replaces it is an event a caller has to opt into. Stated plainly rather than left for someone to discover on a model that refuses one.

**The configuration axis is the better catch**, because it is a gap in the evidence rather than in the write. `schedule`, `custom2`, `custom3` and `geo` only exist once configured in the app, and every confirmation here came from one account where they were. What that establishes is that the frame carries the integer and the station converges on it — not what an unconfigured station does with the same frame, which nothing here exercises and I have no way to exercise without a second account. Whether it refuses, converges anyway, or converges to something else is unknown.

I'm not treating that as a reason to hold the promotion: the alternative is the status quo where all five are refused on every station, configured or not, which is strictly worse for the configured case and no better for the unconfigured one. But it means the unconfigured station is precisely where the `commandUnconfirmed` gap above is most likely to be hit, and the two deferred notes now say so together rather than separately.

---

lenoxys (approved):

PR #175, opened 2026-09-14
fix(model): a read a device's gate withheld is not published under its name

lenoxys:
## Why

The param dictionary names ids for a whole namespace. A capability names the same ids for the devices that carry the read. Both call 1101 `battery`, and they disagree only about which device it describes.

So a gate that withholds a read only closes one of two doors. #128 withheld `level` and `charging` from a mains camera that reports 1101 as a sentinel, and the dictionary went on publishing that sentinel under the same name:

```
T8425, real device record with 1101 = 100 injected, beta-0.2.0 as merged today:
  dev.battery()?.level        -> undefined      (the gate works)
  getProperty("battery")      -> 100            (the gate does not)
```

A caller reading the second cannot tell it from a charge a device really answered. That is the whole point of the gate, and it has been half-defeated since #128 merged.

## What

A param a resolved capability declares, that this device's schema does not carry, takes the `unknown_<id>` passthrough rather than its dictionary name.

Two cases deliberately unchanged:

- **A capability that never resolved withholds nothing.** A reading can arrive before the capability claiming it, so it keeps its dictionary name and a later record widens onto it. That is why the rule asks about resolved capabilities rather than all of them.
- **A param the dictionary owns alone** is untouched. Nothing decided against it, and the loose read is documented to answer params no capability surfaces.

The withheld set is a subtraction the device already had both halves of: every param its resolved capabilities declare, minus the schema it ended up with. Nothing new is threaded through resolution.

## Verified

`npm run verify` green: 190 files, 3545 tests. The new spec fails on `beta-0.2.0` without the change, which is the point of it: `getProperty("battery")` answers the sentinel on a T8425 today.

Against a live account, whole fleet, before and after:

- **No regression.** All fourteen devices identical, including four that report 1103 and keep `cameraInfo`, since no gate refused it.
- **The fix demonstrated on a real record.** A fleet device's own cloud record with 1101 synthesized into it, resolved as a T8425: `getProperty("battery")` goes from 100 to undefined and the value appears as `unknown_1101`. Same record resolved as its real model is unchanged, because nothing gates that model here.

Nothing was written to any device. The two probes are kept in `scratch/`.

## Relation to #170

They compose, and neither replaces the other. This closes the second door for any gate. #170 decides which reads describe a cell and which models are mains, which is the part that needs device evidence.

On my own account the point is moot in both directions: no device reports 1101 on a gated model, so nothing here would notice either change. The synthesized record above is as close as this fleet gets.


martijnpoppen:
This is the same fix I pushed to #170 as `8d0c494` — an hour after yours, and without having seen it. **Mine is reverted** (`da0c8a2`); `device.ts` is this PR's. #170 keeps only the half that needs device evidence, and has #175 merged in so its specs run against the real behaviour.

I compared the two implementations properly rather than just yielding, and **yours is the better one for a reason worth recording.** Mine built the claimed set from `mergeProperties(resolved.capabilities)`, which de-duplicates **by name, first-wins** — so a param claimed under a name an earlier module already claimed never reached the set. That collision is in the tree today:

```
battery  → battery:1101, lock:1101, vacuum_clean:163, display:8001
charging → battery:2111, keypad:1102
rssi     → contact:1141, lock:1141, keypad:1141, vacuum_clean:134
```

163 and 8001 are both `battery` in their own dictionaries. On a device resolving `battery` and `display`, `MODULES` order keeps 1101 and drops 8001, so a gated display battery read comes straight back under `battery` — the exact shape of the bug this PR fixes, surviving the fix. `claimedParams` collects per module, before the dedup, so it has no such hole. Mine was silently coupled to the order of the `MODULES` list.

Verified #170's specs against this branch with mine removed: **190 files, 3538 tests, all green**, including the `getProperty`/`getProperties` pins under both the member's and the dictionary's spelling. It is a drop-in for what I'd written.

---

**One behavioural difference worth naming, not blocking.** Setting `def = undefined` skips `paramDef` entirely, so `def?.encoding` is skipped too. 3100 (`batteryPowerDatas`) carries `encoding: "json"`, and it is one of the seven params #170 gates. Measured on a mains T8425 reporting 1101 and 3100:

```
this PR:  unknown_3100 = "{\"a\":1}"   (string — raw passthrough)
mine:     unknown_3100 = {"a":1}       (object — still decoded)
```

I lean **yours**: every other `unknown_` in the map is `String(rawVal)`, so keeping one decoded because the param happened to have a dictionary entry makes the shape of an `unknown_` value depend on something a caller can't see. The counter-argument, for the record, is that `encoding` is a wire fact rather than a claim about whose reading it is — the gate rejects the subject, not the framing — so decoding costs nothing and `inspectParams` gets more out of it.

Either is defensible; it's a visible change to `unknown_3100` and better decided than discovered. If you keep the current behaviour it might be worth a line in the `applyParams` doc saying the passthrough is deliberately raw even where the dictionary knows an encoding.

Nothing else from me — `claimedParams` sitting in `capabilities/index.ts` beside `mergeProperties` is the right home for it, and the third spec case (a param the dictionary owns alone keeps its name) pins the part I'd have worried about.

---

martijnpoppen (approved):
Approving. The mechanism is right, `claimedParams` is the correct shape for it, and I went looking for the ways it could over-reach rather than re-reading the happy path — everything I found turned out to be the rule applied consistently.

**One thing to fix in the body before this is history, though: the verification only covers cameras.**

"No regression. All fourteen devices identical" is true and I reproduced it — a realistic T8114 record (1101, 2111, 1240, 1241, 1230, 1288, 1045, 1004) is byte-identical before and after. But that's the one device class where nothing changes, because the gates that fire here don't fire on cameras. I diffed `getProperties()` across the whole of `MODEL_REGISTRY`, base vs this branch: **67 of 76 models change at least one property name.** Two classes behave differently from your fleet.

**Vacuums — five names go, and the readings survive under the member's own name.** Realistic T2351 record (clean-namespace params only):

```
                   base            this PR
104   battery_level = 72     ->    unknown_104    = "72"    (163 battery = 72 either way)
103   look_for_sweeper       ->    unknown_103              (160 locating unchanged)
113   CleanType     = 1      ->    unknown_113              (154 cleanType unchanged)
109   ClearTime              ->    unknown_109              (167 clearTime unchanged)
107   forbid_mode            ->    unknown_107              (157 doNotDisturb unchanged)
```

All five are **read-aliases whose alias-level `available` refused them** — `vacuum_clean.battery` owns 163 with an alias on 104, and so on down. `claimedParams` adds the alias ids, the resolved schema drops them, so they withhold. Correct by your rule, and the value is still served under the member's name. This is the fix removing a duplicate the dictionary was publishing under an uglier spelling — I'd call it an improvement, but it is a public-surface change and `battery_level` disappearing off a robot vacuum is the kind of thing someone notices.

**Stations — the reading genuinely leaves the named surface.** On a HomeBase record carrying 1240/1241/1230/1288, `audio` resolves and its members are gated `available: isCameraCodec`:

```
1240  audioMicrophoneSwitch = true  ->  unknown_1240 = "1"
1241  audioSpeakerSwitch    = true  ->  unknown_1241 = "1"
1230  doorbellAudioVolume   = 50    ->  unknown_1230 = "50"
```

Nothing else names these, so they're gone from `getProperty`. And that is **exactly right** — `audio.microphone`'s own doc says "Camera-family only — a HomeBase has no microphone, so the getter and setter are both absent there". The dictionary was contradicting that doc in public, which is your bug, one class over. I flag it only because it's the sharp edge of the change: this converts every existing `available` gate from advisory into binding, so anywhere a gate is *over-broad* a real reading now vanishes instead of quietly surviving under a dictionary name. That's the right default and it will find things.

So: not a defect, but the blast radius is much wider than "a mains camera's cell reads", and the evidence in the body doesn't reach it. Worth saying plainly that this is a breaking change to the public property surface, with the vacuum aliases named — we're pre-1.0 so breaking is fine, discovering it isn't.

**What I checked and it held up:**

- The whole of #170's spec set passes against this implementation with mine removed — 190 files, 3538 tests, including the `getProperty`/`getProperties` pins under both the member's and the dictionary's spelling.
- The two deliberate exclusions are pinned and correct: an unresolved capability withholds nothing, and a param the dictionary owns alone keeps its name. The third spec case is the one I'd have worried about and it's there.
- Aliases are handled the way they have to be. Where `readAvailable` is false and an alias is promoted, the primary withholds, which is right. Where another capability legitimately owns the same id, the subtraction is by param id so it never withholds — I checked the four cross-module name collisions in the tree (`battery` → 1101/1101/163/8001, `charging` → 2111/1102, `rssi` → 1141×3/134, `doNotDisturb` → 1828/157) and none of them produce a wrong withhold.
- `withheld` is recomputed in `resolveInto`, so a widening re-resolve un-withholds correctly rather than only the constructor path being right.
- No over-withhold from non-gate causes: `propertiesOf`'s other filters (`writeOnly`, no param) apply to `module.properties` too, so they never enter the claimed set to begin with.

`claimedParams` living in `capabilities/index.ts` next to `mergeProperties` is the right home, and building it from the module lists rather than the merged one is what makes it correct — that's the part mine got wrong, as I've written up on #170.

The `unknown_3100` encoding question is in my comment above and I still lean your way; it needs a decision, not a change.


---

PR #179, opened 2026-09-15
fix(registry): type the wired T8200 as a doorbell

rwunsch:
## What

Adds a curated `MODEL_REGISTRY` row for the **Wired Doorbell 2K (T8200)**:

```ts
T8200: { codec: "camera", caps: ["doorbell"], name: "Wired Doorbell 2K" },
```

## Why

Without a row, T8200 falls through to the camera codec and never gains the `doorbell` capability.
Consumers that gate on it get nothing: `ha-eufy-sdk` builds no doorbell event entity
(`event.py`: `if "doorbell" in caps`) and no "Ringing" device trigger
(`device_trigger.py`: `"ringing": ("doorbellPress", "doorbell")`), so a doorbell press cannot be
acted on at all.

Observed on a real unit — `devices.list` returned
`["video","snapshot","motion","person_detection","camera","rtsp","arming","info"]` for a T8200.

## `caps: ["doorbell"]`, deliberately no `battery`

T8210 and T8214 are battery models and carry `["doorbell", "battery"]`. The T8200 is mains-powered,
so the row must not hand it a battery it does not have; the codec baseline plus inference stays the
only source of truth for power on this model.

`T8200X` appears alongside T8200 in the devices gallery, but I have no first-party evidence for it
and the registry doc is explicit that rows are added only as specifics are *confirmed* — so this
PR adds T8200 only.

## Test

`model.spec.ts` gains a case in the existing doorbell block asserting T8200 resolves with
`doorbell`, without `battery`, and with the curated name. Verified it fails without the registry
row:

```
AssertionError: expected [ 'video', 'snapshot', 'motion', …(3) ] to include 'doorbell'
 ❯ src/model/__tests__/model.spec.ts:247:28
```

— which reproduces the live symptom exactly.

## Gate

`npm run verify` green on Node 24: format, typecheck, all guards, build, `.d.ts` seal, ESM smoke,
`typecheck:examples`, and **182 test files / 3488 tests passing**.

## Not in scope

The same device also emits `doorbellPress` with no `deviceSn`, which breaks consumer attribution
independently of this fix. Reported separately in the linked issue; I could not confirm the cause
and did not want to bundle a guess into this change.

## Verified live on real hardware

Built eufy-sdk **v0.1.0** (the version the bridge add-on bundles) with only this row, confirmed the
resulting bundle is byte-identical to the shipped one apart from the 77-byte insertion, and ran it
on a real Home Assistant install with a T8200:

**Before** — `devices.list`:
`["video","snapshot","motion","person_detection","camera","rtsp","arming","info"]`

**After**:
`["doorbell","video","snapshot","motion","person_detection","camera","rtsp","arming","info"]`

and after a config-entry reload `ha-eufy-sdk` created the doorbell event entity
(`device_class: doorbell`, `event_types: ["pressed"]`) plus the doorbell-only Chime Switch
property. Both were absent before.

**The attribution defect is confirmed independent of this change.** With the device fully typed, a
physical press still logs `push in: doorbellPress sn=?`, and no push-driven entity fires. See the
linked issue — it is not limited to doorbells: `motion` pushes arrive with `sn=?` too.


src/model/registry.ts
```diff
@@ -54,6 +54,11 @@ export const MODEL_REGISTRY: Readonly<Record<string, RegistryEntry>> = {
   T8210: { codec: "camera", caps: ["doorbell", "battery"], name: "Video Doorbell" },
   // Confirmed against a real owned unit (named "Doorbell"): Video Doorbell Dual.
   T8214: { codec: "camera", caps: ["doorbell", "battery"], name: "Video Doorbell Dual" },
+  // Confirmed against a real owned unit: the mains-powered Wired Doorbell 2K. No `battery` row
+  // member on purpose — this model is wired, so the codec baseline plus inference is the whole
+  // truth for power. Without this row it classified as a plain camera, so the doorbell capability
+  // never appeared and consumers got no doorbell event entity or ring trigger.
+  T8200: { codec: "camera", caps: ["doorbell"], name: "Wired Doorbell 2K" },
```
lenoxys, line 61: Blocker. Setters are gated by `installs` — `requires`/`requiresRead`/`available` — not by what the device reported, and no doorbell member declares any of those. So this row hands a wired unit `setMechanicalChimeSwitch`, `setWdrSwitch`, `setDingdongVolume`, `setDingdongRingtone` and `setVideoQuality`, every one of them captured on a T8210/T8214 and both of those are battery doorbells. The capability file excludes 2015/2022/1306 because they "appear on NO owned device" — those are the wired ids, and this PR puts the first wired doorbell on the account, so that sentence stops being true the moment it merges. Fire-and-forget writes, so a wrong one reads as success — the same reason 1702's write is refused rather than guessed two members up.

`isWiredDoorbell` has been sitting in `device-family.ts` unused since it was written, keyed on exactly DeviceType 5. Gate the battery-only members' `available` on it, or the row is claiming a control surface no capture backs.

src/model/registry.ts
```diff
@@ -54,6 +54,11 @@ export const MODEL_REGISTRY: Readonly<Record<string, RegistryEntry>> = {
   T8210: { codec: "camera", caps: ["doorbell", "battery"], name: "Video Doorbell" },
   // Confirmed against a real owned unit (named "Doorbell"): Video Doorbell Dual.
   T8214: { codec: "camera", caps: ["doorbell", "battery"], name: "Video Doorbell Dual" },
+  // Confirmed against a real owned unit: the mains-powered Wired Doorbell 2K. No `battery` row
```
lenoxys, line 57: Twelve lines below this you'll find: "ptz is NOT curated on these rows on purpose — infer.ts proves it from evidence for the whole fleet ... via vendor deviceType 31 = INDOOR_PT_CAMERA". Doorbell is the same shape of fact and `DetectionSpec.deviceTypes` is already there for it; the reason it isn't used is that `doorbell` detects on `modelHints: [/doorbell/i]` alone — a regex over the model code and whatever the owner named the device, which is precisely why your unit came back a plain camera and why T8210/T8214 need `caps` in their rows at all.

Declare the doorbell DeviceType set (5, 7, 16, 91, 93, 94, 95, 96, and 203 for the LOCK_85V0 that classify.ts already calls a battery doorbell) on the capability. I ran it: row deleted, set added, `vitest run src/model` → 2350 pass, one failure, the curated name. That types T8200 plus T8200X / T8201 / T8203 — three more wired doorbells already in the gallery — and every doorbell on a known type, instead of a row each. Then this PR is a name-only row, and the `caps` on T8210/T8214 can go too: they're load-bearing only on a record carrying no `device_type` at all.

src/model/registry.ts
```diff
@@ -54,6 +54,11 @@ export const MODEL_REGISTRY: Readonly<Record<string, RegistryEntry>> = {
...
+  // member on purpose — this model is wired, so the codec baseline plus inference is the whole
+  // truth for power. Without this row it classified as a plain camera, so the doorbell capability
+  // never appeared and consumers got no doorbell event entity or ring trigger.
```
lenoxys, line 60: Nit: four lines of comment on a one-line row whose neighbours get one. The battery reasoning is already in the PR body and asserted in the test; the "without this row" history stops being true if the fact moves to detection.

src/model/__tests__/model.spec.ts
```diff
@@ -243,6 +243,16 @@ describe("doorbell — confirmed against real T8214", () => {
     expect(r.properties.some((p) => [2015, 2022, 1306].includes(p.paramType))).toBe(false);
   });
 
+  it("types the wired T8200 as a doorbell, not a plain camera, and claims no battery", () => {
```
lenoxys, line 246: A T8200 case under `describe("doorbell — confirmed against real T8214")`. Also: the `doorbellPress` claim in the PR body overstates it. Push 3103 has a single claimant, and `decodeEvent`'s push path only consults capabilities to break ties between contested ids — so the SDK was already emitting the press with no capability at all. What the missing row blocked is the consumer's own `"doorbell" in caps` gate and the property surface. Your own live note (press logged before the change) says the same thing.

lenoxys (commented):
Right symptom, wrong tier — and the row is half a change.

Blocker: a row saying `doorbell` on a mains-powered model installs the whole `CMD_BAT_DOORBELL_*` write surface on it, and none of those wires were captured on anything but a battery doorbell.

Cut: this fact belongs in `doorbell.ts` `detection.deviceTypes`, not in one row per SKU — the same file already argues that for `ptz` twelve lines above the row you added. Verified: with the DeviceType set and no row at all, T8200 gets `doorbell` and no `battery`, 2350 model tests pass, and only the curated-name assertion fails — which is the one thing a row is for.

lenoxys (approved):
Approved and merging as-is: the row is correct for what it claims and the live evidence stands. The two findings above are follow-up work, not this PR's — the DeviceType-based detection (which makes T8200X/T8201/T8203 and every future doorbell free) and the wired-vs-battery write gate come as their own PRs.

Issue #178, opened 2026-09-15
T8200 (Wired Doorbell 2K): no registry row, so no `doorbell` capability — and `doorbellPress` is emitted without `deviceSn`


rwunsch:
## Summary

A **Wired Doorbell 2K (T8200)** delivers `doorbellPress`, but nothing usable reaches Home
Assistant. Two separate gaps:

1. **T8200 has no `MODEL_REGISTRY` row**, so it resolves as a plain camera without the `doorbell`
   capability — `ha-eufy-sdk` therefore builds no doorbell event entity and no "Ringing" device
   trigger. *(PR attached.)*
2. `doorbellPress` is emitted with **no `deviceSn`**, so the forwarded event cannot be attributed
   to a device at all. *(Not addressed by the PR — I could not confirm the cause.)*

## Environment

- `ha-eufy-sdk-addon` (eufy-sdk bridge) 0.2.1, `ha-eufy-sdk` via HACS
- Home Assistant Core 2026.9.1, HAOS 18.2, aarch64
- Region DE. Bridge authenticated; all devices discovered and otherwise working.

## 1. T8200 is not typed as a doorbell

`devices.list` reports, for the doorbell:

```
model: T8200
capabilities: ["video","snapshot","motion","person_detection","camera","rtsp","arming","info"]
```

No `doorbell`. `MODEL_REGISTRY` curates `T8210` and `T8214` but has no `T8200` row, so the model
falls through to the camera codec. Downstream in `ha-eufy-sdk`:

- `event.py` — `if "doorbell" in caps:` never fires, so no `EufySdkDoorbellEvent` / `pressed`.
- `device_trigger.py` — `"ringing": ("doorbellPress", "doorbell")` is capability-gated, so the
  named "Ringing" trigger is absent.

The attached PR adds the row, with a regression test that fails without it.

## 2. `doorbellPress` carries no `deviceSn`

With the doorbell physically pressed, the bridge logs:

```
[bridge:event] push in: doorbellPress sn=? [detection → HA refreshes Last event] → broadcast to 1 frontend client(s)
```

`sn=?` is `boot.mjs` reporting `payload?.deviceSn ?? "?"`, and the same payload is what gets
broadcast, so the HA side receives an event with no `deviceSn`.
`custom_components/eufy_sdk/event.py::_handle_event` returns early at
`if data.get("deviceSn") != self._sn`, so no entity can fire regardless of capability. Confirmed
— nothing changed state on the press:

[...]

PR #180, opened 2026-09-15
fix(push): keep the whole envelope, so every push carries its device serial

rwunsch:
Fixes the attribution half of #178.

## Root cause

A eufy push carries **identity in the app_data envelope** and **detail in the base64 `payload`
entry**. `device_sn` and `station_sn` are *siblings* of `payload`, never children of it:

```
app_data: device_sn, station_sn, type, title, content, event_time, payload(base64) …
payload → { event_type, channel, pic_url, session_id, create_time, … }      ← no serial
```

`handleDataMessage` emitted:

```ts
payload: data.payload ?? data,
```

A `payload` entry is always present, so the envelope was always discarded and only the nested
detail survived. `event_type` lives in the detail and came through; `device_sn` lives in the
envelope and did not. Every push therefore reached consumers with `deviceSn: undefined`.

Two things already assumed the envelope shape, which is what makes this look like an oversight
rather than a decision:

- `normalizePushEvent` reads `env.payload ?? env` and falls back to `env.device_sn` — that
  fallback was unreachable.
- The existing specs in `__tests__/thumbnail-candidate.spec.ts` construct
  `{ payload: { device_sn, payload: { … } } }`, i.e. the envelope with the detail nested.

## The change

```diff
- payload: data.payload ?? data,
+ payload: data,
```

Producer, normaliser and specs now agree. Nested detail still decodes because
`normalizePushEvent` descends into `env.payload`. The `RawPushMessage.payload` doc comment is
updated to describe what it now carries.

## Impact

Consumer-visible and not niche: with `deviceSn` undefined, a consumer cannot route an event to a
device, so **no push-driven entity ever updates** — motion, person, package, doorbell alike. Only
polled state works.

## Test

New `__tests__/envelope-identity.spec.ts` drives `handleDataMessage` with a realistic app_data
frame (redacted serials) and asserts the emitted event carries `deviceSn`, `stationSn` and the
nested `eventType`. Without the change it fails:

```
AssertionError: expected undefined to be 'T8000P0000000000'
```

A second case pins `normalizePushEvent`'s envelope fallback directly; it passes either way, which
is the point — the normaliser was always correct.

## Verified on real hardware

Built into the bridge add-on's bundled SDK and run against a live account. Same log, across the
restart:

```
18:30:38  push in: motion        sn=?                 ← before
18:31:48  push in: motion        sn=T8200XXXXXXXXXXXX ← after
18:32:47  push in: doorbellPress sn=T8200XXXXXXXXXXXX
```

Home Assistant then fired the doorbell event (`event_type: pressed`) and the motion binary sensor
turned `on` — neither had ever fired before on this account.

## Gate

`npm run verify` green on Node 24: format, typecheck, all guards, build, `.d.ts` seal, ESM smoke,
snippets, `typecheck:examples`, **198 test files / 3589 tests**.


src/transport/push/push-client.ts
```diff
@@ -243,7 +243,12 @@ export class PushClient extends EventEmitter {
       persistentId: object?.persistentId,
       ttl: object?.ttl,
       sent: object?.sent,
-      payload: data.payload ?? data,
+      // Keep the WHOLE envelope, not just the decoded `payload` entry. The identity keys
+      // (`device_sn`, `station_sn`) are app_data siblings of `payload`, never inside it, so
+      // narrowing to `data.payload` here discarded them and every push arrived unattributed.
+      // `normalizePushEvent` already reads `env.payload ?? env` and falls back to `env.device_sn`,
+      // so it wants the envelope — that fallback was simply unreachable.
+      payload: data,
```
lenoxys, line 251: **Blocker.** `normalizePushEvent` descends exactly one level (`env.payload ?? env`), and this moves the whole tree one level down — so it now lands *on* the decoded JSON instead of inside it. Correct for a push whose decoded JSON **is** the detail. For a push whose decoded JSON nests a `payload` of its own it silently drops `eventType`, `thumbnailUrl`, `cipher` and `thumbnailCandidate`.

Measured, same frame through `handleDataMessage` on both sides: before → `deviceSn` undefined, `eventType` 3102, thumbnail resolved; after → `deviceSn` resolved, `eventType` undefined, thumbnail undefined, no candidate. The frame I used is the one this repo's own wire fixture in `src/__tests__/push.spec.ts` encodes — base64 JSON of `{device_sn, payload:{event_type, pic_url}}` — and it passes either way because it only asserts on `JSON.parse`, so nothing in the suite sees the loss.

And the wire claim carrying this is the part that needs the evidence. The live run shows the envelope carries identity for *that* push; it does not establish "never children of it" for every push, and the categorical version is what licenses a single-level descent. `types.ts` and the wire fixture have stated the nested shape since before this PR — overturning what the repo already documents takes the app_data key list off a capture, per the ground-every-wire-claim rule in `AGENTS.md`. Absent that, the safe shape is the one that needs no claim: walk the descent to the deepest `payload` and gather identity claims from every level passed, and the level stops being something anyone has to be right about.

src/transport/push/push-client.ts
```diff
@@ -243,7 +243,12 @@ export class PushClient extends EventEmitter {
       persistentId: object?.persistentId,
       ttl: object?.ttl,
       sent: object?.sent,
-      payload: data.payload ?? data,
+      // Keep the WHOLE envelope, not just the decoded `payload` entry. The identity keys
```
lenoxys, line 246: **Cut.** Three of these five lines are archaeology — what the old expression discarded, that the fallback was unreachable. A doc states ground truth, not its own history; that belongs in the commit, where you already wrote it.

`AGENTS.md` puts the rationale in a JSDoc above the declaration, not in `//` lines inside the body, and it names a body comment as the tell that the JSDoc is missing. `handleDataMessage` has none at all — that's the actual gap. The wire fact (identity keys are app_data siblings of `payload`) goes above the method; the body then needs no comment.

src/transport/push/types.ts
```diff
@@ -33,7 +33,7 @@ export interface RawPushMessage {
   persistentId?: string;
   ttl?: number;
   sent?: string;
-  /** The decoded eufy payload (the `payload` app_data entry, JSON-parsed). */
+  /** The eufy envelope: every app_data entry, with the `payload` entry JSON-parsed in place. */
```
lenoxys, line 36: This is now the only line in the file that knows. The module JSDoc at the top still says the base64 JSON "is the EufyPushMessage; its nested `payload` is device-type specific", and `EufyPushMessage` is still introduced as "the eufy JSON envelope inside the push" — both describe the shape you just moved away from. One fact, three places, two of them now wrong.

src/transport/push/__tests__/envelope-identity.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+    });
+
+    expect(seen).toHaveLength(1);
+    expect(seen[0]?.deviceSn).toBe("T8000P0000000000");
+    expect(seen[0]?.stationSn).toBe("T8000P0000000000");
+    // the nested detail still decodes — the envelope is kept in addition to it, not instead
+    expect(seen[0]?.eventType).toBe(3103);
+  });
+
+  it("normalizePushEvent resolves the serial from the envelope when the payload lacks one", () => {
```
lenoxys, line 41: **Delete.** You say yourself it passes either way — then it isn't a test of this change, and `thumbnail-candidate.spec.ts` already pins the envelope-level `device_sn` claim against a nested one. Eleven lines for coverage that exists.

src/transport/push/__tests__/envelope-identity.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+    const seen: PushEvent[] = [];
+    client.on("push", (e: PushEvent) => seen.push(e));
+
+    (client as unknown as { handleDataMessage(o: unknown): void }).handleDataMessage({
+      appData: appData(),
+    });
+
+    expect(seen).toHaveLength(1);
+    expect(seen[0]?.deviceSn).toBe("T8000P0000000000");
+    expect(seen[0]?.stationSn).toBe("T8000P0000000000");
```
lenoxys, line 36: Same placeholder for both keys, so this passes just as well if the normaliser reads station into device or the reverse. `availability.spec.ts` two files over uses `...0000` / `...0001` for exactly this reason.

src/transport/push/__tests__/envelope-identity.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+import { PushClient, normalizePushEvent } from "../push-client.js";
+import type { PushEvent } from "../types.js";
+
+/**
+ * A real eufy push carries its identity in the app_data envelope and its detail in the base64
+ * `payload` entry — `device_sn` is a SIBLING of `payload`, never a child. Narrowing the emitted
+ * envelope to the decoded `payload` therefore drops the serial, and every consumer receives an
+ * event it cannot attribute to a device. Shapes below are redacted but structurally real.
+ */
+const appData = (): { key: string; value: string }[] => [
```
lenoxys, line 11: Nit: one call site, and it returns a fresh array nobody mutates — inline it into the `handleDataMessage` call.

lenoxys (commented):
The producer change is the right call — `EufyPushMessage` carries `google.c.sender.id`, which can only ever be an app_data sibling, so the type has always described the envelope and the producer was the odd one out.

One blocker: the normaliser still descends exactly one level, so this trades identity for event detail on any push whose decoded JSON nests its own `payload` — measured, and this repo's own wire fixture encodes that shape. Then cuts to the comment block and the second test case.

`npm run verify` reproduces green here (198 files / 3589 tests). Separately, and not on a line: the PR body names a specific consumer and its stack. The consumer-agnostic hard rule covers PR bodies too — please scrub it.

lenoxys:
Two process notes, neither about the code.

Once this lands both halves of #178 are done (#179 took the registry row), so the body wants a `Closes #178` trailer — otherwise the issue outlives the fix and the next person re-reads it as open.

src/transport/push/push-client.ts
```diff
@@ -243,7 +243,12 @@ export class PushClient extends EventEmitter {
...
+      // (`device_sn`, `station_sn`) are app_data siblings of `payload`, never inside it, so
+      // narrowing to `data.payload` here discarded them and every push arrived unattributed.
+      // `normalizePushEvent` already reads `env.payload ?? env` and falls back to `env.device_sn`,
+      // so it wants the envelope — that fallback was simply unreachable.
```
martijnpoppen, line 250: This block trips two of `AGENTS.md`'s comment rules at once:

- **"Comments: JSDoc above the declaration, ground-truth, no inline body comments."** — `handleDataMessage` has no JSDoc at all right now, and this is exactly the "a body comment is a smell that the JSDoc is incomplete" case the rule describes.
- **"A JSDoc does not narrate its own history."** — "narrowing to `data.payload` here discarded them and every push arrived unattributed" and "that fallback was simply unreachable" describe what the old code did wrong. That reasoning belongs in the commit message (where you already have it, in full).

Suggested shape — keep the protocol fact, drop the diff narration, move it up:

```ts
/**
 * Decode one MCS `DataMessageStanza` into a {@link RawPushMessage} and the normalised event.
 *
 * `payload` carries the WHOLE app_data envelope, not just the decoded `payload` entry: the identity
 * keys (`device_sn`, `station_sn`) are app_data siblings of `payload`, never children of it, and
 * {@link normalizePushEvent} descends into `env.payload` for the detail.
 */
private handleDataMessage(object: any): void {
```

Then line 246 is just `payload: data,`.

---

src/transport/push/__tests__/envelope-identity.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+    const event = normalizePushEvent({
+      payload: {
+        device_sn: "T8000P0000000000",
+        payload: { event_type: 3103 },
+      },
+    } as never);
+
+    expect(event.deviceSn).toBe("T8000P0000000000");
+    expect(event.eventType).toBe(3103);
+  });
```
martijnpoppen, line 51: This second case calls `normalizePushEvent` directly, which this PR doesn't change — and `__tests__/thumbnail-candidate.spec.ts` already constructs the same `{ payload: { device_sn, payload: {…} } }` shape in six cases. You note in the PR body that "it passes either way, which is the point", but a spec that passes on both sides of the diff pins nothing about *this* change; the first case is the one carrying the regression.

If you want a second case that earns its place, the gap the first one leaves is the no-`payload`-entry envelope — app_data with only `device_sn`/`type` and no base64 entry at all. That used to hit the `?? data` fallback and now hits the same path unconditionally, and nothing currently covers it.

---

martijnpoppen (commented):
Reviewed the diff and checked out the branch locally. **The fix is correct** and I'd like to see it land.

The producer really was the odd one out: `EufyPushMessage` declares `device_sn`, `station_sn` and `google.c.sender.id` at envelope level alongside a nested `payload`, `normalizePushEvent` reads `env.payload ?? env` and falls back to `env.device_sn`, and every existing spec in `thumbnail-candidate.spec.ts` constructs the envelope shape. Three things assumed the envelope; only `handleDataMessage` narrowed past it.

**Verified locally** on `9408bfa`: `tsc --noEmit` clean, `prettier --check` clean, all five guards (`decorrelation`, `lines`, `docrefs`, `consumer-agnostic`, `capability-ownership`) pass, and the suite is green at 198 files / 3589 tests — matching your gate.

**I agree with @lenoxys that the one-level descent is the blocker**, so I won't restate it beyond a concrete suggestion: making the normaliser descend *while* `inner.payload` is an object, rather than exactly once, makes the fix independent of which nesting the wire actually sends, and costs three lines:

```ts
let inner: unknown = env;
while (inner && typeof inner === "object" && "payload" in inner) inner = (inner as { payload: unknown }).payload;
```

That also keeps `p.device_sn` / `p.s` resolution working, since identity is still read from both `p` and `env`.

Two smaller things inline. Separately, and echoing @lenoxys: the PR body names a specific consumer and its stack in the "Verified on real hardware" section — the consumer-agnostic hard rule covers PR bodies, so that wants scrubbing before merge.

One process note: `mergeable_state` is `unstable` and the API reports **zero check runs** on the head commit, so CI may not have run here at all. Worth confirming before merge rather than relying on the local gate.

---

max246:
Hi @rwunsch — following up to get this one moving again. The direction is right and both reviewers want it to land, but it's been idle for a week with the blocker still open, so a quick summary of what's needed to unblock:

**1. Blocker — the normaliser only descends one level (the real fix).** Your producer change (`payload: data` instead of `data.payload ?? data`) correctly keeps the envelope so `device_sn` survives. But `normalizePushEvent` does `let inner = env.payload ?? env` — a single descent. Moving the whole tree up one level means that for a push whose decoded JSON nests its own `payload` (the shape this repo's own wire fixture in `push.spec.ts` encodes), the normaliser now lands on the mid-level object and drops `eventType` / `thumbnailUrl` / `cipher`. Both @lenoxys and @martijnpoppen measured this and converged on the same 3-line fix — descend *while* there's a nested payload rather than exactly once:

```ts
let inner: unknown = env;
while (inner && typeof inner === "object" && "payload" in inner) inner = (inner as { payload: unknown }).payload;
```

Identity is still gathered from `env` (and each level passed), so this works regardless of which nesting the wire actually sends — and it removes the categorical "identity is never a child of payload" claim that would otherwise need a fresh app_data capture to justify under AGENTS.md's ground-every-wire-claim rule.

**2. Smaller items (all raised inline):**
- `handleDataMessage` comment block: move the wire fact into a JSDoc *above* the method and drop the `//` body lines / history narration (AGENTS.md). See martijn's suggested shape.
- `types.ts`: the module JSDoc + `EufyPushMessage` doc still describe the old "decoded payload is the message" shape — update them so one fact isn't stated three ways with two now wrong.
- New test: it passes on both sides of the diff, so it pins nothing about this change. The gap worth covering instead is the **no-`payload`-entry** envelope (app_data with only `device_sn`/`type`, no base64 entry). Also use distinct placeholder serials (`…0000` / `…0001`) so device/station can't be silently swapped.
- PR body: the "Verified on real hardware" section names a specific consumer + stack — the consumer-agnostic rule covers PR bodies too, please scrub it.

Once the normaliser descends to the deepest payload and the suite has a case that actually fails without the fix, this is good to go. Ping me if you'd like a hand with any of it. 🙏

danymexi:
The fix looks right to me, but I have a live counterexample to the premise, and it narrows the impact statement rather than the change.

I captured station `alarm` pushes on real hardware yesterday and today while working on #223, running the published `@mega-yfue/eufy-sdk@0.2.0-beta.25` — which still carries the unpatched line:

```
$ grep -oE 'payload: data[^,;]{0,30}' dist/index.js
payload: data.payload ?? data
```

Two `alarm` events off that build, verbatim, as they reached the consumer:

```jsonc
{ "event": "alarm", "deviceSn": "T8010P0000000000", "eventType": 10, "a": 10,
  "s": "T8010P0000000000", "n": "<home>", "c": 255, "t": "<timestamp>",
  "type": 4,  "automation_id": 0, "mode": 1, "phase": "triggered" }

{ "event": "alarm", "deviceSn": "T8010P0000000000", "eventType": 10, "a": 10,
  "s": "T8010P0000000000", "n": "<home>", "c": 255, "t": "<timestamp>",
  "type": 16, "automation_id": 0, "user_name": "<user>",
  "mode": 1, "phase": "triggered" }
```

`deviceSn` is populated, and so are the envelope siblings `s`, `n`, `c`, `t`, `type`,
`automation_id`, `mode` and `user_name`.

The reason is in the line itself: `data.payload ?? data` only discards the envelope **when
`data.payload` exists**. A station alarm push is flat — no base64 `payload` entry — so the
nullish coalescing falls through to `data` and the whole envelope survives by accident.

So "A `payload` entry is always present, so the envelope was always discarded" holds for the
pushes that carry a detail blob (motion, person, package, doorbell — the ones with `pic_url`
and `session_id`), and not for the ones that don't. The **Impact** section's "no push-driven
entity ever updates" is therefore too broad: alarm-type pushes were already routable, which is
how I could read `type` and `user_name` off them in mega-yfue/eufy-sdk#223.

None of this argues against the change. `payload: data` makes the two shapes agree instead of
one of them working by luck, and it means a consumer can stop wondering which kind of push it
is holding. It might be worth saying that in the PR body, though: a reviewer who tests with an
alarm push will see `deviceSn` populated before the fix and wonder whether the bug is real.

Happy to re-capture on the patched build if a before/after on the same hardware would help —
the alarm path I can drive on demand; the detail-carrying pushes I'd have to wait for.

_Device identifiers, home name, user name and timestamps redacted by a maintainer (repo rule: no real device data)._

lenoxys:
@rwunsch the direction still looks right, and danymexi's capture backs it. It can't merge with 8 inline review threads still open, though, including the blocker on the normaliser only descending one level. Please push a fix for those, or reply on the threads you disagree with.

Tralla:
Adding another real-hardware datapoint for this change.  
I applied ths PR cleanly on top of the current beta-0.2.0 and tested a package lifecycle on a T8214 / Video Doorbell E340 to test #220

The observed push shape was:
app_data
├── device_sn
├── station_sn
└── payload
    ├── event_type
    ├── device_sn
    ├── channel
    └── package fields
So for this device/event family there is one application payload level before event_type.
With this applied, the outer identity was preserved, the SDK correctly resolved the T8214 doorbell capability, and live testing produced:
 3301 → packageDelivered
 3302 → packageTaken
Both semantic events were emitted immediately and attributed to the correct device. 
(3304 / packageStranded uses the same existing mapping path, but I didnt test that live)

And no, this doesnt address the concern about push types with an additional nested payload level, it only confirms that the T8214 package events observed here use the one-level shape and that the PR fixes.

PR #181, opened 2026-09-15
fix(doorbell): the HomeBase-as-chime switch has a captured wire, so it ships a setter

lenoxys:
## What

1702 gains its write, and its published meaning is corrected: it is the **HomeBase acting as the doorbell's chime**, not the separate plug-in chime the app constant's name (`APP_CMD_BAT_DOORBELL_CHIME_SWITCH`) suggested.

```ts
chimeSwitch: {
  param: DOORBELL_CMD.CHIME_SWITCH,
  provenance: "verified",
  write: (v, ctx) => setScalar(DOORBELL_CMD.CHIME_SWITCH, asBool(v) ? 1 : 0, ctx, "direct-binary"),
}
```

## Why the write can ship now

The read was always confirmed; the frame never was, so the member carried `unverified: true` and no setter — sharing the 1702-1719 range with its captured siblings is not evidence of a shared frame shape.

It is now captured. Outer P2P cmd 1702 itself, signCode 8, on the device's own channel, the same 136-byte direct-binary `[u32 channel][u32 value][account_id pad 128]` struct as its 1703 sibling, value a plain boolean:

```
cmd=1702 sign=8 ch=2   VALUE=0        (app: switch off)
cmd=1702 sign=8 ch=2   VALUE=1        (app: switch on)
```

Attributed by the level-2 key that opened the frames plus the parameter that moved on the cloud record (`1702  1 -> 0` then `0 -> 1` on the doorbell's channel) — not by the capture tool's peer label, which is unreliable.

## Why the label changed

Both chime switches were pinned by toggling **one at a time** and reading which parameter moved:

| switch in the app | parameter written | nothing else moved |
| --- | --- | --- |
| "HomeBase as chime" | **1702** `chimeSwitch` | 1703 stayed put |
| "existing doorbell chime" | **1703** `mechanicalChimeSwitch` | 1702 stayed put |

The account has no plug-in chime at all, and its ding-dong settings (1717/1718) are not even offered in the app — consistent with those belonging to a chime unit that is not present. Both descriptions now carry the app's own wording plus that evidence, so no consumer re-derives which is which.

## Test

`npm run verify` green: 197 files / 3598 tests.

The spec that asserted the *absence* of a setter becomes the one that asserts the frame — value 1/0, on `ctx.channel` rather than a hardcoded 2 — and that `setChimeSwitch` is on the bound surface.

## Verified on hardware

Replayed through the **bound surface** (`dev.doorbell().setChimeSwitch()`), on a T8210 behind a HomeBase:

```
write 1: sent false → reads "0"  ✓
write 2: sent true  → reads "1"  ✓   (two consecutive writes in one session, no revert)
restore: device left as it was found
```

Two writes in one session on purpose: a setter that composes against state read at bind time passes a single write and silently reverts on the second, which a mocked reader cannot catch.

## Not in scope

`chimeSwitch` is **not renamed** to something like `homeBaseChime`. It is a published property name that reaches the manifest and consumers' entities, so renaming it breaks every host for a clarification the description now carries. Worth deciding before the beta, separately.

`1709` (`electronicRingtoneTime`) and `1714` (unknown) still carry app-constant labels no screen has confirmed, and the ding-dong pair stays unverifiable on this account.


martijnpoppen (approved):

src/model/capabilities/doorbell.ts
```diff
@@ -30,6 +31,19 @@ export const DOORBELL_CMD = {
...
+   * `CHIME_SWITCH`, which is why this was long described as a separate plug-in chime; toggling that one
+   * switch alone on a T8210 behind a HomeBase wrote 1702 and nothing else, and the account has no
+   * plug-in chime at all. Same 136-byte direct-binary shape as
+   * that sibling: `[u32 channel][u32 value][account_id ASCII, zero-padded to 128 bytes]`, outer P2P cmd
+   * 1702 itself, signCode 8, on the device's own channel. Reversed from a live capture (outer-cmd 1702,
+   * signCode 8, on the doorbell's `device_channel` — 2 for that unit, NOT a wire constant, always use
+   * `ctx.channel`): turning OFF sent `value=0`, turning ON sent `value=1` — a plain boolean.
+   * App `APP_CMD_BAT_DOORBELL_CHIME_SWITCH`.
+   */
+  CHIME_SWITCH: 1702,
```
martijnpoppen, line 46: **Placement falsifies three neighbouring doc claims.** Inserting `CHIME_SWITCH` *after* `MECHANICAL_CHIME_SWITCH` breaks the positional references the surrounding JSDoc already relies on:

1. Line 25 (this PR's own edit) — 1703's doc now says "param 1702 `chimeSwitch` **above**", but 1702 is thirteen lines **below** it.
2. Line 51 — `WDR_SWITCH`'s doc says "SAME 136-byte direct-binary shape/polarity/capture session as `MECHANICAL_CHIME_SWITCH` **immediately above**", and 1702 now sits between them.
3. `DINGDONG_VOLUME`'s doc says "UNLIKE the 1703/1704 direct-binary switches above" — as of this PR 1702 has joined that set, so the list is now incomplete.

Moving the `CHIME_SWITCH` block to sit **before** `MECHANICAL_CHIME_SWITCH` fixes all three at once and needs no prose edits: 1702's own doc already forward-references 1703 by `{@link}` rather than by position, so it reads correctly from either slot.

---

src/model/capabilities/doorbell.ts
```diff
@@ -30,6 +31,19 @@ export const DOORBELL_CMD = {
    * App `APP_CMD_BAT_DOORBELL_MECHANICAL_CHIME_SWITCH`.
    */
   MECHANICAL_CHIME_SWITCH: 1703,
+  /**
+   * Whether the HOMEBASE acts as the doorbell's chime — the app's "HomeBase as chime" switch, as opposed
+   * to the wired chime box {@link DOORBELL_CMD.MECHANICAL_CHIME_SWITCH} drives. The app constant reads
+   * `CHIME_SWITCH`, which is why this was long described as a separate plug-in chime; toggling that one
+   * switch alone on a T8210 behind a HomeBase wrote 1702 and nothing else, and the account has no
+   * plug-in chime at all. Same 136-byte direct-binary shape as
+   * that sibling: `[u32 channel][u32 value][account_id ASCII, zero-padded to 128 bytes]`, outer P2P cmd
```
martijnpoppen, line 40: `AGENTS.md`: **"A JSDoc does not narrate its own history. Not what an earlier version did, … not which guess was wrong."**

"The app constant reads `CHIME_SWITCH`, *which is why this was long described as a separate plug-in chime*" is exactly that — it documents the old wrong label rather than the wire. It also dates the doc: once nobody remembers the old description, the clause is just confusing.

The evidence in the same sentence is the part worth keeping. Something like:

> Whether the HOMEBASE acts as the doorbell's chime — the app's "HomeBase as chime" switch, as opposed to the wired chime box {@link DOORBELL_CMD.MECHANICAL_CHIME_SWITCH} drives. Toggling that switch alone on a T8210 behind a HomeBase wrote 1702 and nothing else. Same 136-byte direct-binary shape as that sibling: …

The "why we used to think otherwise" belongs in the commit message.

Same note, lighter, applies to the member doc below ("rather than the range-sharing guess the member refused before") — though that phrasing is in the spec, which is exempt.

---

src/model/capabilities/doorbell.ts
```diff
@@ -211,21 +222,20 @@ export function parseQuickResponses(
...
     kind: "boolean",
-    provenance: "mega",
-    unverified: true,
+    provenance: "verified",
     description:
-      "Wireless indoor chime enabled (1702 CMD_BAT_DOORBELL_CHIME_SWITCH; READ confirmed on T8214 — the " +
-      "WRITE wire is unconfirmed, so no setter is offered rather than guessing it).",
+      "The HomeBase plays the doorbell's chime (1702; the app's \"HomeBase as chime\" switch — write " +
+      "wire-confirmed live on T8210).",
+    write: (v, ctx) => setScalar(DOORBELL_CMD.CHIME_SWITCH, asBool(v) ? 1 : 0, ctx, "direct-binary"),
```
martijnpoppen, line 238: A question about how far the evidence reaches, not a blocker — the wire itself looks well captured.

The observation is: on a **T8210 behind a HomeBase**, toggling the app's "HomeBase as chime" switch alone wrote 1702, and that account has no plug-in chime. That's good grounding for "1702 is the non-wired chime switch", and for "on this topology the non-wired chime is the HomeBase". It doesn't yet separate those two readings — a doorbell paired with a plug-in chime and no HomeBase would drive the same app switch, and the app constant is the generic `CHIME_SWITCH`.

That matters because this string is caller-facing: it's what a host renders next to the control. If a user with a plug-in chime and no HomeBase sees "The HomeBase plays the doorbell's chime", the SDK is describing hardware they don't own.

Unless you can rule that out, something topology-neutral in the `description` is truer to the capture:

```ts
description: "The wireless chime rings (1702; the app's non-wired chime switch — write wire-confirmed live on T8210)."
```

…and the JSDoc above keeps the full HomeBase observation as the evidence trail, which is where it's authoritative.

---

src/model/capabilities/__tests__/doorbell.spec.ts
```diff
@@ -165,19 +165,24 @@ describe("doorbell capability module", () => {
...
+        param: DOORBELL_CMD.CHIME_SWITCH,
+        value: 1,
+        form: "direct-binary",
+        channel: 2,
+      });
+      expect(buildCommand("chimeSwitch", false, ctx(3))).toMatchObject({ value: 0, channel: 3 });
       const { acts } = bind<DoorbellActions>("doorbell", ctx(2));
-      expect("setChimeSwitch" in acts).toBe(false);
-      expect(DOORBELL.properties.some((p) => p.name === "chimeSwitch")).toBe(true);
+      expect("setChimeSwitch" in acts).toBe(true);
```
martijnpoppen, line 185: Small coverage loss: the rewrite drops

```ts
expect(DOORBELL.properties.some((p) => p.name === "chimeSwitch")).toBe(true);
```

That was the assertion pinning the **read** half — that promoting the write doesn't accidentally take the property out of the derived schema. It's one line and still meaningful after the change; worth keeping alongside the new `"setChimeSwitch" in acts` check so both directions stay covered.

---

martijnpoppen (commented):
Reviewed the diff and checked the branch out locally (noting it's still a draft, and already approved).

The substance looks right: a captured 136-byte direct-binary frame on the doorbell's own channel, 1=on/0=off, promoted from `unverified` to a real setter with the `provenance` bumped to `verified` and `model.spec.ts` updated to match. Promoting the wire to a `DOORBELL_CMD` constant rather than leaving `param: 1702` as a bare literal is the right call too — it's now consistent with its 1703/1704 siblings.

**Verified locally** on `0c71854`: `tsc --noEmit` clean, `prettier --check` clean, all five guards pass, suite green at 197 files / 3598 tests.

Four comments inline. The one I'd actually hold the merge for is the **constant placement** — inserting `CHIME_SWITCH` between 1703 and 1704 silently falsifies three "above"/"immediately above" claims in the surrounding JSDoc, including one this PR writes itself. Moving the block up two entries fixes all three with no prose changes. The rest are a JSDoc history-narration rule, one question about how far the "HomeBase" label generalises, and a dropped spec assertion.

---

PR #184, opened 2026-09-16
feat(solix): AE1X0 meter bindings + gate the tag->name table to meter frames (folds #186)

max246:
Promotes the Anker Solix Smart Meter (AE1X0) tag→field bindings from "structural inference" to confirmed, and removes a field the meter does not have.

## Evidence

**1. The vendor app's own field vocabulary.** The app names *exactly* twelve meter fields — `meterVoltageL1/L2/L3`, `meterCurrentL1/L2/L3`, `meterPowerL1/L2/L3`, `meterPowerTotal`, `meterImportEnergy`, `meterExportEnergy` — and no others. There is **no `meterCurrentTotal`**, no frequency and no power-factor field. So the earlier candidate `0xb2 meterCurrentTotal` was invented (the same class of mistake as the old `gridVoltage`).

**2. A live single-phase `param_info` frame** confirms the tag→field magnitudes:
- `0xac` ≈ a nominal mains voltage → `meterVoltageL1`
- `0xa8` == `0xab` (equal pair, one going negative on export) → `meterPowerL1` / `meterPowerTotal` (line == total on one phase)
- `0xaf` ≈ the line current → `meterCurrentL1`
- `0xb3` slowly cumulative → `meterImportEnergy`
- L2/L3 slots (`0xa9/aa/ad/ae/b0/b1`) read 0 on a single-CT install

**`0xb2` holds a small constant that does not track load** — so it is reserved, not a current total (as are `0xb5/0xb6/0xb7`).

## Changes

- **`SOLIX_METER_FIELD_NAMES`** — name the twelve app fields (add `0xa8`–`0xab`, `0xad`/`0xae`, `0xaf`–`0xb1`, `0xb3`/`0xb4`; keep `0xac`). `0xb2`/`0xb5`/`0xb6`/`0xb7` stay raw `channel_<hex>`. The stale `meterCurrentTotal` line is gone.
- **`SOLIX_ENERGY_METER_MEMBERS`** — promote the ten **electrical** fields as typed, evidence-gated members (V / A / W), each gated on its tag so L2/L3 stay absent on a single-phase device until a multi-phase frame carries them. The **energy counters are deliberately NOT members**: their tag→name is confirmed but their unit **scale** (Wh vs kWh) is not, so they stay named raw values rather than ship a member with a guessed unit.
- **Specs** — the decoder now names the app fields and keeps `0xb2` raw-only (never `meterCurrentTotal`); the members schema is the ten electrical fields; a promoted field (`meterCurrentL1` on `0xaf`) is evidence-gated.

## Verified
`npm run verify` green — **198 files, 3603 tests**, all guards; the docs build (TypeDoc `treatWarningsAsErrors`) is green too.


src/transport/mqtt/solix-mqtt.ts
```diff
@@ -41,23 +41,34 @@ export interface SolixParamFrame {
...
- * structural inference until a known-load capture pins it, and a mislabelled live float is worse than an
- * honest `channel_<tag>`. The recovered candidates, to re-add one line each (moving the tag from this
- * comment to the map above) as a known-load capture confirms each binding:
- *
- *   0xa8 meterPowerL1   0xa9 meterPowerL2   0xaa meterPowerL3   0xab meterPowerTotal
- *   0xad meterVoltageL2 0xae meterVoltageL3 0xaf meterCurrentL1 0xb0 meterCurrentL2
- *   0xb1 meterCurrentL3 0xb2 meterCurrentTotal 0xb3 meterImportEnergy 0xb4 meterExportEnergy
+ * The frame carries sixteen float slots (`0xa6`..`0xb7`). The four that name no field — `0xb2`, `0xb5`,
+ * `0xb6`, `0xb7` — are reserved and stay raw `channel_<hex tag>` (see {@link solixReadings}). `0xb2` in
+ * particular is NOT a current total: it holds a small constant that does not track load.
```
martijnpoppen: **The capture this rests on isn't in the repo, and the one that is contradicts part of it.**

`AGENTS.md`, the headline rule: *"Ground every wire claim in evidence you can point at."*

The only real frame in the tree is `solix-mqtt.spec.ts`'s `FRAME_HEX` — captured from an AE1X0, grid idle. Decoding it, **every float slot reads `0.0` except `0xac` (237.5)**:

```
a8 00000000   a9 00000000   aa 00000000   ab 00000000
ac 00806d43   ad 00000000   ae 00000000   af 00000000
b0 00000000   b1 00000000   b2 00000000   b3 00000000
b4 00000000   b5 00000000   b6 00000000   b7 00000000
```

So nothing committed can support the confirmations this doc asserts:

- "`0xa8` == `0xab` an equal power pair … one of them going negative on export" — both are 0, which is equally consistent with every other binding.
- "`0xaf` the line current" — 0.
- "`0xb3` a slowly-cumulative import counter" — 0, and a single frame can't show cumulative behaviour at all.
- "`0xb2` … holds a small constant that does not track load" — 0 here. (#186's doc says it read ~0.007; also not in any committed frame.)

And this PR adds no new fixture — the transport spec change asserts `expect(values.meterPowerL1).toBe(0)` against the same idle frame, which passes for any tag→name mapping you could write.

That's what makes this hard to review rather than wrong: I believe you saw these values, but the reviewer can't check them and neither can the next person to touch this table. **Adding the load-varying capture as a second redacted `FRAME_HEX` and asserting the bindings against it would settle this PR completely** — non-zero L1 power and current, a non-zero total matching L1, and ideally two frames minutes apart so `0xb3` visibly increments. That single fixture converts every claim above from assertion into something the suite enforces.

`0xb2` is worth calling out separately: "NOT `meterCurrentTotal`" is the right conclusion, and it's the one claim a load capture proves cheaply (it stays put while `0xaf` moves).

---

src/model/capabilities/solix.ts
```diff
@@ -54,6 +51,87 @@ export const SOLIX_ENERGY_METER_MEMBERS = {
...
+    type: "number",
+    kind: "scalar",
+    unit: "V",
+    provenance: "apk",
+    description: "Meter line-3 voltage (V) — ff09 tag 0xAE; 0 on a single-phase install.",
+  },
+  /** Line-1 current (A), ff09 tag `0xAF` — confirmed live (the line's CT current). */
+  meterCurrentL1: {
+    param: 0xaf,
+    type: "number",
```
martijnpoppen, line 81: **`provenance: "apk"` claims more than the evidence supports for the L2/L3 members.**

`types.ts` defines `apk` as *"extracted from the v6 app itself (the ids the app actually sends — authoritative)"*. What the app authoritatively supplies here, per this PR's own transport doc, is the **field list** — "the app's field vocabulary is exactly these twelve". Which *tag* carries L2 vs L3 comes from assuming the frame lays out as `a8-aa` power / `ab` total / `ac-ae` voltage / `af-b1` current, anchored by the three L1-position values.

That's a structural inference — and it's specifically the thing the comment this PR deletes was written to refuse:

> the app exposes the field *list*, but the tag→name *binding* below is a structural inference until a known-load capture pins it, and a mislabelled live float is worse than an honest `channel_<tag>`

The block layout is a reasonable hypothesis and I'd probably bet on it too. But the deleted comment isn't wrong about the standard, and `guessed` exists for exactly this ("a plausible placeholder; lowest trust").

Two ways out, either fine:
- If the decompiled app maps tag bytes to field names **directly**, quote that in the JSDoc and `apk` is correct for all nine.
- If it's the ordering, mark L2/L3 `guessed` and keep `verified` for the four positions the capture actually anchors.

Worth noting a mislabel here is not recoverable by the caller the way `channel_<hex>` is: `meterCurrentL3` is a typed member with a unit, so a wrong binding ships as confident wrong data rather than an honest raw channel.

---

src/model/capabilities/solix.ts
```diff
@@ -26,26 +26,23 @@ export type SolixCapability =
...
 
 /**
- * The `energyMeter` surface, declared once. Only the ONE confirmed tag→name binding is a member:
- * `meterVoltageL1` (ff09 tag `0xAC`), confirmed against a live single-phase frame. The evidence gate
- * (`bindMembers`) installs its getter only once a frame carrying tag `0xAC` has landed and answers
- * `undefined` before.
+ * The `energyMeter` surface — the electrical readings the vendor app names, as typed members. Each is
+ * read-only and evidence-gated: `bindMembers` installs a getter only once a frame carrying that tag has
+ * landed, and answers `undefined` (not a fabricated `0`) before — so on a single-phase / single-CT
+ * install the L2/L3 members stay absent until a multi-phase frame carries them.
```
martijnpoppen: **This contradicts the member descriptions twelve lines down, and the members are the ones telling the truth.**

Here: *"on a single-phase / single-CT install the L2/L3 members stay absent until a multi-phase frame carries them."*
Below: *"Meter line-2 voltage (V) — ff09 tag 0xAD; **0 on a single-phase install**."*

The gate in `bindMembers` is:

```ts
const reported = available && (reads(m, ctx) || (from !== undefined && ctx.paramIds.has(from.param)));
```

— the param having been **reported**, not carrying a non-zero value. And the committed idle frame reports `0xa9`, `0xaa`, `0xad`, `0xae`, `0xb0`, `0xb1` (all at `0.0`). So on a single-phase install the L2/L3 getters **do** install and answer `0`; they are not absent.

That leaves a caller unable to tell "this meter has no L2" from "L2 is currently idle", which is the distinction the paragraph is reaching for. Worth fixing the paragraph, since the members' own wording is already correct:

> …and answers `undefined` (not a fabricated `0`) before any frame has landed. Note the gate is *reported*, not *non-zero*: a single-phase meter reports its L2/L3 slots as `0.0`, so those members install and read `0` rather than staying absent.

If you'd rather have the stronger behaviour the paragraph describes, that's a real change (gating on a non-zero observation), and I'd say no — it would make a genuinely-zero L1 power reading disappear too.

---

src/transport/mqtt/solix-mqtt.ts
```diff
@@ -41,23 +41,34 @@ export interface SolixParamFrame {
...
- * nothing on the wire is lost — a caller reads unconfirmed tags there. The names are deliberately NOT
- * asserted for the rest: the app exposes the field *list*, but the tag→name *binding* below is a
- * structural inference until a known-load capture pins it, and a mislabelled live float is worse than an
- * honest `channel_<tag>`. The recovered candidates, to re-add one line each (moving the tag from this
- * comment to the map above) as a known-load capture confirms each binding:
- *
- *   0xa8 meterPowerL1   0xa9 meterPowerL2   0xaa meterPowerL3   0xab meterPowerTotal
- *   0xad meterVoltageL2 0xae meterVoltageL3 0xaf meterCurrentL1 0xb0 meterCurrentL2
- *   0xb1 meterCurrentL3 0xb2 meterCurrentTotal 0xb3 meterImportEnergy 0xb4 meterExportEnergy
+ * The frame carries sixteen float slots (`0xa6`..`0xb7`). The four that name no field — `0xb2`, `0xb5`,
```
martijnpoppen: Off by two: `0xa6..0xb7` is **18** slots, not sixteen.

The sixteen figure is right — it's what makes your own arithmetic work (12 named + 4 reserved) — so the range is `0xa8..0xb7`. `0xa6` and `0xa7` aren't float slots:

- `0xa6` is present in the captured frame as type `0x03`, not `0x05`, so `solixReadings` drops it — the existing spec asserts exactly that (`expect(values["channel_a6"]).toBeUndefined()` with the comment "a6 is a non-float type (0x03) → excluded from readings").
- `0xa7` doesn't appear in the frame at all.

Worth fixing because the `tag < 0xa6` filter in `solixReadings` is the reason `0xa6`/`0xa7` are in anyone's head here, and this sentence is now the closest thing to documentation of where the measurement block actually starts.

---

martijnpoppen (commented):
Reviewed alongside #185, #186 and #187 (all four checked out locally). **Of the four this is the one I'd hold**, and the reason is evidence rather than code.

**Verified locally** on `8621f8d`: `tsc --noEmit` clean, `prettier --check` clean, all five guards pass, suite green at 198 files / 3603 tests. The mechanics are fine; the members table, the schema derivation and the `0xb2` exclusion are all wired up correctly.

### Merge order first

**#184 and #186 conflict.** I tested all six pairings of the Solix PRs locally and this is the only conflicting one:

```
PR 184 + PR 186: CONFLICT
```

Both rewrite the `SOLIX_METER_FIELD_NAMES` doc block and the same spec case. The ordering also matters semantically: **#186 must land first.** This PR grows that table from 1 entry to 12 *without* #186's product-code gate, so merged on its own a Solarbank frame would pick up twelve wrong meter names instead of one — multiplying by twelve exactly the mislabelling #186 exists to prevent. Worth noting the dependency in the description so it can't get lost.

### The blocker

**The capture these bindings rest on isn't in the repo, and the committed one contradicts parts of the doc.** The only real frame in the tree is grid-idle: every float slot reads `0.0` except `0xac`. So "`0xa8` == `0xab` an equal power pair", "`0xaf` the line current", "`0xb3` a slowly-cumulative import counter" and "`0xb2` holds a small constant" can't be checked against anything committed — and the new spec asserts `expect(values.meterPowerL1).toBe(0)` against that same idle frame, which would pass for any mapping at all. Meanwhile `meterPowerL1`, `meterPowerTotal` and `meterCurrentL1` ship `provenance: "verified"`.

I'm not doubting that you saw these values — the block layout you're inferring is coherent and I'd bet on it too. The problem is that neither a reviewer nor the next person to edit this table can verify it, which is what `AGENTS.md`'s "Ground every wire claim in evidence you can point at" is protecting against. **One redacted load-varying capture added as a second `FRAME_HEX`, with the bindings asserted against it, would resolve essentially this entire review** — and it would make the `0xb2` exclusion self-evident rather than a claim.

Related, inline: the L2/L3 members are labelled `apk` when the binding looks like ordering inference rather than something the app states directly; the table JSDoc contradicts the member descriptions about whether L2/L3 stay absent on a single-phase install (the descriptions are right); and "sixteen float slots (`0xa6`..`0xb7`)" is off by two.

One thing I'd keep exactly as-is: holding `meterImportEnergy`/`meterExportEnergy` back from the members table because the unit scale is unconfirmed, while still naming them in the raw values. That's the right instinct, and it's the same instinct the rest of the PR is arguing against.

---

max246:
Have a look at the admin repo for evidence

Given that, I'll take this PR the way you suggested rather than argue it:
1. Commit the **load-varying meter frame** we have (`a8=11.4 ab=11.4 ac=240 af=1.58 b2=0.007 b3=61.059`, in `SOLIX_BATTERY_FIELDS_LIVE.md`) as a second `FRAME_HEX` and assert the bindings against it — makes the anchored tags checkable and the `0xb2`-constant claim self-evident.
2. **Downgrade L2/L3 from `apk` → `guessed`** (never non-zero in any frame, no code, no thomluther).
3. Fix the two nits (`0xa6..0xb7` → `0xa8..0xb7`; the JSDoc/member L2/L3 contradiction).

Thanks for holding the line on this — the standard caught a real overclaim.

max246:
Folded #186 in here (commit 50bdbd3) and closed it, so there's no merge-order dependency left — the 12-name meter table and its meter-family gate now land together. Without the gate this PR would apply all twelve names to a Solarbank frame (0xac etc.), which is exactly what #186 existed to prevent, so combining removes the footgun rather than sequencing around it.

#186's three review points are addressed in the fold: `solixReadings(frame, productCode)` takes a **required** productCode (pass `""` for unknown origin), `SOLIX_METER_PRODUCT_PREFIXES` is a `readonly string[]` kept in lockstep with model-side `SOLIX_METER_MODELS` (cross-referenced both ways), and the inline call-site comment is gone. Spec covers meter-frame naming, the Solarbank (AE103) non-meter path, and the unknown-origin path.

`npm run verify` green: 198 files / 3605 tests, all five guards, format, typecheck, build.

martijnpoppen (commented):
Re-reviewed at `50bdbd3`. **Every finding from my review is addressed, and folding #186 in was the right call** — it removes the ordering hazard entirely rather than leaving it as a merge instruction someone had to remember.

Checked each one against the new head:

- **Load-varying fixture** — `LOAD_FRAME` decodes as claimed: `0xa8` and `0xab` both 11.4 W (the L1==total mirror), `0xac` 240.0 V, `0xaf` 1.58 A, `0xb3` 61.059, `0xb2` 0.007. And the recomputed XOR is genuinely load-bearing here, not decorative: `decodeSolixParamFrame` rejects on `xor !== 0`, so a botched synthesis would return `null` and fail the spec rather than passing quietly. ✅
- **L2/L3 `apk` → `guessed`**, with the reason stated ("the app receives the meter as named JSON — no tag→phase decoder in the binary — so the phase assignment is block-ordering inference"). That's a better answer than I asked for: it names *why* the app can't be the authority, which is the part a future reader needs. ✅
- **Members-table JSDoc contradiction** — now says the gate is "reported", not "non-zero", and that L2/L3 install and read `0`. ✅
- **`0xa6..0xb7` → `0xa8..0xb7`.** ✅
- **`productCode` required, not optional**, with `""` as the explicit unknown-origin value. ✅
- **`SOLIX_METER_PRODUCT_PREFIXES` is an array mirroring `SOLIX_METER_MODELS`**, and each side's JSDoc names the other as the lockstep partner. ✅
- **Inline comment at the call site dropped.** ✅

**Verified locally** on `50bdbd3`: `tsc --noEmit` clean, `prettier --check` clean, all five guards pass, suite green at 198 files / 3605 tests.

One residual, explicitly **not** a blocker — just so it's on the record rather than discovered later. `LOAD_FRAME` is synthesised from the capture's magnitudes rather than being the capture, and the comment says so plainly, which I appreciate. What it pins is the decoder: if someone reshuffles `SOLIX_METER_FIELD_NAMES` the spec now fails, which is real protection and the thing that was missing before. What it can't do is independently corroborate the magnitudes — `meterPowerL1 === meterPowerTotal` holds because the same four bytes were written to both slots. So the four `verified` bindings still rest on the live observation, with the fixture as a regression guard rather than a second witness. If a redacted real load frame ever turns up (the idle one is already committed with a synthetic serial, so there's precedent for shape), swapping it in would close that last gap for free.

No objection from me at this point. The L2/L3 members shipping as typed getters marked `guessed` is a judgement call that's yours rather than mine — the label is now honest, which was my actual concern.

---

max246:
Addressed the last residual (commit 3f007e6): swapped the synthesised `LOAD_FRAME` for a **real** second AE1X0 capture — meter under a small net import, only the serial redacted + XOR recomputed. So `meterPowerL1 === meterPowerTotal` is now the device's own report (it sent both slots = 2.7 W) rather than the same bytes written twice, which independently corroborates the L1==total binding. Assertions updated to the real magnitudes (power 2.7, voltage 239.3, current 1.371, import 74.108, b2 0.009). Full verify green (3605 tests).

martijnpoppen (commented):
Verified `3f007e6` locally. **The real capture closes the residual** — that was the last thing I had open on this PR, and it's now a genuine second witness rather than just a regression guard.

Decoded the new `LOAD_FRAME` byte by byte:

```
declaredLen 160  bufLen 160  xor 0 (valid)
serial: AE1X0EXAMPLE00001
0xa8 2.7000   0xa9 0.0000   0xaa 0.0000   0xab 2.7000
0xac 239.3000 0xad 0.0000   0xae 0.0000   0xaf 1.3710
0xb0 0.0000   0xb1 0.0000   0xb2 0.0090   0xb3 74.1080
0xb4 0.0000   0xb5 0.0000   0xb6 0.0000   0xb7 0.1000
```

Every assertion matches, the XOR validates after the serial swap, and the serial is the same synthetic `AE1X0EXAMPLE00001` as the idle fixture — I grepped both fixtures for serial and IP shapes and found nothing real. `a8 == ab` at 2.7 W now genuinely comes from the device reporting both slots, which is exactly the corroboration that was missing. Gate green here too: `tsc --noEmit`, `prettier --check`, all five guards, 198 files / 3605 tests.

### One thing the new capture changes, and it's in the doc rather than the code

Having two real frames makes the `SOLIX_METER_FIELD_NAMES` claim about the reserved tags checkable for the first time — and it doesn't survive:

| tag | idle frame | load frame |
|-----|-----------|-----------|
| `0xb2` | `0.000` | `0.009` |
| `0xb7` | `0.000` | `0.100` |

The JSDoc says *"`0xb2` in particular is NOT a current total: it holds a small constant that does not track load."* The "not a current total" half is now solidly evidenced — a current total would read ~1.371 alongside `0xaf`, and it reads 0.009. But "a small constant that does not track load" is contradicted by your own two fixtures: `0.000` at idle, `0.009` under load. It moves with load, just not proportionally to current. `0xb7` does the same thing (`0.000` → `0.100`) while being described as reserved.

This changes no behaviour — both stay unnamed and raw, which is right either way — so it's a wording fix, not a blocker. Something like:

> The four that name no field — `0xb2`, `0xb5`, `0xb6`, `0xb7` — stay raw `channel_<hex tag>`. `0xb2` in particular is NOT a current total: under a 1.371 A line current it reads 0.009, three orders of magnitude off. Both `0xb2` and `0xb7` read zero at idle and non-zero under load, so they carry *something* load-related; what, is not established.

Flagging it only because "a constant that does not track load" is the kind of claim that gets cited later as settled — and this PR has been careful about exactly that everywhere else.

One note for whenever a third capture turns up: `0xb2` ≈ 0.009 against P 2.7 W / V 239.3 / I 1.371 puts it near the power factor (2.7 ÷ (239.3 × 1.371) ≈ 0.0082), and a PF field would read 0 at idle too. Not close enough to claim, and your reading of the app says it names no power-factor field — but it's the hypothesis one capture at a different power factor would settle outright.

Nothing else outstanding from me on this PR.

---

max246:
Good catch — the two fixtures do falsify it, and 'a constant that does not track load' is exactly the kind of thing that shouldn't ship as settled. Fixed in c105f27: the JSDoc now reads your wording — `0xb2` is NOT a current total (0.009 under a 1.371 A line current, three orders off), and both `0xb2` and `0xb7` read zero at idle and non-zero under load, so they carry something load-related that isn't established. Fixed the two test comments that repeated the old claim too. No behaviour change — both stay raw and unnamed.

On the power-factor lead: nice — 2.7 ÷ (239.3 × 1.371) ≈ 0.0082 vs the observed 0.009 is close, and a PF field reading 0 at idle fits. Left it as your note rather than a claim: the app names no power-factor field (so no authority for the label), and one capture can't separate PF from any other load-correlated quantity. A second load frame at a different power factor would settle it — flagged for whenever one turns up. Thanks for decoding the frame byte-by-byte; that's the check I'd want on a redacted fixture.

martijnpoppen (commented):
Verified `c105f27` — wording is right, the old claim is gone from the source and the two test comments, and nothing behavioural moved. **Clear from my side; no open findings on this PR.**

Gate green locally: `tsc --noEmit`, `prettier --check`, all five guards, 198 files / 3605 tests. I also ran `cd docs && npx typedoc` this time — exits `0`, no warnings. (Worth doing on every branch from now on: it isn't part of `npm run verify`, and it's what's currently red on #196. Detail over there.)

Agreed on leaving the power-factor lead as a note rather than a label — one capture can't separate PF from any other load-correlated quantity, and the app naming no such field means there's no authority for it either way. It's in the right place: recorded for whoever gets a second load frame, claimed nowhere.

---

martijnpoppen (approved):

PR #185, opened 2026-09-16
fix(solix): detect a Solarbank's capabilities (category trailing space + newer models)

max246:
## Detect a Solarbank's capabilities — category trailing space + newer models

Two fixes so a Solarbank resolves its capabilities correctly:

1. **Category trailing space.** The live `product_categories` catalog returns the home-battery category
   as `"Plug-in Home Battery "` (trailing space), which missed `CATEGORY_CAPABILITIES` and left a
   Solarbank with no capabilities. Fixed by trimming the category **at ingest** in `buildModelIndex`, so
   `identity().category`, `detectSolixCapabilities`, and any exact-match category comparison all see the
   clean name (one normalisation at the source, not per call site).

2. **Product-code prefixes.** `SOLARBANK_MODELS` grants `battery` + `solarInput` by product code so a
   device built without the catalog is still classified. Prefixes are **grounded in the live catalog**:
   `A17C0`–`A17C5` (Solarbank 1/2/3) and `AE100` (SOLIX Power Dock) all list under `"Plug-in Home
   Battery "`, and `AE103` (Solarbank 4 E5000 Pro) by the device spec → `A17C`, `AE10`.

### Review follow-ups (all addressed)
- Trim moved from the lookup to ingest (`buildModelIndex`) + a dedicated trim test.
- The two `//` comment blocks are now JSDoc; the meter-family gate (#186) is cross-linked.
- The speculative `A17E` / `AE11` prefixes were **dropped** — `A17E` is absent from the catalog, and the
  only `AE11x` product (`AE113` "XE 6/8kW") is a Residential Storage System, a different unverified
  family; grouping it as a Solarbank would over-grant capabilities.

Full verify green (3596 tests).


src/model/capabilities/solix.ts
```diff
@@ -102,7 +109,9 @@ export function detectSolixCapabilities(rec: SolixDetectionInput, category?: str
   const caps = new Set<SolixCapability>(["identity"]);
   if (rec.device_sw_version) caps.add("firmware");
   if (rec.wifi_online !== undefined || rec.rssi != null || rec.wifi_name) caps.add("connectivity");
-  for (const c of (category && CATEGORY_CAPABILITIES[category]) || []) caps.add(c);
+  // Trim the category: the catalog returns some names with trailing whitespace (e.g. "Plug-in Home
+  // Battery "), which would otherwise miss the map and leave a Solarbank with no capabilities.
+  for (const c of (category && CATEGORY_CAPABILITIES[category.trim()]) || []) caps.add(c);
```
martijnpoppen: **Right diagnosis, but fixed one level too low — trim at ingest instead.**

`.trim()` here repairs *this* lookup, but the untrimmed string is still what the rest of the SDK sees. Your own new spec asserts it:

```ts
expect(sb.identity().category).toBe("Plug-in Home Battery ");
```

So a caller writing `dev.identity().category === "Plug-in Home Battery"` gets `false`, and `buildModelIndex` stores the trailing-space name in the index every consumer reads.

**This already bites #187.** That PR adds `CATEGORY_FAMILY` and `isSolixSolarbank`, both comparing `category === "Plug-in Home Battery"` exactly. I merged this branch with #187 locally and ran a catalogued plug-in home battery whose product code isn't in `SOLARBANK_MODELS`:

```
category = "Plug-in Home Battery " | family = unknown | caps = identity,battery,solarInput,acOutput
```

Right capabilities (your fix works), but `family: "unknown"` — so `site.solarbanks()` silently omits it. `AE103` escapes only because this PR *also* adds the `AE10` prefix; the category path stays broken for anything else.

Trimming once where the catalog is parsed — in `buildModelIndex`, as the category is read off `SolixProductCategory.name` — fixes this call site, `identity().category`, and both of #187's comparisons, with no coordination needed between the two PRs. Then this line goes back to `CATEGORY_CAPABILITIES[category]` and the comment below becomes unnecessary.

---

src/model/capabilities/solix.ts
```diff
@@ -67,7 +67,11 @@ export type SolixEnergyMeterReads = Surface<typeof SOLIX_ENERGY_METER_MEMBERS>;
  */
 export const CATEGORY_CAPABILITIES: Readonly<Record<string, readonly SolixCapability[]>> = {
   "Portable Power Station": ["battery", "acOutput", "solarInput"],
-  "Plug-in Home Battery": ["battery", "solarInput", "acOutput", "energyMeter"],
+  // NOT `energyMeter`: a home battery measures its own grid/PV/output power, but that is a different
+  // ff09 tag family than the AE1X0 Smart Meter's — the `energyMeter` members are AE1X0-specific, so a
+  // Solarbank frame decoded through them mislabels (e.g. tag `0xac` is power on a Solarbank, voltage on
+  // the meter). `energyMeter` is model-detected for the meter (SOLIX_METER_MODELS), not category-detected.
```
martijnpoppen: The reasoning is good and I'd keep every word of it — but `AGENTS.md` wants it as JSDoc, not a `//` block ("Comments: JSDoc above the declaration … no inline body comments"). For an object entry that means:

```ts
export const CATEGORY_CAPABILITIES: Readonly<Record<string, readonly SolixCapability[]>> = {
  "Portable Power Station": ["battery", "acOutput", "solarInput"],
  /**
   * NOT `energyMeter`: a home battery measures its own grid/PV/output power, but that is a different
   * ff09 tag family than the AE1X0 Smart Meter's — tag `0xac` is power on a Solarbank, voltage on the
   * meter. `energyMeter` is model-detected for the meter ({@link SOLIX_METER_MODELS}), not
   * category-detected.
   */
  "Plug-in Home Battery": ["battery", "solarInput", "acOutput"],
```

Same note applies to the `// Trim the category…` block in `detectSolixCapabilities` below, which belongs in that function's JSDoc (or disappears entirely if the trim moves to ingest).

Worth cross-linking #186 here too — it's the PR that actually enforces the `0xac`-is-different-per-family claim this comment rests on.

---

src/model/capabilities/solix.ts
```diff
@@ -80,9 +84,12 @@ export const SOLIX_METER_MODELS: readonly string[] = ["AE1X0"];
...
 /**
  * Product-code prefixes for the grid-tie Solarbank / home-battery family (detects `battery` +
- * `solarInput` regardless of category): A1790 = Solarbank E1600 gen-1, A17C* = Solarbank 2 / 3.
+ * `solarInput` regardless of category, so a caller that builds a device without the catalog still gets
+ * them): A1790 = Solarbank E1600 gen-1, A17C* = Solarbank 2 / 3, A17E* = Solarbank Max AC, AE10* =
+ * Solarbank 4 E5000 Pro / Power Dock, AE11* = Solarbank Max. (`AE1X0`/`AE1R0` meters start `AE1X`/`AE1R`,
+ * so `AE10`/`AE11` do not catch them.)
  */
-export const SOLARBANK_MODELS: readonly string[] = ["A1790", "A17C"];
+export const SOLARBANK_MODELS: readonly string[] = ["A1790", "A17C", "A17E", "AE10", "AE11"];
```
martijnpoppen: Three prefixes added; what's the source for each?

- `AE10` — grounded by the new AE103 spec case. ✅
- `A17E` = "Solarbank Max AC" and `AE11` = "Solarbank Max" — named confidently, but nothing in the diff or PR body says where those mappings come from, and neither has a test.

Not the same severity as a wire claim (a product-code prefix is pure classification, and a wrong one over-grants `battery`/`solarInput` rather than mislabelling a value), but `AGENTS.md`'s "Ground every wire claim in evidence you can point at" cuts the same direction, and the note about third-party eufy clients being wrong for this protocol applies to Anker model tables too.

If these came from the catalog response on your account, saying so in the PR body is enough. If they came from a community model list, they're worth marking as unconfirmed — a false positive here grants capabilities to a device that doesn't have them, which is the failure mode `detectSolixCapabilities` is otherwise careful to avoid.

Also worth double-checking the exclusion argument in the JSDoc holds for these two: `AE1R0` (named in your comment as the other meter) starts `AE1R`, so `AE11` doesn't catch it — correct. But if a meter prefix ever lands in the `AE10`/`AE11` space, the two tables collide silently, and `detectSolixCapabilities` would add both `energyMeter` and `battery`.

---

martijnpoppen (commented):
Reviewed alongside #184, #186 and #187 (all four checked out locally).

Both halves of the diagnosis are real — a trailing space in the live catalog category silently costing a Solarbank every one of its capabilities is a good catch, and it's the kind of bug that would have been very annoying to chase from the outside. Dropping `energyMeter` from the home-battery category is right too, and reads as the model-side half of #186.

**Verified locally** on `9d29bdf`: `tsc --noEmit` clean, `prettier --check` clean, all five guards pass, suite green at 198 files / 3595 tests.

**Main ask: trim at ingest rather than at the lookup.** Details inline, but the short version is that `.trim()` inside `detectSolixCapabilities` fixes this one call site while `identity().category` and `buildModelIndex` keep handing callers the untrimmed string — and #187 lands two more exact-match comparisons on that same category, which I confirmed break when the two branches are merged together. One trim in `buildModelIndex` fixes all four places and decouples the two PRs.

Two smaller things inline: the added `//` blocks want to be JSDoc, and I'd like a source for the `A17E` / `AE11` prefixes.

For ordering: this merges cleanly with all three of the others, so it isn't blocked on anything — but #186 before #184 still matters (see my note there).

---

src/model/capabilities/solix.ts
max246: Done (822af5c) — moved the `.trim()` into `buildModelIndex` at ingest, so `identity().category`, `detectSolixCapabilities`, and any exact-match category test all get the clean name. The lookup is back to `CATEGORY_CAPABILITIES[category]` and the trim comment is gone. Added a `buildModelIndex` trim test, and the `identity().category` spec now asserts the trimmed value — which also unblocks #187's `CATEGORY_FAMILY`/`isSolixSolarbank` comparisons.

src/model/capabilities/solix.ts
max246: Done (822af5c) — both `//` blocks are JSDoc now. The `CATEGORY_CAPABILITIES` "Plug-in Home Battery" entry note is a JSDoc block above the entry with the #186 cross-link; the `detectSolixCapabilities` trim comment is gone entirely (trim moved to ingest).

src/model/capabilities/solix.ts
max246: Grounded against the live `product_categories` catalog on the account (822af5c): `A17C0`–`A17C5` and `AE100` all list under `"Plug-in Home Battery "`, confirming `A17C`/`AE10` (AE103 by spec). `A17E` isn't in the catalog at all, and the only `AE11x` product is `AE113` "XE 6/8kW" — a Residential Storage System, a different unverified family — so I dropped both `A17E` and `AE11` rather than over-grant battery/solarInput; the JSDoc now states the catalog grounding and why. (Side find from the same catalog: `AE1R0` = "Anker SOLIX P1 Meter" under "Accessory" — a real meter `SOLIX_METER_MODELS` (AE1X0 only) doesn't yet detect; noting as a separate follow-up, out of scope here.)

martijnpoppen (commented):
Verified `822af5c`. **All three addressed, and the catalog grounding is a better answer than I asked for.**

**The trim at ingest clears the #187 bug with no change needed there.** That was the part worth checking rather than taking on trust, so I merged this branch into #187 and re-ran the exact repro from my comment over there:

```
before:  category = "Plug-in Home Battery " | family = unknown   | caps = identity,battery,solarInput,acOutput
after:   category = "Plug-in Home Battery"  | family = solarbank | caps = identity,battery,solarInput,acOutput
```

So `CATEGORY_FAMILY` and `isSolixSolarbank` now match on their own terms, and `site.solarbanks()` stops silently dropping the device. One trim, four call sites fixed, and the two PRs stay decoupled.

**Dropping `A17E`/`AE11` rather than keeping them is the right call**, and checking them against the live `product_categories` catalog is exactly the grounding `AGENTS.md` asks for. `AE113` being a Residential Storage System rather than a Solarbank is a good catch — over-granting `battery`/`solarInput` to a different family is precisely the false positive the prefix list can't recover from at runtime.

The `AE1R0` side-find is worth the follow-up issue: `SOLIX_METER_MODELS` is `["AE1X0"]`, so a P1 Meter gets no `energyMeter` and its frames go unnamed. Agreed it's out of scope here. Worth flagging when you file it that #184 now has `SOLIX_METER_PRODUCT_PREFIXES` transport-side as the documented lockstep partner — a second meter prefix has to land in both or the decode silently won't follow.

**Verified locally** on `822af5c`: `tsc --noEmit` clean, `prettier --check` clean, all five guards pass, 198 files / 3596 tests. CI green too.

One leftover, genuinely trivial — `solix.ts:123` picked up a new `//` body comment where the two old ones were removed:

```ts
// The category is already trimmed at ingest (buildModelIndex), so an exact lookup is safe here.
for (const c of (category && CATEGORY_CAPABILITIES[category]) || []) caps.add(c);
```

Same rule as before; it belongs in `detectSolixCapabilities`' JSDoc ("…reads the catalog category, which {@link buildModelIndex} has already trimmed, so the lookup is exact"). Not worth a push of its own — fold it into whatever touches this file next.

Nothing else outstanding from me on this PR.

---

martijnpoppen (approved):

PR #186, opened 2026-09-16
fix(solix): gate meter tag→name table to meter-family frames

max246:
## What
The ff09 tag→name table `SOLIX_METER_FIELD_NAMES` (`0xac` = `meterVoltageL1`) is **meter-specific** — the same tag carries a different quantity on another Solix device. On a live **Solarbank 4 E5000 Pro (AE103)**, tag `0xac` reads a power value, so `solixReadings` was emitting `meterVoltageL1 = -400`: nonsense for a battery.

## Fix
- `solixReadings(frame, productCode?)` now applies the meter names **only** to a frame from the meter family (`SOLIX_METER_PRODUCT_PREFIX = "AE1X0"`); `onMessage` threads the product code from the telemetry topic.
- Every device's tags still surface raw as `channel_<hex>` — the model layer (`energyMeter()` members) names them per capability, gated on `has("energyMeter")`.
- With no `productCode` the names are withheld (the table cannot be known to fit).
- Corrects the reserved-tag note: `0xb2` is **not** `meterCurrentTotal` (a live single-phase frame read ~0.007 there while L1 current was tens of mA).

## Evidence
Live capture (2026-09-16) of an AE103 Solarbank + AE1X0 meter on one account: the Solarbank streams tag `0xac` = `-400`, the meter streams `0xac` = `240` (nominal mains). Product-code prefix is a decode-table selector, not a model import — `transport ⊥ model` is untouched.

## Tests / gates
- New spec: a non-meter (`AE103`) frame does not borrow the meter table; a bare call withholds names.
- `vitest` (solix mqtt/watch/device), `typecheck`, all 5 guards, and the docs build pass.


src/transport/mqtt/solix-mqtt.ts
```diff
@@ -101,16 +116,21 @@ export function decodeSolixParamFrame(buf: Buffer): SolixParamFrame | null {
...
  * param and contributes nothing. Each measurement is emitted under `channel_<hex tag>`, and additionally
- * under its name when the tag has a confirmed one in {@link SOLIX_METER_FIELD_NAMES}.
+ * under its name when the tag has a confirmed one in {@link SOLIX_METER_FIELD_NAMES} AND the frame is
+ * from the meter family — pass `productCode` (from the telemetry topic) so a non-meter device's tags
+ * stay raw `channel_<hex>` rather than borrowing the meter's tag→name table (see
+ * {@link SOLIX_METER_PRODUCT_PREFIX}). With no `productCode` the names are not applied, since the table
+ * cannot be known to fit.
  */
-export function solixReadings(frame: SolixParamFrame): Record<string, number> {
+export function solixReadings(frame: SolixParamFrame, productCode?: string): Record<string, number> {
```
martijnpoppen, line 125: **Make `productCode` required rather than optional.**

`solixReadings` is a public export (re-exported through `transport/mqtt/index.ts`, and named in its barrel doc). With the parameter optional, an existing external caller doing `solixReadings(frame)` keeps compiling and silently stops receiving `meterVoltageL1` — the exact failure mode this PR exists to prevent, just relocated from "wrong name" to "no name".

Making it required turns that into a compile error, which pre-1.0 is the outcome you want (`AGENTS.md`: "No backward-compat below 1.0. Remove speculative and dead code rather than keeping old shapes beside new ones"). It also removes the "with no `productCode` the names are not applied" branch from the contract entirely, so the JSDoc gets shorter rather than longer.

The only internal call site already has the value to hand (`parts[2]`), and the spec's third assertion (`solixReadings(decodeSolixParamFrame(FRAME)!)` with no code) becomes `solixReadings(frame, "")`, which still exercises the not-a-meter path.

---

src/transport/mqtt/solix-mqtt.ts
```diff
@@ -45,21 +45,36 @@ export interface SolixParamFrame {
...
 };
 
+/**
+ * Product-code prefix of the Smart Meter family that {@link SOLIX_METER_FIELD_NAMES} decodes. The table
+ * is meter-specific, so {@link solixReadings} applies its named fields ONLY to a frame from this family;
+ * a Solarbank (`AE103`) reporting the same `0xac` tag would otherwise be mislabelled `meterVoltageL1`
+ * with a nonsensical (negative-power) value. This is a product-code prefix used to select a decode
+ * table — not a model import — so the `transport ⊥ model` rule is untouched.
+ */
+export const SOLIX_METER_PRODUCT_PREFIX = "AE1X0";
```
martijnpoppen, line 76: Agreed with the reasoning in the JSDoc — decorrelation forbids sharing this with `model/`, so duplicating it here is correct, not a smell. One drift hazard worth noting though:

`model/capabilities/solix.ts` holds `SOLIX_METER_MODELS: readonly string[] = ["AE1X0"]` — an **array**, matched with `.some(m => code.startsWith(m))`. This is a **single string**. They agree today, but the shapes have different capacity: the moment a second meter prefix is added model-side (`["AE1X0", "AE1R0"]`, say — `AE1R0` is already named in #185's doc), `detectSolixCapabilities` will grant `energyMeter` to a device whose frames this decoder refuses to name, and nothing fails.

No guard can catch that (a `model/` spec can't import `transport/`, which is the point). Cheapest mitigation is symmetry plus a pointer: make this `readonly string[]` too, and have each side's JSDoc name the other as the thing to update in lockstep. Then at least the diff that adds a prefix looks incomplete.

---

src/transport/mqtt/solix-mqtt.ts
```diff
@@ -373,12 +393,15 @@ export class SolixMqtt extends EventEmitter {
...
+    const productCode = parts[2] ?? "";
     const reading: SolixReading = {
       deviceSn: frame.deviceSn ?? parts[3] ?? "",
-      productCode: parts[2] ?? "",
+      productCode,
       topic,
       frame,
-      values: solixReadings(frame),
+      // Pass the product code so meter tag→name binding is applied only to a meter frame; a Solarbank's
+      // tags stay raw channel_<hex> (the model names them per capability) rather than being mislabelled.
```
martijnpoppen, line 403: `AGENTS.md`: **"no inline body comments"** — and this one is redundant twice over, since `solixReadings`' own JSDoc now states the gating rule and `SOLIX_METER_PRODUCT_PREFIX`'s states the reason. The call reads fine bare:

```ts
values: solixReadings(frame, productCode),
```

---

martijnpoppen (commented):
Reviewed this alongside #184, #185 and #187 (all four checked out locally). **This is the strongest of the four and I'd land it first** — the reasoning is right, the fix is minimal, and the spec pins both the meter and non-meter paths.

**Verified locally** on `504a3bb`: `tsc --noEmit` clean, `prettier --check` clean, all five guards pass (including `decorrelation` — the product-code prefix is not a model import, as the JSDoc argues), suite green at 198 files / 3594 tests.

### Merge order — this one matters

**#184 and #186 conflict**, and the order isn't just textual. I tested every pair of the four Solix PRs locally; #184 + #186 is the only conflicting pair:

```
PR 184 + PR 186: CONFLICT
PR 184 + PR 185: clean    PR 185 + PR 186: clean
PR 184 + PR 187: clean    PR 185 + PR 187: clean    PR 186 + PR 187: clean
```

Both rewrite the `SOLIX_METER_FIELD_NAMES` doc block and the same spec case. More importantly: **#184 grows that table from 1 entry to 12 without this gate**. Landed on its own, a Solarbank frame would pick up twelve wrong meter names instead of one — it multiplies by twelve exactly the mislabelling this PR exists to stop. So: **this PR first, then rebase #184 onto it.** Worth saying so in #184's description so the ordering doesn't get lost.

Three comments inline — one I'd act on before merge (the optional parameter is a silent behaviour change for external callers), one drift hazard between this and `SOLIX_METER_MODELS`, one comment-style nit.

---

max246:
Folded into #184 (commit 50bdbd3) so the 12-name meter table and this gate land as one change — they conflict on the same file/spec, and #184 without this gate would mislabel twelve channels on every Solarbank frame, so they're really one unit.

All three review points here were addressed in the fold:
- **`productCode` is now required** on `solixReadings` (not optional) — pass `""` for a frame of unknown origin; no names apply.
- **`SOLIX_METER_PRODUCT_PREFIXES` is a `readonly string[]`** mirroring model-side `SOLIX_METER_MODELS`, and each side's JSDoc names the other as the lockstep partner (the drift hazard).
- **Inline call-site comment dropped.**

Closing in favor of #184. Thanks for the review — the required-param call was the right one.

PR #187, opened 2026-09-16
feat(solix): product-family classification + SolixSite system aggregate

max246:
Adds the Solix analogue of eufy's **HomeBase** modelling — the "can we categorize a power station like we do a HomeBase" ask — in two parts.

## 1. Product-family classification (`model/solix-family.ts`)
Pure, param-agnostic predicates over product code + catalog category — the Solix counterpart of `device-family.ts` `isHomeBase()`, answering *what KIND of device is this*, kept separate from capability resolution:

- `solixProductFamily(input)` → `powerStation | solarbank | smartMeter | powerBank | cooler | evCharger | charger | unknown` (never a guess)
- `isSolixPowerStation` / `isSolixSolarbank` / `isSolixSmartMeter`
- `SolixDevice.family` (and `identity().family`)

It **reuses** the existing model-code sets (`SOLARBANK_MODELS` / `SOLIX_METER_MODELS`) as the single source of truth rather than duplicating them.

## 2. `SolixSite` system aggregate (`model/solix-site.ts`)
The "My Home" energy system as a first-class **grouping** of its member devices — the HomeBase grouping role:

- `discoverSolixSites(client)` composes three wire reads (sites + devices + catalog) entirely model-side, via a **structural** `SolixSiteReader` (so `transport ⊥ model` holds — proven by a hand-written fake in the spec), resolving each site member to its full `SolixDevice`.
- Grouping accessors: `solarbanks()` / `smartMeters()` / `powerStations()` / `withFamily()` / `withCapability()` / `device(serial)`.
- The realtime **system-flow** aggregate (instantaneous SoC/solar/grid) is deliberately **not** modelled: there is no confirmed, stable read surface for this device generation, so no fabricated getter — callers read each member's telemetry. (Honest per the hard "never ship a guessed getter" rule.)

## Transport / core
- `getSites()` retyped `unknown[]` → `SolixSiteRecord[]`; new core types `SolixSiteRecord` / `SolixSiteDeviceEntry`.

## Scope / interaction with open PRs
Purely **additive** — does **not** touch `detectSolixCapabilities`, so it does not collide with #185 (Solarbank detection). Because the family module reads `SOLARBANK_MODELS`, it picks up #185's widened set automatically once #185 lands (until then, on `beta-0.2.0` alone, an `AE103` still classifies via its catalog category `"Plug-in Home Battery"`).

## Evidence
Live account probe (2026-09-16): an **AE103 Solarbank 4 E5000 Pro** + **AE1X0 Smart Meter Gen 2** grouped under one **"My Home"** site (`power_site_type` 20).

## Gates
`vitest` (3604 pass, incl. new `solix-family` + `solix-site` specs), `typecheck`, all 5 guards (decorrelation/lines/docrefs/consumer-agnostic/capability-ownership), and the docs build all pass. No real device data in the diff.


src/model/solix-family.ts
```diff
@@ -0,0 +1,84 @@
...
+  /** Anker catalog category, when a catalog resolved one (e.g. `"Portable Power Station"`). */
+  category?: string;
+}
+
+/** The catalog categories that map one-to-one onto a family, for the codes a prefix set does not cover. */
+const CATEGORY_FAMILY: Readonly<Record<string, SolixProductFamily>> = {
+  "Portable Power Station": "powerStation",
+  "Plug-in Home Battery": "solarbank",
+  "Power Bank": "powerBank",
+  "Powered Cooler": "cooler",
```
martijnpoppen, line 58: **Bug: these exact-match category keys miss the form the live catalog actually returns.**

#185 documents that the catalog returns `"Plug-in Home Battery "` — with a trailing space — and adds a spec asserting `identity().category` keeps it. Both this table and `isSolixSolarbank` below compare with `===`, so neither matches.

I merged this branch with #185 and ran a catalogued plug-in home battery whose product code isn't in `SOLARBANK_MODELS`:

```
category = "Plug-in Home Battery " | family = unknown | caps = identity,battery,solarInput,acOutput
```

Capabilities resolve (#185's trim covers that path), but the family doesn't — so `site.solarbanks()` silently omits a device the system's own capability detection just classified as a home battery. That's the worst shape for this bug: no error, no empty result, just one device quietly missing from a grouping accessor.

`AE103` is masked by #185 also adding the `AE10` prefix, so this won't show up in the current tests.

The clean fix is in #185 — trim once in `buildModelIndex` where the category is read off the catalog — and then this table needs no change at all. I've left the same note there. If you'd rather not couple the PRs, `input.category?.trim()` in `solixProductFamily` and `isSolixSolarbank` closes it locally, but then `identity().category` still leaks the untrimmed string to callers.

---

src/model/solix-family.ts
```diff
@@ -0,0 +1,84 @@
...
+ * catalog-independent, so a device classifies even before a catalog is fetched); the catalog category
+ * fills in the families that have no prefix set yet. Returns `"unknown"` when neither identifies one —
+ * never a guess, mirroring `device-family.ts` returning `false` on an unknown device type.
+ */
+export function solixProductFamily(input: SolixFamilyInput): SolixProductFamily {
+  // Meter before Solarbank: their catalog category can both be battery-ish, but the meter's AE1X0 prefix
+  // is unambiguous, and a meter is never a Solarbank.
+  if (isSolixSmartMeter(input)) return "smartMeter";
+  if (hasPrefix(input.product_code, SOLARBANK_MODELS)) return "solarbank";
+  return (input.category ? CATEGORY_FAMILY[input.category] : undefined) ?? "unknown";
```
martijnpoppen: Two things here.

**1. Re-implements `isSolixSolarbank` instead of calling it.** Line 82 repeats the prefix half of the predicate exported sixteen lines above, so the two can drift — and a fix to one (the trailing-space issue in my other comment, say) won't reach the other. Since `isSolixSmartMeter` is already checked first, calling the predicate is exactly equivalent:

```ts
if (isSolixSmartMeter(input)) return "smartMeter";
if (isSolixSolarbank(input)) return "solarbank";
return (input.category ? CATEGORY_FAMILY[input.category] : undefined) ?? "unknown";
```

The `CATEGORY_FAMILY["Plug-in Home Battery"]` entry then becomes dead for the solarbank case (it's already covered by the predicate's own category check), which is worth noticing on its own — right now the same category-to-solarbank mapping is stated in two places.

`AGENTS.md`, "Reuse before abstraction": *"Reuse or extend an existing primitive when it owns the same operation and satisfies the required semantics."*

**2. The `//` block is an inline body comment** (hard rule). The ordering rationale is genuinely worth keeping — move it into the function's JSDoc, which already explains the precedence and just needs the "and a meter is never a Solarbank" sentence.

---

src/transport/http/solix-client.ts
```diff
@@ -358,10 +359,15 @@ export class SolixClient {
...
+  /**
+   * The account's sites (systems); devices are grouped under a site. Each record carries its
+   * `site_device_list` (the member devices), which {@link discoverSolixSites} resolves into a
+   * capability-driven `SolixSite`. Asserted to {@link SolixSiteRecord} at this trust boundary — every
+   * field beyond `site_id` is optional on the record, so a caller reads them defensively.
+   */
+  async getSites(): Promise<SolixSiteRecord[]> {
     const data = await this.authed<{ site_list?: unknown[] }>("POST", SOLIX_ENDPOINTS.getSiteList, {});
-    return data?.site_list ?? [];
+    return (data?.site_list ?? []) as SolixSiteRecord[];
```
martijnpoppen: The JSDoc says "every field beyond `site_id` is optional on the record, so a caller reads them defensively" — but `site_id` is precisely the field the cast promises and nothing checks, and it's the one the model layer then trusts unconditionally:

```ts
// solix-site.ts
readonly id: string;
this.id = record.site_id;                     // undefined if the cloud omits it
this.name = record.site_name || record.site_id; // undefined too
```

A site record without `site_id` gives you `SolixSite.id === undefined` sitting behind `readonly id: string`, and `site.device(sn)` lookups against a `Map` keyed on it. `AGENTS.md` puts validation exactly here: *"Trust-boundary validation and independently authoritative declarations remain local."*

Cheapest fix that keeps the type honest — filter at the boundary:

```ts
const list = (data?.site_list ?? []) as SolixSiteRecord[];
return list.filter((s) => typeof s?.site_id === "string" && s.site_id.length > 0);
```

Then the JSDoc's claim is true as written. The alternative — making `site_id` optional on `SolixSiteRecord` — pushes the problem into `SolixSite`, which would then need `id?: string`, and that's worse for callers.

---

src/model/solix-site.ts
```diff
@@ -0,0 +1,119 @@
...
+    client.getSites(),
+    client.getDevices(),
+    opts.catalog ? Promise.resolve(opts.catalog) : client.getProductCatalog().catch(() => [] as SolixProductCategory[]),
+  ]);
+  const bySerial = new Map(records.map((r) => [r.device_sn, r]));
+  return sites.map((site) => {
+    const members = (site.site_device_list ?? []).map((entry) => {
+      // Prefer the account's full device record; fall back to a minimal record from the site entry so a
+      // declared member is never dropped for being absent from the flat device list.
+      const record: SolixDeviceRecord = bySerial.get(entry.device_sn) ?? {
```
martijnpoppen, line 108: Two reuse notes, both small.

**The catalog-resolution ternary on line 103 is duplicated verbatim** from `discoverSolixDevices` (`solix-device.ts:191`) — same expression, same `.catch(() => [])` fallback, same `opts.catalog` shortcut. Worth one shared helper in `solix-device.ts` next to the reader interface:

```ts
const resolveCatalog = (client: SolixDeviceReader, opts: { catalog?: SolixProductCategory[] }) =>
  opts.catalog ? Promise.resolve(opts.catalog) : client.getProductCatalog().catch(() => [] as SolixProductCategory[]);
```

Both discovery functions then read the same way, and the "a failed catalog degrades to no categories" decision lives in one place instead of two.

**The `//` block on line 108** is an inline body comment (hard rule), and it restates this function's own JSDoc nearly verbatim — "A site member is resolved to its FULL device record from `getDevices()` when the account lists one … a member the flat device list omits is built from the site entry alone". The JSDoc already says it better; just delete the inline copy.

---

src/model/index.ts
```diff
@@ -133,3 +133,15 @@ export {
...
+export {
+  solixProductFamily,
+  isSolixPowerStation,
+  isSolixSolarbank,
+  isSolixSmartMeter,
+  type SolixProductFamily,
+  type SolixFamilyInput,
+} from "./solix-family.js";
+export { SolixSite, discoverSolixSites, type SolixSiteReader, type SolixSiteOptions } from "./solix-site.js";
+export type { SolixSiteRecord, SolixSiteDeviceEntry } from "../core/solix-types.js";
```
martijnpoppen: Nit: `core/index.ts:8` already does `export * from "./solix-types.js"`, so both of these reach `src/index.ts` through the core barrel. Re-exporting them from the model barrel is redundant.

It's harmless — both star-exports resolve to the same declaration, so there's no ambiguity and I confirmed `tsc --noEmit`, `build` and `check:esm` are all green — but `AGENTS.md` describes `src/index.ts` as "one `export *` per layer barrel", and a type owned by `core/` travelling out through two of them muddies which layer owns it. `SolixDeviceRecord` and `SolixProductCategory` are already handled the other way (core-only), so this is also inconsistent with the file's existing convention.

---

martijnpoppen (commented):
Reviewed alongside #184, #185 and #186 (all four checked out locally).

The structure here is good. Splitting "what kind of thing is this" (family) from "what can it do" (capabilities) is the right cut and the module doc argues it well; typing `SolixSiteReader` structurally keeps `transport ⊥ model` intact without ceremony; and I appreciate that the site aggregate explicitly declines to model the realtime system totals rather than fabricating a getter for them — that's the house rule applied without being asked.

**Verified locally** on `14b6acd`: `tsc --noEmit` clean, `prettier --check` clean, all five guards pass (`decorrelation` included), suite green at 200 files / 3604 tests.

**One real bug**, and it's an interaction with #185 rather than anything visible in this diff alone: `CATEGORY_FAMILY` and `isSolixSolarbank` compare the catalog category with `===`, but #185 establishes that the live catalog returns `"Plug-in Home Battery "` with a trailing space. Merging the two branches and running it gives a device with correct capabilities and `family: "unknown"` — so `site.solarbanks()` drops it silently. Repro and both fix options are inline; the clean one lives in #185.

The rest are smaller: a trust-boundary cast that doesn't validate the one field it promises, `solixProductFamily` re-implementing the predicate it exports, a duplicated catalog-resolution ternary, two inline body comments, and a redundant re-export.

Ordering: this merges cleanly against all three siblings, so nothing blocks it — but it wants #185 to land first (or to carry its own trim), otherwise the family bug ships.

---

martijnpoppen (commented):
Closing my main finding here — **it's fixed by #185, with nothing to change on this branch.**

`822af5c` over there moves the `.trim()` into `buildModelIndex` at ingest, so the catalog category reaches this PR already clean. I merged that branch into this one and re-ran the repro from my earlier comment:

```
before:  category = "Plug-in Home Battery " | family = unknown
after:   category = "Plug-in Home Battery"  | family = solarbank
```

`CATEGORY_FAMILY` and `isSolixSolarbank` match on their own terms now, and `site.solarbanks()` no longer drops a catalogued home battery whose product code isn't in `SOLARBANK_MODELS`. No `input.category?.trim()` needed here — leaving these as exact comparisons is right once the string is clean at the source.

That does make **#185 a merge-order dependency for this PR**: landing this one first reintroduces the family gap until #185 follows.

Still open from my review, unchanged:
- `getSites()` casts to `SolixSiteRecord[]` without validating `site_id`, the one field the type marks required and `SolixSite` then trusts behind `readonly id: string`.
- `solixProductFamily` re-implements the Solarbank prefix test instead of calling the `isSolixSolarbank` it exports above it.
- The duplicated catalog-resolution ternary in `discoverSolixSites`, two inline `//` body comments, and the redundant `model/index.ts` re-export.

---

src/model/solix-family.ts
max246: Handled the clean way you pointed to: #185 now trims the category once at ingest in `buildModelIndex`, so `input.category` here is already the clean form — no trailing-space handling needed in this table (f0eb139). And with the reuse fix below, the `"Plug-in Home Battery"` -> solarbank mapping now lives only in `isSolixSolarbank`, so I removed the (now dead) `CATEGORY_FAMILY` entry too. Merge-order: this wants #185 first (noted on the PR), otherwise the untrimmed category would miss.

src/model/solix-family.ts
max246: Done (f0eb139). `solixProductFamily` now calls `isSolixSolarbank(input)` instead of re-testing the prefix — since `isSolixSmartMeter` is checked first they're equivalent, and the two can't drift. That makes the `CATEGORY_FAMILY["Plug-in Home Battery"]` entry dead (the predicate already covers the category), so I removed it — the solarbank category mapping is now stated once. The `//` ordering rationale moved into the function JSDoc (with the 'a meter is never a Solarbank' sentence).

src/transport/http/solix-client.ts
max246: Fixed at the boundary (f0eb139): `getSites()` now `.filter((s) => typeof s?.site_id === "string" && s.site_id.length > 0)`, so a record without a usable id is dropped rather than becoming a `SolixSite` with `id === undefined`. Updated the JSDoc so the claim is true as written, and added a spec asserting the missing/empty-id records are filtered.

src/model/solix-site.ts
max246: Both done (f0eb139). Extracted `resolveSolixCatalog(client, opts)` next to `SolixDeviceReader` in solix-device.ts; `discoverSolixDevices` and `discoverSolixSites` both call it, so the 'a failed catalog degrades to no categories' decision lives in one place. Deleted the inline `//` block that restated the JSDoc.

src/model/index.ts
max246: Removed (f0eb139). `SolixSiteRecord` / `SolixSiteDeviceEntry` are core-owned and reach `src/index.ts` through the core barrel's `export * ./solix-types`, so the model-barrel re-export is gone — consistent with how `SolixDeviceRecord` is handled. Left a one-line note where it was so the omission reads as intentional.

max246:
All six review points addressed in f0eb139 (replies inline, threads resolved):
- `solixProductFamily` reuses `isSolixSolarbank` (no re-implementation); the now-dead `CATEGORY_FAMILY["Plug-in Home Battery"]` entry dropped; ordering `//` moved to JSDoc.
- `getSites()` validates `site_id` at the boundary (+ spec) so no `SolixSite` with `id === undefined`.
- `resolveSolixCatalog()` shared by both discovery functions; duplicated ternary + inline `//` gone.
- Redundant model-barrel re-export of the core-owned site types removed.

**Merge order:** the trailing-space family bug is fixed by #185's ingest trim (`buildModelIndex`), not locally here — so this wants **#185 to land first**. Full verify green (3605 tests).

martijnpoppen (commented):
Verified `f0eb139`. **All six findings addressed — nothing outstanding from me on this PR.**

Checked each against the tree rather than the commit message:

- **Predicate reuse** — `solixProductFamily` calls `isSolixSolarbank(input)`, and the ordering rationale (including "a meter is never a Solarbank") is in the function JSDoc. ✅
- **`CATEGORY_FAMILY` dead entry removed**, and the JSDoc now says *why* the home-battery category is deliberately absent rather than leaving a future reader to wonder if it was an oversight. That's the better half of this fix: the solarbank category mapping is now stated once, in the predicate that owns it. ✅
- **`site_id` validated** at the boundary — `list.filter((s) => typeof s?.site_id === "string" && s.site_id.length > 0)`, so `readonly id: string` is now true rather than asserted. The JSDoc says what's dropped and why. ✅
- **Catalog resolution deduped** into `resolveSolixCatalog`, exported from `solix-device.ts` and consumed by `discoverSolixSites`. The doc naming the shared decision ("a failed catalog is non-fatal lives in exactly one place") is exactly the reason to extract it. ✅
- **Inline `//` body comments**: zero across `solix-family.ts`, `solix-site.ts` and `solix-device.ts`. ✅
- **`model/index.ts` re-export removed**, replaced by a comment explaining the core types leave through the core barrel — matching how `SolixDeviceRecord` is handled. Better than silently deleting the lines. ✅

**Verified locally** on `f0eb139`: `tsc --noEmit` clean, `prettier --check` clean, all five guards pass, 200 files / 3605 tests.

Merge-order still stands, and it's now load-bearing in a new way: with the `CATEGORY_FAMILY` entry gone, the home-battery category is matched only by `isSolixSolarbank`'s exact `=== "Plug-in Home Battery"`. Landing this before #185 means the untrimmed catalog string misses *both* paths, so a catalogued Solarbank outside `SOLARBANK_MODELS` would come back `unknown` with no fallback. #185 first, then this.

---

martijnpoppen (approved):

PR #190, opened 2026-09-16
fix(session): stop reading a gtoken mismatch as an expired token

martijnpoppen:
## What changed

Two fixes to the same live failure, plus the log line that measures how often it bit.

**`tokenRejected()` no longer matches the gateway's gtoken-mismatch message.** Two independent
properties of the match were missing, and restoring either alone leaves it matching:

- *A wildcard could cross a clause boundary.* `token = <echo>, gtoken not equal userid error` anchored
  on the echoed credential's own `token` and reached past the comma for an `error` belonging to the next
  clause. Each alternative's `.*` is now `[^,]*`, so a wildcard stays inside the clause its anchor sits
  in.
- *`gtoken` contains `token`.* Bound the wildcard alone and the match simply re-anchors inside the other
  header's name, so `token` is now matched as a whole word.

Both are bounds on the match rather than edits to the message. Removing the echo before classifying —
the first approach here — loses real signal: for any wording that states its reason beside the echo
(`token = <echo> expired`) the anchor *is* the echo, and a genuinely dead session then never gets
cleared.

**`gtoken` is hashed from the account's own `user_id`.** The login reply carries two ids and the client
prefers `ap_cloud_user_id` for `auth_.userId` — the Anker Passport cloud's id, while `/app/*`
authenticates against `user_id`.

Flipping that preference is the obvious fix and the wrong one: `auth_.userId` also feeds the Tuya
binding, the P2P acting-user fallback and stored-image partitioning, and it is persisted. None of those
asked for a different id. So the account id is carried alongside as `accountUserId`, and only the one
header the gateway checks is pinned to the one id it checks against, read through a single accessor so
the two header paths cannot drift.

**A record stored before `accountUserId` existed is refused rather than restored.** Hydrating one
reinstates a session whose header is hashed from the other id, and nothing in a restored session can
recover the right one — it arrives with a login reply. Combined with the classifier fix, which stops
that 401 clearing the session, such an install would fail identically forever. `isSessionValid` now
treats the account id as part of the credential and `persist()` always writes it, so absence means a
pre-field record rather than a reply that carried one id. The stored device identity is still reused, so
the one login this costs is not a new device.

## Why

A sign-in succeeds and the first authenticated call fails with
`401: token = <echo>, gtoken not equal userid error`. The gateway recomputes `gtoken` from the user its
token belongs to and rejects a disagreement — a **header** fault on a token that is otherwise perfectly
good. The classifier read it as a dead credential, cleared a healthy session and re-logged in, which
cannot change a header. On an account with 2FA that spends a verification code while the person is still
typing the previous one; without 2FA it re-authenticates silently and the caller simply never receives a
device list.

Where they sit in the call path is why both are worth fixing separately: the id fix is what makes the
call succeed, and the classifier fix is what stops one refused call from destroying the session behind
it. Either alone leaves a real failure.

## Evidence

A host integration's v0.8.0 user-diagnostics report — an EU account routed correctly to `eu-pr`,
`fa_info.step: 0` so no challenge raised, signed in first try, then:

```
[session] → ok
[session] expired, re-authenticating /app/house/get_house_list failed (401/401):
          token = <redacted>, gtoken not equal userid error   ← 1.0s later
[session] → reauth
```

The gateway's own error string is what names the mismatch, and `gtoken` is `md5` of the id held in
`auth_.userId` — for the two to disagree, that id has to be something other than the user the token is
bound to, and the `??` chain is the only thing that can make it so.

Diagnosed in the host integration's investigation notes, updated alongside this in its own pull request.

## Verified how

- [x] `npm run verify` is green locally — 3602 tests / 199 files, plus format, typecheck, all five
      guards, build, ESM check, doc snippets, examples typecheck. Run on this branch rebased onto
      `beta-0.2.0`.
- [ ] Ran against a real device — **no.** See below.
- [x] Added or updated specs (offline, synthetic fixtures) — `gtoken-header.spec.ts` for the header and
      the refused-restore path, two cases in `session-expiry.spec.ts` for the classifier. Each was
      checked against the unfixed code rather than assumed: reverting the id pin fails two, reverting
      the `isSessionValid` guard fails two, and restoring the strip approach fails the
      `expired`-beside-echo case.
- [x] Docs updated — the investigation lives in the host integration's repo, in its own pull request.

## Unverified or deferred

**Neither fix has been driven against a live account.** The classifier is fully exercised offline. The
header change is verified as byte-identical where the reply's two ids agree, and as correct by the
gateway's own error string where they do not, but it has not been observed repairing a live 401.

**Whether any account's two ids actually differ is still uncaptured.** The login debug logged key names,
not values, so the report above proves both fields are *present* and not that they disagree. The new
`ids agree:` line answers that on the next report. The fix is shaped to be safe ahead of the answer:
where the ids agree the header is exactly the string it has always been, and there is a test pinning
that.

**The same `ap_cloud_user_id` chain exists in `solix-client.ts` and is recorded, not switched.** No
response from that gateway has been observed refusing the header, so a change there would rest only on
the shape looking alike — the guess this change is specifically trying not to be. Noted on
`classifyLogin` as a wire fact; @max246 is taking the Solix half separately, starting with the debug
line that would turn the resemblance into evidence.

**One thing deliberately not changed.** The hold-off warning that blames two clients sharing a device
identity (`CONTENDED_SESSION_HINT`) misfired on this failure and advised an `openudid` the caller
already had. It needed no edit: it is reached only through `recoverRejectedSession`, which is reached
only when `tokenRejected` says the credential is finished, so it stops firing for a cause it never
fitted. For the repeated displacement it was written about, the wording is accurate.

https://claude.ai/code/session_01JrCBqDsVPx4V9FTHHmXiaq


max246 (commented):
Reviewed this end-to-end — the root-cause reasoning is solid and the change is pleasingly conservative (byte-identical whenever the two ids agree, which is likely most accounts). I traced the classifier by hand: after `withoutTokenClause`, `"token = …, gtoken not equal userid error"` → `" gtoken not equal userid error"`, and `\btoken\b` correctly does **not** match inside `gtoken` (preceded by `g`, no boundary), so it returns `false` and the session is kept ✅. The `"token not exist, token = …"` fixture still classifies as rejected ✅. Both `gtoken(this.auth_.userId)` call sites (authed headers + media download) are migrated — no site missed.

Two small suggestions + one follow-up I'd like to see addressed:

**1. The header fix has no test.** Both new specs cover the *classifier*; the "makes the call actually succeed" half — that `gtoken` is now `md5(user_id)` when the reply's two ids differ — is untested, and it's the whole reason the PR exists. `session-expiry.spec.ts` fully mocks `login()`, so it's the wrong home; better to drive login-response parsing (or `gtokenHeader()`/`authTokenHeaders()` directly) with a reply where `ap_cloud_user_id !== user_id`, and assert (a) the outgoing `gtoken` header === `md5(user_id)`, and (b) `accountUserId` round-trips through persistence (save → load → header still uses `user_id`). That locks in the exact behavior the change is for.

**2. Minor false-negative edge in `withoutTokenClause`.** `\btoken\s*[=:]\s*\S+` uses greedy `\S+`, so a delimiter-less message like `token=abc&error=expired` would swallow `expired` and lose the real signal. Observed messages use `", "` (spaces) so it's low-risk today, but reusing the bounded token charclass from `withoutTokenEcho` (`[A-Za-z0-9._-]{8,}`) instead of `\S+` is safer and keeps the two helpers consistent.

**3. Follow-up (separate change): the same latent bug is in `solix-client.ts`.** `userId = ap_cloud_user_id ?? user_id` (line ~280) then `gtoken: gtoken(userId)` (line ~293) — identical shape. If a Solix account's two ids ever differ, its plain reads would 401 the same way. We do NOT yet have the Solix gateway's own "gtoken not equal userid" evidence (our current Solix reads work, so that account's ids presumably agree), so this shouldn't be a blind switch — but the same conservative fix applies (pin gtoken to `user_id`, byte-identical when the ids agree) plus the same `ap_cloud_user_id vs user_id: same/DIFFERENT` debug line to gather the evidence. Happy to open a separate PR for the Solix half so this one stays focused on the mega client.

Net: LGTM on the logic and the blast radius; I'd just like the header behavior pinned by a test (1) before merge.

src/core/store.ts
```diff
@@ -17,6 +17,12 @@ import type { RegionShard } from "../transport/http/mega-client.js";
  */
 export interface PersistedSession {
   userId: string;
+  /**
+   * The eufy account's own `user_id`, when the login reply carried one distinct from `userId` (which
+   * prefers the Anker Passport cloud's id). The `gtoken` header is hashed from this; absent in a record
+   * written before it was tracked, which falls back to `userId` as before.
+   */
+  accountUserId?: string;
```
lenoxys, line 27: **Blocker.** The fallback makes this a dead end for the only accounts it exists for. `isSessionValid` doesn't look at this field, so an install upgrading with a stored session hydrates with the Passport id and keeps sending the header the gateway rejects — and the classifier change in this same commit means that 401 no longer clears the session, so it never logs in fresh and never records the id. Before this commit the wrong classification at least churned into a login that would have fixed it.

Refuse to restore a record written before the field existed, or the fix only ever reaches installs that happen to sign in from scratch.

src/transport/http/mega-client.ts
```diff
@@ -241,12 +241,19 @@ const CONTENDED_SESSION_HINT =
...
+ *  - `token` is matched as a whole word, so `gtoken` — a different header with a different meaning — is not
+ *    read as the credential.
  */
 function tokenRejected(code: number | undefined, msg: string | undefined): boolean {
   return (
     code === EufyCloudErrorCode.SESSION_KICKED ||
-    /user_id is empty|invalid.*token|token.*(expired|error|not exist)|kicked|(?:token|session).*does not exist|unauthor/i.test(
-      msg ?? "",
+    /user_id is empty|invalid.*\btoken\b|\btoken\b.*(expired|error|not exist)|kicked|(?:\btoken\b|\bsession\b).*does not exist|unauthor/i.test(
+      withoutTokenClause(msg ?? ""),
```
lenoxys: **Blocker.** `\S+` isn't clause-bounded, so the strip eats past the echo and takes the anchor with it. Probed old-vs-new with `withoutTokenEcho` applied first, as the call site at 603 does: `token = <redacted>,token not exist` goes true → false. The strip swallowed the *second* `token`. `[^\s,]+` fixes that one.

What it doesn't fix, and is the design question: `token = <redacted> expired` also goes true → false, because once the clause is gone there is no `token` left to anchor on. The anchor now lives entirely in the reason clause, so any expiry whose reason is a bare adjective sitting after the echo is unclassified — and a session that really is dead never gets cleared. That's the more expensive direction of the same mistake this JSDoc warns about, and it should be a decision rather than something inherited from `\S+`.

src/transport/http/mega-client.ts
```diff
@@ -262,6 +269,19 @@ function withoutTokenEcho(text: string): string {
...
+/**
+ * Drop the echoed `token = …` clause entirely — including the `token = <redacted>` placeholder that
+ * {@link withoutTokenEcho} leaves behind. For classification only; see {@link tokenRejected}.
+ *
+ * Keeping a readable placeholder is right for a message a person reads and wrong for one a pattern matches:
+ * the placeholder still contains the word `token`, so it supplies an anchor for a wildcard that then spans
+ * the rest of the message. The clause never explains anything — it is the credential — so the explanation is
+ * what is left once it is gone.
+ */
+function withoutTokenClause(text: string): string {
```
lenoxys: **Cut.** Two functions for one fact, and they already disagree about what an echoed token looks like: eight-char floor plus optional quotes above, anything non-space here. `token:abc expired` is stripped by this one and left intact by the other. A replacement argument on the existing function serves both callers and kills the drift.

src/transport/http/mega-client.ts
```diff
@@ -428,15 +453,32 @@ export class MegaHttpClient {
   }
 
   /**
-   * The account-credential headers every authed call carries — `x-auth-token` + `gtoken` (`md5(userId)`).
+   * The `gtoken` header: `md5` of the eufy account's `user_id`. That id is the one the gateway recomputes
+   * the header from, and rejects a mismatch on — in its own words, `"gtoken not equal userid error"`.
+   *
+   * It is tracked separately from {@link auth_.userId} because the login reply carries TWO ids and the SDK
+   * prefers `ap_cloud_user_id` for `userId` — the Anker Passport cloud's id, not the eufy account's. That
```
lenoxys: **Nit, standards.** "the same string the header has always carried" and "persisted before this was recorded" narrate the fix — AGENTS.md is explicit that a JSDoc doesn't do that. "That preference is load-bearing elsewhere (the Tuya binding, the P2P acting-user fallback, stored-image partitioning)" is commit-message reasoning about three other modules.

What earns its place: the gateway's own wording, and which of the two ids it checks against. Two sentences.

src/transport/http/mega-client.ts
```diff
@@ -428,15 +453,32 @@ export class MegaHttpClient {
...
+   * preference is load-bearing elsewhere (the Tuya binding, the P2P acting-user fallback, stored-image
+   * partitioning), so it is left alone; only the one header the gateway checks is pinned to the one id the
+   * gateway checks it against. Where the two ids are equal — which may be every account — this is the same
+   * string the header has always carried.
+   *
+   * Falls back to `userId` for a session persisted before this was recorded.
+   */
+  private gtokenHeader(auth: { userId: string; accountUserId?: string }): string {
```
lenoxys: **Cut.** A one-line delegate whose only parameter is the object it hangs off, and both call sites pass `this.auth_`. `authTokenHeaders` eight lines below reads `this.auth_` directly with no parameter — same class, same field, opposite answer. Inline `gtoken(this.auth_.accountUserId ?? this.auth_.userId)` at both sites and put the two surviving sentences of doc on the `auth_` field.

src/transport/http/mega-client.ts
```diff
@@ -428,15 +453,32 @@ export class MegaHttpClient {
...
+    return gtoken(auth.accountUserId ?? auth.userId);
```
lenoxys: Nothing asserts this. No fixture in `src/` carries both ids, so neither the pin, the store round-trip, nor the "byte-identical where the ids agree" claim that justifies shipping ahead of the capture is checked. That claim is the whole safety argument for landing an unconfirmed hypothesis — it's the one line of test worth having.

src/transport/http/mega-client.ts
```diff
@@ -428,15 +453,32 @@ export class MegaHttpClient {
...
+   * The account-credential headers every authed call carries — `x-auth-token` + `gtoken`.
    * One place so the signed path, the key-exchange and the bearer path can't drift on what "authed" means.
    */
   private authTokenHeaders(): Record<string, string> {
     if (!this.auth_) return {};
     return {
       "x-auth-token": this.auth_.authToken,
       authorization: this.auth_.authToken,
-      gtoken: gtoken(this.auth_.userId),
+      gtoken: this.gtokenHeader(this.auth_),
```
lenoxys: The Solix client builds the same header off the same `ap_cloud_user_id ?? user_id` chain, and its own JSDoc says `md5(user_id)`. Untouched here, so the two clients now claim different truths about one header. Either it's the same defect and belongs in this commit, or it's a different gateway and that's worth one line somewhere.

lenoxys (commented):
Two fixes, and they undo each other on the upgrade path: a stored session has no `accountUserId`, so the header stays wrong, and the classifier change means that 401 no longer clears it. Net effect on exactly the accounts defect 1 is for is worse than before.

Blockers on the restore path and on `\S+`. Then about 60 lines of cuts, mostly prose and one helper. Commit subject isn't `type(scope):` — CONTRIBUTING asks for it and the three commits before this one comply.

src/transport/http/mega-client.ts
```diff
@@ -1157,10 +1199,21 @@ export class MegaHttpClient {
     {
       const cap = Object.keys(res).filter((k) => /captcha|answer|picture|image|fa_/i.test(k));
       this.logger.debug("[mega] login resp keys:", Object.keys(res).join(","));
+      // Whether the reply's two ids actually differ is the fact that says how often the gtoken header was
+      // wrong before gtokenHeader() pinned it — and it cannot be read from the key list above. Logged as a
+      // comparison, never as values: either one identifies the account.
+      if (res.ap_cloud_user_id !== undefined && res.user_id !== undefined)
```
lenoxys: **Cut.** This is the evidence that was supposed to gate the change above it, shipping in the same commit as the change. And it can't settle the question: it only runs inside `attemptLogin`, so it's silent on the restored-session path — which is the one that fails.

If it stays, it's one line without the guard and without three lines of comment.

src/transport/http/__tests__/session-expiry.spec.ts
```diff
@@ -124,6 +138,31 @@ describe("mega authenticated session rejection", () => {
...
+   */
+  it("keeps the session on a gtoken mismatch, which a re-login cannot fix", async () => {
+    const { mega, login, clearSession } = client([GTOKEN_MISMATCH]);
+
+    await expect(call(mega)).rejects.not.toBeInstanceOf(SessionExpiredError);
+    expect(login).not.toHaveBeenCalled();
+    expect(clearSession).not.toHaveBeenCalled();
+  });
+
+  it("still reports the gtoken mismatch as a failure, rather than swallowing it", async () => {
```
lenoxys: **Cut.** Passes against the old classifier. Reverted only the `tokenRejected` line in a worktree and ran the file: 1 failed, 17 passed, and this is one of the 17. The message and the redaction are identical either way, and the redaction test below already covers the echo. The test above it is the real regression test.

src/transport/http/__tests__/session-expiry.spec.ts
```diff
@@ -31,6 +31,20 @@ const TOKEN_NOT_EXIST: Response = {
   status: 401,
   data: { code: 401, msg: "token not exist, token = 0123456789abcdef0123" },
 };
+/**
+ * A 401 that is NOT the credential failing: the gateway recomputes the `gtoken` header from the user its
```
lenoxys: **Nit.** Ten lines of doc over a four-line fixture, and the 2FA cost is stated again on the test below. The two-traps sentence is the part that isn't in the code.

src/core/store.ts
martijnpoppen: You're right, and this was the worst of it — the two halves cancelling on exactly the accounts defect 1 exists for, which I'd missed entirely.

Fixed as you asked: `isSessionValid` now treats `accountUserId` as part of the credential and refuses a record without it. To make absence unambiguous, `persist()` always writes the field (equal to `userId` when the reply carried only one id), so a missing one means a pre-field record rather than a single-id reply.

Two details worth naming. The stored device identity is still read — `openudid`, `phoneModel` and the media UA are resolved in the constructor from `store.load()` directly, not through the hydration path — so the one re-login this costs is not a new device and doesn't re-trigger a challenge. There's a test pinning that. And `isSessionValid` is the right home rather than `hydrateFromStore`, because recovery re-reads the store too: a shared store holding a pre-field record must be refused there as well, not just at construction.

Three tests in the new `gtoken-header.spec.ts`, each checked against the unguarded version.

---

src/transport/http/mega-client.ts
martijnpoppen: Confirmed both, and you've talked me out of the approach rather than the regex. I probed the three cases against the shipped version:

| message (after `withoutTokenEcho`) | want | shipped |
|---|---|---|
| `token = <redacted>,token not exist` | true | **false** |
| `token = <redacted> expired` | true | **false** |
| `token=<redacted>&error=expired` | true | **false** |

Your second point is the real one: `[^\s,]+` fixes the first case and leaves the other two, because the anchor genuinely lives in the echo for any wording that states its reason beside it. Editing the message before classifying it is the mistake — it can only ever remove signal.

So the strip is gone and the bound moved into the match: `.*` → `[^,]*` on each alternative, so a wildcard stays inside the clause its anchor sits in, with `\btoken\b` still keeping `gtoken` from posing as the credential. That's the two traps closed without touching the message, and all three of your cases plus the original mismatch now classify correctly. `token = <redacted> expired` has a regression test.

Side effect you'll like: `withoutTokenClause` disappears entirely, which also settles your next comment.

---

src/transport/http/mega-client.ts
martijnpoppen: Gone — the classifier no longer edits the message at all, so there's one definition of an echoed token again and nothing to drift.

---

src/transport/http/mega-client.ts
martijnpoppen: Fair, and AGENTS.md is explicit on both counts — history narration and the three-other-modules reasoning belong in the commit. The JSDoc is now two sentences on the `auth_` field (the declaration survived, the function didn't — see your next comment): which of the two ids each field holds, and that `gtoken` hashes from `accountUserId` because that's what the gateway recomputes and rejects a disagreement on. The blast-radius argument moved to the commit message.

---

src/transport/http/mega-client.ts
martijnpoppen: Inlined at both sites as `gtoken(this.auth_.accountUserId ?? this.auth_.userId)`, and the surviving doc is on the `auth_` field. The inconsistency with `authTokenHeaders` eight lines below was the tell — thanks.

---

src/transport/http/mega-client.ts
martijnpoppen: Agreed, and @max246 asked for the same thing — it was the gap that mattered most, since that claim is the entire safety argument.

New `gtoken-header.spec.ts` drives `attemptLogin` with a synthetic reply and asserts on the outgoing header:

- ids differ → `gtoken` is `md5(user_id)`, and explicitly *not* `md5(ap_cloud_user_id)`
- ids agree → the byte-identical claim
- reply carries only `user_id` → unchanged
- persistence round-trip → `accountUserId` saved, and a client reconstructed from that store sends the same header

Checked against the unfixed code rather than assumed: reverting the pin fails tests 1 and 4, reverting the `isSessionValid` guard fails the two restore tests.

---

src/transport/http/mega-client.ts
martijnpoppen: Confirmed — `solix-client.ts:280` is the same `ap_cloud_user_id ?? user_id` chain feeding `gtoken(userId)` at 293.

Taking your second option, because @max246 makes the case against the first: there's no Solix response observed refusing the header, so a switch there would rest on nothing but the shape looking alike — the guess this PR is specifically trying not to be. Recorded on `classifyLogin` as a wire fact rather than a note about this change: which id it hashes from, that whether this gateway checks against `user_id` is unverified here, and that no Solix response has been seen refusing it, which is consistent with those accounts' ids agreeing.

@max246 — yes please to a separate PR for the Solix half. The `ids agree:` debug line is the cheap part and worth porting first; it's what turns this from a shape match into evidence.

---

src/transport/http/mega-client.ts
martijnpoppen: Kept, in the form you allowed for: one line, no guard, no comment.

```ts
this.logger.debug("[mega] ids agree:", res.ap_cloud_user_id === res.user_id);
```

On it not being able to settle the question — that was true when you wrote it and the blocker fix removes it. A pre-field record is no longer restorable, so every existing install passes through `attemptLogin` exactly once on upgrade, which is where this line sits. The restored-session path can't reach it because there is no longer a restored session that skipped a login.

On shipping the evidence alongside the change it was meant to gate: agreed in principle, but the gate came off for a different reason — the fix is byte-identical where the ids agree, so it doesn't need the answer to be safe. The line is now telling us how *often* the header was wrong, not whether to make the change. Happy to drop it if you'd still rather it went.

---

src/transport/http/__tests__/session-expiry.spec.ts
martijnpoppen: Cut. You're right that it didn't discriminate — I'd seen the same 1-failed/17-passed split when I checked the pair against the old classifier and drew the wrong conclusion from it, which is exactly what that split was telling me.

Its slot now holds a test that does discriminate: `token = <echo> expired`, the false negative the clause bound protects. Fails against the strip approach.

---

src/transport/http/__tests__/session-expiry.spec.ts
martijnpoppen: Trimmed to three lines, keeping the two-traps sentence. The 2FA cost now appears once, on the test rather than the fixture.

---

martijnpoppen:
Pushed `a596d24`. Thanks both — the review caught two things that would have shipped a net regression.

**@lenoxys' upgrade-path blocker was the serious one.** The two fixes cancelled each other on exactly the accounts defect 1 exists for: a stored session hydrated with the Passport id, kept sending the rejected header, and the classifier change meant that 401 no longer churned into the login that would have repaired it. `isSessionValid` now treats `accountUserId` as part of the credential and refuses a record written without it; `persist()` always writes the field, so absence means a pre-field record rather than a single-id reply. Costs one login per existing install on upgrade — and the stored device identity is still read in the constructor, so it isn't a new device and shouldn't prompt.

**The `\S+` blocker changed the approach, not just the character class.** All three probe cases confirmed false-negative against what I'd pushed:

| message (after `withoutTokenEcho`) | want | was |
|---|---|---|
| `token = <redacted>,token not exist` | true | false |
| `token = <redacted> expired` | true | false |
| `token=<redacted>&error=expired` | true | false |

`[^\s,]+` fixes the first and leaves the rest, because for any wording that states its reason beside the echo the anchor *is* the echo. Editing the message before classifying it can only remove signal. So the strip is gone and the bound moved into the match — `.*` → `[^,]*` per alternative, `\btoken\b` unchanged — which closes both original traps without touching the message, and drops `withoutTokenClause` and its drift with `withoutTokenEcho` along the way.

**The header half now has tests**, which both of you asked for and was the real gap: new `gtoken-header.spec.ts` drives `attemptLogin` with a synthetic reply and asserts the outgoing header for ids-differ, ids-agree (the byte-identical claim that justifies shipping ahead of a capture), single-id, and a persistence round-trip. Every new test was checked against the unfixed code rather than assumed — reverting the pin fails two, reverting the `isSessionValid` guard fails two, restoring the strip fails the `expired`-beside-echo case.

Cuts done: delegate inlined, JSDoc narration replaced by two sentences on the `auth_` field, body comments removed, id log down to one line, the non-discriminating spec dropped (it passed against the old classifier — I'd seen that split myself and misread it), fixture doc trimmed. Commit subjects are now `fix(session):` on both, and the PR title with them.

`npm run verify` green end to end: 3602 tests / 199 files, format, typecheck, all five guards, build, ESM, doc snippets, examples.

Two threads left open on purpose:

- **Solix** — recorded on `classifyLogin` as a wire fact, not switched. @max246's argument against a blind switch is the right one and matches this PR's own standard, so it's yours to take in a separate PR; the `ids agree:` debug line is worth porting first since it's what turns a shape match into evidence.
- **The id debug line** — kept in the one-line form @lenoxys allowed for, and his "can't settle it from the restored path" objection is answered by the blocker fix (every install now passes through `attemptLogin` once). Still happy to drop it if you'd rather.

Unchanged and still worth a reviewer's eye: neither fix has been driven against a live account. The classifier is fully exercised offline; the header change is verified as byte-identical where the ids agree and correct by the gateway's own error string where they don't, but nobody has watched it repair a live 401.

---

max246 (commented):
**Reviewed at `a596d24`.** Verified locally: `npm run verify` green — `tsc --noEmit`, `prettier --check`, all five guards, suite 199 files / 3602 tests. I also ran the new `tokenRejected` regex against a table of real and adversarial messages offline (the gtoken echo redacted *and* unredacted, expiry beside vs across a comma, `gtoken invalid` alone): it matches exactly what it should and drops exactly what it should.

The diagnosis is right and the split is the important part. A `gtoken` mismatch is a **header** fault on a token that is otherwise fine, so the id fix is what makes the call succeed and the classifier fix is what stops one refused call from clearing a healthy session behind it — either alone leaves a real failure, and both are cheap. Pinning only the one header the gateway checks (`accountUserId`) while leaving `auth_.userId` on `ap_cloud_user_id` for the Tuya/P2P/image-partition consumers that never asked for a different id is the correct, minimal cut, and requiring `accountUserId` in `isSessionValid` (one re-login for a pre-upgrade session, since the id only arrives with a login reply) is the honest call rather than restoring a session whose every authed call fails identically. The test coverage is genuinely complete — both fixes, persistence round-trip, the legacy-session refusal, and that the refused session still contributes its device identity so the re-login isn't a new device.

Three notes, none blocking:

**1. The PR description describes an implementation that isn't in the diff.** The first bullet says classification "drops the clause outright (`withoutTokenClause`); the echo is the credential and never the explanation." There is no `withoutTokenClause` — `withoutTokenEcho` is unchanged (still redaction-only, at the two log sites), and the actual classifier fix is the regex bounds on `tokenRejected` (mega-client.ts:260): a wildcard that can't cross a comma (`[^,]*`) plus whole-word `\btoken\b` so `gtoken` isn't read as the credential. That's a cleaner fix than dropping a clause, but the write-up reads as a superseded approach — worth reconciling so the record matches what shipped (the "address review" commit presumably changed the approach and the body didn't follow).

**2. The comma bound trades over-clearing for a narrow under-clearing risk — worth stating as deliberate.** `\btoken\b[^,]*(expired|error|not exist)` will not match an expiry whose reason is comma-separated from the anchor (`"token = X, expired"`) — I confirmed this empirically. Your evidence shows reasons sit beside the anchor (`"token = … expired"`, no comma), so this is fine today, but it rests on the one captured message; if a future diagnostics report ever shows a comma before a genuine expiry reason, that session would stop being cleared. A one-line comment that the clause bound is the intentional trade (and the failure mode it accepts) would flag it for whoever reads the next 401.

**3. Solix is left asymmetric — confirm it's a conscious "no evidence yet" call.** `solix-client.ts` gets only the JSDoc note; its `gtoken` still hashes `ap_cloud_user_id ?? user_id`, i.e. the exact shape the mega client just fixed. The note is honest that it's unverified and that the ids have agreed on accounts seen, and holding to evidence-first is the house rule — so I'd keep it as-is, just calling out that the identical backend pattern makes this the most likely *next* place to see the same 401, and the same `accountUserId` treatment would be cheap insurance if one turns up.

No objection — the two code fixes and the tests are solid; item 1 is the only one I'd want tidied before merge, and it's the description rather than the code.

---
_Reviewed with [Claude Code](https://claude.ai/code)_

max246 (commented):
### Against `AGENTS.md` specifically

Walked the PR through the repo's own rules. The code is clean on the hard ones; two items to fix, and the first is a rule breach rather than a preference.

**❌ Consumer-agnostic — the PR body names a specific host (rule violation).** `AGENTS.md`: *"Source, JSDoc, guides, examples, commit messages **and PR bodies** never name a specific consumer or its stack. Say 'caller' or 'host'."* The body names it three times — "a Homey Pro (Early 2023), Homey v13.5.1-rc.4", "Diagnosed in `homey-eufy-app`'s …", and the checklist's "the investigation in `homey-eufy-app`". The **evidence is exactly as strong** stated generically: "a host integration's v0.8.0 diagnostics report on an EU account, routed to `eu-pr`, `fa_info.step: 0`, then a 401 with `gtoken not equal userid error` 1.0 s after sign-in." Credit to the failure and the wire, not the host. (The source, tests and commit message are all clean — I grepped the diff; it's only the PR description.)

**⚠️ Reduce after green — the gtoken-id selection is stated twice (Reuse before abstraction).** `this.auth_.accountUserId ?? this.auth_.userId` is the "which id does `gtoken` hash from" decision, and it appears at both header sites (mega-client.ts:463 in `authTokenHeaders`, :849 in the media download). `AGENTS.md` §Reuse before abstraction: *"After focused tests pass, inspect the diff for same-contract helpers … Resolve every match."* A one-line private accessor — `private gtokenUserId() { return this.auth_.accountUserId ?? this.auth_.userId; }` — makes the two header paths read the id from one place, so they can't drift on which id the gateway checks (the exact bug this PR fixes). The persist default at :1225 is a write-time fallback, so it can stay as-is or reuse the same accessor.

**✅ Clean on the rest:**
- **Comments** — no new `//` body comments; the rationale is in JSDoc above `auth_`, `isSessionValid`, `PersistedSession.accountUserId` and `tokenRejected`, and reads as ground-truth/protocol fact, not iteration history.
- **No real device data** — tests use synthetic ids (`AP_CLOUD_ID`/`ACCOUNT_ID`/`STORED-UDID`), and the captured token is redacted. (The host name above is a consumer-agnostic issue, not a PII one.)
- **Evidence** — the `gtoken`-from-`user_id` change and the regex bounds are grounded in the captured `401 … gtoken not equal userid error`, not a borrowed third-party client.
- **Wire hard-rules** — N/A here (a header/classifier fix; no new parameter id, getter or write wire), so "unverified writes throw" / "no guessed getter" don't apply.
- **Architecture** — changes sit in `transport/http/` + `core/store.ts`; nothing crosses the `model/ ⊥ transport/` line.

So: genericize the body (the one that's actually a rule), optionally fold the id-selection into one accessor, and the `withoutTokenClause` description drift from my first pass. None touch the correctness of the two fixes.

---
_Reviewed with [Claude Code](https://claude.ai/code)_

max246:
The two code fixes look good to me — happy to approve once the **PR description** is genericized, which is the one `AGENTS.md` rule it currently trips (*"…and PR bodies never name a specific consumer or its stack. Say 'caller' or 'host'."*). The source, tests and commit message are all clean — it's only the description. Three spots, each fixable in place with the evidence fully intact:

**1. The "Evidence" opener**
> A v0.8.0 user diagnostics report from a **Homey Pro (Early 2023), Homey v13.5.1-rc.4** — an EU account routed correctly to `eu-pr`, `fa_info.step: 0` …

→ drop the host/hardware:
> A host integration's v0.8.0 user-diagnostics report — an EU account routed correctly to `eu-pr`, `fa_info.step: 0` …

**2. The "Diagnosed in …" line**
> Diagnosed in **`homey-eufy-app`**'s `docs/investigations/login-gtoken-401.md`, which is updated alongside …

→ name the artifact by role, not the repo:
> Diagnosed in the host integration's investigation notes, updated alongside …

**3. The checklist item**
> - [x] Docs updated — the investigation in **`homey-eufy-app`**, in its own PR.

→
> - [x] Docs updated — the investigation lives in the host integration's repo, in its own PR.

That's the only thing I'd want changed before a clean approve. Two optional-but-nice follow-ups (neither blocking): fold the `this.auth_.accountUserId ?? this.auth_.userId` selection into one private accessor so the two header sites can't drift, and reconcile the body's `withoutTokenClause` wording with what actually shipped (the `tokenRejected` regex bounds — there's no such function). Details in the two reviews above.

src/transport/http/mega-client.ts
```diff
@@ -428,15 +452,15 @@ export class MegaHttpClient {
...
+   * The account-credential headers every authed call carries — `x-auth-token` + `gtoken`.
    * One place so the signed path, the key-exchange and the bearer path can't drift on what "authed" means.
    */
   private authTokenHeaders(): Record<string, string> {
     if (!this.auth_) return {};
     return {
       "x-auth-token": this.auth_.authToken,
       authorization: this.auth_.authToken,
-      gtoken: gtoken(this.auth_.userId),
+      gtoken: gtoken(this.auth_.accountUserId ?? this.auth_.userId),
```
max246: Reduce-after-green (`AGENTS.md` §Reuse before abstraction): this `this.auth_.accountUserId ?? this.auth_.userId` — the "which id does `gtoken` hash from" decision — is repeated verbatim at the media-download header (`:849`). A one-line private accessor keeps both header paths reading the id from one place, so they can't drift on the exact id the gateway checks (the bug this PR fixes):

```ts
private gtokenUserId(): string {
  return this.auth_!.accountUserId ?? this.auth_!.userId;
}
```

Both sites then read `gtoken(this.gtokenUserId())`. `auth_` is already guarded non-null on both paths, so no new check is needed. Non-blocking — just so the id selection lives in one place next to where it's persisted.

martijnpoppen: Done in `c8b75d0`, as `gtokenUserId()` with your body verbatim.

Worth noting it doesn't conflict with @lenoxys' earlier cut of the delegate, which I want to be explicit about since the two reads look opposed. His objection was the *parameter* — a one-line method taking as an argument the object it hangs off, while `authTokenHeaders` eight lines below reads `this.auth_` directly. Parameterless, it reads the same way as its neighbour and removes the duplication, so both notes land.

`this.auth_!` matches the three existing non-null assertions in `src/`, and both call sites already guard.

---

martijnpoppen:
All three done, plus one you didn't catch. Head is `c8b75d0`.

**Description genericized.** All three spots taken as written — the report is now "a host integration's v0.8.0 user-diagnostics report", the investigation is named by role, and the checklist item likewise. The evidence survives intact: the wire facts that matter (`eu-pr`, `fa_info.step: 0`, the gateway's own error string) were never the host-specific part.

**Your commit-message assessment was too generous.** The first commit said *"Diagnosed in `homey-eufy-app`'s docs/investigations/…"*, which trips the same rule. Reworded to name the artifact by role. `guard:consumer-agnostic` scans `src/`, `docs/` and `examples/` only, so neither the body nor the commit was ever going to be caught by CI — which is worth knowing about the guard, not a complaint about it. The second commit was clean as you said.

**Stale `withoutTokenClause` wording fixed.** You're right that there's no such function — it existed in the first push, and @lenoxys' `\S+` blocker is what removed it. The description now describes the bounds that actually shipped (`.*` → `[^,]*` per alternative, plus `\btoken\b`) and says why editing the message before classifying it was the wrong idea rather than leaving a reader to reconcile two accounts.

**Accessor folded in** as `gtokenUserId()` — details on the thread, including why it isn't a re-litigation of @lenoxys' cut of the earlier delegate.

The description also gained a paragraph the first version was missing entirely: the refused-restore path for pre-`accountUserId` records. That was @lenoxys' blocker, it's the largest behaviour change here, and it deserves to be in the description rather than only in a resolved thread.

`npm run verify` green on the new head: 3602 tests / 199 files.

Still open and still yours: the Solix follow-up, and @lenoxys' call on whether the `ids agree:` debug line stays.

---

max246 (approved):
Re-reviewed at `c8b75d0` — all three of my earlier points are addressed, and cleanly:

- **Consumer-agnostic:** the host name is out of the body; the evidence reads generically now.
- **Reduce-after-green:** `gtokenUserId()` reads the id from one place and both header paths (`authTokenHeaders`, the media download) go through it — the two can no longer drift on which of the session's ids the gateway checks, which is the exact failure this PR fixes. Accessor JSDoc is ground-truth and notes the `auth_`-guard precondition.
- **Description drift:** the `withoutTokenClause` wording is gone, so the write-up now matches what shipped (the `tokenRejected` comma/whole-word bounds).

Verified locally on `c8b75d0`: `npm run verify` green (3602 tests, all guards, prettier), and CI is green. The two code fixes were correct to begin with — the id fix makes the call succeed, the classifier fix stops one refused call from clearing a healthy session — and the tests cover both directions plus the legacy-session refusal. Nothing else outstanding from me.

---
_Reviewed with [Claude Code](https://claude.ai/code)_

PR #192, opened 2026-09-17
feat(model): classify the T9000 station as the station it is

lenoxys:
Device type 27 sat between `CAMERA3_PRO` (26) and `HOMEBASE_MINI` (28) with nothing declaring it, and `T9000` is the one station model family outside the T8 band, so every station prefix rule missed it and the residual rule took it for a camera.

A base taken for a camera carries no arming surface, and the cameras that name it their parent resolve against a station nothing classifies as one.

Evidence: a field record and a diagnostics archive give type 27, model T9000, hardware P2, firmware 4.1.5.8, with p2p, push and ndt identifiers present, and attached cameras naming it in `parent_sn` while resolving as `topology=attached` with `stationModel=T9000`. The app's own model registry files T9000 as family `STATION_9000` — that is where the device-type constant's name comes from, and it is the only name any source here supplies, so there is no curated registry row and the T-code stays the inferred name.

Both routes are covered, the numeric type and the model code, because a record that states one does not always state the other.

It is a hub, not an NVR. `HOMEBASE_TYPES` is the narrower set that decides which stations are hubs — it gates `reboot()`, the hub audio and the hub siren — so the T9000 joins it too, on the same shared-hub-hardware basis as the chime and the mini. A station with attached cameras left out of that set loses restart and advertises no siren, by omission rather than by decision.

## What this does not claim

**No fix for the connect failure.** The archive shows the session to that base never connecting with both lookup channels available, and classification is not shown to fix that.

It does, however, touch the connect path: `stationPower()` short-circuits on `deviceClass === "homebase"`, which this classification now produces, so the base joins the wired prewarm set by that branch rather than by its capabilities. The tier is unchanged only because the camera path also answered `wired` for a base with no `battery` capability.

## Verified

`npm run verify`: prettier, tsc, the five guards, build, ESM check, snippets, examples, 3594 tests over 198 files. Both new contracts fail against the parent commit's sources, with `expected 'camera' to be 'station'`.


src/model/classify.ts
```diff
@@ -51,6 +51,7 @@ import { PRINTER_CATEGORY_RE } from "../core/types.js";
 const STATION_TYPES: ReadonlySet<number> = new Set([
   DeviceType.STATION,
   DeviceType.HB3,
+  DeviceType.HOMEBASE_S1_PRO,
```
martijnpoppen, line 54: **`HOMEBASE_TYPES` in `device-family.ts` needs this too, and missing it leaves the S1 Pro half-classified.**

`STATION_TYPES` here drives the codec. `HOMEBASE_TYPES` (`device-family.ts:101`) is the deliberate subset that drives hub-only *behaviour*, and this PR doesn't touch it:

```ts
export const HOMEBASE_TYPES: ReadonlySet<number> = new Set<number>([
  DeviceType.STATION, DeviceType.HB3, DeviceType.MINIBASE_CHIME, DeviceType.HOMEBASE_MINI,
]);
```

So after this PR the two disagree — on the branch as it stands:

```
HomeBase 3 (HB3)   type=18  classify=station  isHomeBase=true
HomeBase S1 Pro    type=27  classify=station  isHomeBase=false
```

That set's own JSDoc says the exclusion exists for the NVRs — *"those resolve to `station` for arming/storage but have no speaker, so they must NOT expose the hub-audio controls (they'd fire at nothing)"*. An S1 Pro is a HomeBase with a speaker, so it lands in the NVR bucket by omission rather than by decision.

Three consequences today, all from `isHomeBase`:

- `capabilities/siren.ts:281` and `:296` — two members gated `available: isHomeBase`, so the S1 Pro advertises no siren surface.
- `capabilities/audio.ts:161` — same for hub audio (alarm / voice-prompt volume).
- `client/eufy-mega.ts:2006` — `reboot()` throws outright: `reboot: <sn> is not a HomeBase (deviceType 27, …) — restart is a hub-only operation`.

Note `isHomeBase` reads `ctx.deviceType` only, so the `T9000` model rule you add below doesn't cover for it either — a device known by model alone is still excluded.

Assuming the S1 Pro does have the hub speaker (its name says so, but you have the hardware), adding `DeviceType.HOMEBASE_S1_PRO` to `HOMEBASE_TYPES` is the fix. Worth a spec line beside the new ones asserting `isHomeBase({ deviceType: DeviceType.HOMEBASE_S1_PRO })` — that's the assertion that keeps the two sets from drifting again.

---

src/model/device-types.ts
```diff
@@ -42,6 +42,7 @@ export const DeviceType = {
   PROFESSIONAL_247: 24,
   MINIBASE_CHIME: 25,
   CAMERA3_PRO: 26,
+  HOMEBASE_S1_PRO: 27,
```
martijnpoppen, line 45: Checked that `27` is free — it's not referenced anywhere else in `src/`, and it fills the gap between `CAMERA3_PRO: 26` and `HOMEBASE_MINI: 28`, so nothing is displaced. ✅

What the PR body doesn't say is where `27` and `T9000` come from. Both are the kind of claim `AGENTS.md` wants anchored ("Ground every wire claim in evidence you can point at", and specifically that older third-party eufy clients are wrong for this protocol and a model will confidently reproduce them). A device-type number is lower-risk than a param id — a wrong one misclassifies rather than firing a bad write — but it's also the sort of value that's easy to pick up from a stale community table.

If this came off a real S1 Pro's device record (the `device_type` the cloud reports for it), one line in the PR body saying so is enough. Neighbouring entries in this file don't carry per-entry provenance, so I'm not suggesting a JSDoc here — just that the PR states it.

---

martijnpoppen (commented):
Picked this up as a new PR while watching the others. Small and well-shaped — both classification paths covered (device type *and* model prefix), the registry entry added so the name resolves, and the spec asserting all three. The "one station outside the T8 band" observation is a good one to have written down; that's exactly the sort of thing that silently rots otherwise.

**Verified locally** on `6c104c3`: `tsc --noEmit` clean, `prettier --check` clean, all five guards pass, suite green at 198 files / 3594 tests.

**One thing I'd fix before merge**, inline: `STATION_TYPES` gets the new type but `HOMEBASE_TYPES` in `device-family.ts` doesn't, so the S1 Pro classifies as a `station` while `isHomeBase()` still answers `false`:

```
HomeBase 3 (HB3)   type=18  classify=station  isHomeBase=true
HomeBase S1 Pro    type=27  classify=station  isHomeBase=false
```

That set is the one gating siren, hub audio and `reboot()` — and its JSDoc says the exclusion is there for the speakerless NVRs, which an S1 Pro isn't. As it stands `reboot()` on an S1 Pro throws "is not a HomeBase". Details and the three call sites are in the inline comment.

Second, minor: the PR doesn't say where `device_type 27` and the `T9000` prefix come from. If they're off a real S1 Pro's device record, worth one line in the body saying so.

Nothing else — the inline `//` above the `T9000` rule matches the existing comment style in `codecFromModel`, so I'd leave it alone rather than make this file inconsistent with itself.

---

src/model/classify.ts
lenoxys: Blocker, or at least an answer I need in the diff: `HOMEBASE_TYPES` in `device-family.ts` is the other station set in this layer, and it is untouched. It is what gates the hub-audio surface (alarm volume, voice prompt) — on device types, not on the codec — so after this the S1 Pro is a station that arms and stores but has no speaker controls, exactly the way the NVRs are deliberately shaped.

If that is deliberate because nothing in the field record proves a speaker, that is the correct call under the evidence rule — but say it, here or beside that set. As written a reader cannot tell a decision from a miss, and the set two files over answers the sibling question the other way for every other hub.

src/model/classify.ts
```diff
@@ -207,6 +209,8 @@ export function codecFromModel(model: string | undefined): Codec | undefined {
   //  HomeBase family: T8001/T8002/T8010/T8023/T8025/T8030 ; NVR: T8N00 ; PoE NVR: T8E00.
   if (/^T8(00[0-9]|010|023|025|030)/.test(m)) return "station";
   if (/^T8N0/.test(m) || /^T8E0/.test(m)) return "station";
+  // HomeBase S1 Pro: the one station model outside the T8 band, so no prefix rule above reaches it.
```
lenoxys: Cut. The JSDoc list thirty lines up already states T9000 is the one station outside the T8 band; this says it again, the new spec's JSDoc a third time, the PR body a fourth. Four lines of code do not need four statements of the same fact — and the JSDoc is the one that ships.

src/model/__tests__/model.spec.ts
```diff
@@ -28,6 +28,18 @@ describe("classify (device_type → codec)", () => {
     expect(codecForType(10)).toBe("sensor"); // motion sensor
     expect(codecForType(11)).toBe("keypad");
     expect(codecForType(9)).toBe("camera"); // CAMERA2 → residual camera bucket
+    expect(codecForType(27)).toBe("station"); // HomeBase S1 Pro
```
lenoxys: This and `classify({ deviceType: 27 })` below assert the same fact through the same set membership. Keep one — the table line fits the existing style, so the new test could carry only the two model-code assertions, which are the route nothing else covers.

src/model/__tests__/model.spec.ts
```diff
@@ -28,6 +28,18 @@ describe("classify (device_type → codec)", () => {
...
+  });
+
+  /**
+   * The S1 Pro is the only station whose model code sits outside the T8 band, so every prefix rule for a
+   * station misses it and the residual camera rule cannot catch it either. A base taken for anything but a
+   * station carries no arming surface and its attached cameras resolve against a parent nothing calls one.
+   */
```
lenoxys: Five lines of rationale over three assertions, and it repeats the JSDoc in `classify.ts`. One line stating why the model route exists at all is enough; the protocol reasoning belongs where it ships, not in a spec that never does.

src/model/registry.ts
```diff
@@ -45,6 +45,7 @@ import { inferName } from "./infer.js";
  */
 export const MODEL_REGISTRY: Readonly<Record<string, RegistryEntry>> = {
   // Stations / hubs.
+  T9000: { codec: "station", name: "HomeBase S1 Pro" },
```
lenoxys: Where does the display name come from? The evidence in the body covers type, model, hardware and firmware — not the retail name, and "S1 Pro" is a name the vendor also uses on the clean line. A row's only unique contribution over `inferName` is that string, and nothing asserts it, so a wrong name here is both unsourced and unguarded. One clause of provenance would settle it.

lenoxys (commented):
Classification itself is right and the evidence carries it. Two things: the hub-audio set in the same layer was left behind, and the same sentence is now written four times for four lines of code.

One body correction: this does touch the connect path. `stationPower()` short-circuits on `deviceClass === "homebase"`, which this classification now produces, so the base joins the wired prewarm set and gets the persistent idle policy. The outcome is unchanged only because the camera fallback also answered `wired` for a base with no `battery` cap — that is a reason, not an absence of effect.

Gate re-run here: green, 3594 tests, and the new contract does fail without the source change.

src/model/classify.ts
lenoxys: Right, and `reboot()` is the one that mattered most — I first answered this by documenting the exclusion, which was wrong: the set's own policy already admits the chime and the mini on shared hub hardware rather than their own capture, so a station with attached cameras belongs there the same way. In as `STATION_9000` (edb071b), with the spec asserting both sets for type 27 since nothing covered the hub gate at all.

src/model/device-types.ts
lenoxys: Provenance is in the body now: the field record and diagnostics archive for the type, and the app's own model registry filing T9000 as family `STATION_9000` — which is also where the constant's name comes from now, replacing the retail name nothing here sourced.

src/model/classify.ts
lenoxys: Answered by adding it rather than documenting the exclusion — see martijnpoppen's thread above, `reboot()` was throwing on a hub.

src/model/classify.ts
lenoxys: Cut, the JSDoc list carries it.

src/model/__tests__/model.spec.ts
lenoxys: Dropped the `classify({ deviceType: 27 })` assertion; the table line keeps the type route and the new test keeps the two model-code ones.

src/model/__tests__/model.spec.ts
lenoxys: Down to three lines, and they now state the one thing the code cannot: that the two sets are asserted together because their drift was the bug.

src/model/registry.ts
lenoxys: No provenance existed, so the row is gone — the app registry gives T9000 no retail name, only family `STATION_9000`, and the row carried nothing else. `inferName` answers `T9000` until a capture curates a real one.

lenoxys:
**Second pass, please** — two commits on top of the one you reviewed, and one of them reverses my own earlier answer.

- `ebb287a` — the device type is `STATION_9000` now, not `HOMEBASE_S1_PRO`, and the curated registry row is gone. Your provenance point stands: nothing here sources the retail name. The app's own model registry files T9000 as family `STATION_9000` and gives no product name, so that is the name the constant carries, and the row held nothing but the unsourced string.
- `edb071b` — `HOMEBASE_TYPES` gains it, as you asked. I first tried documenting the exclusion instead; that was wrong. `reboot()` was throwing on a hub, and the set already admits the chime and the mini on shared hub hardware rather than their own capture. The spec now asserts both sets for type 27, since their drift was the bug and nothing covered the hub gate.

Body rewritten: the provenance, the hub decision, and a correction — this does touch the connect path, because `stationPower()` short-circuits on `deviceClass === "homebase"`; the tier is unchanged only because the camera path also answered `wired`.

`npm run verify` green. Three assertions fail against the parent commit's sources: the two model routes and the hub gate.

danymexi:
I have a HomeBase S1 Pro (T9000) in hand and can supply field data for the connect failure this PR explicitly leaves open.

One constraint worth flagging up front: the Eufy account allows a single active session, so capturing anything means taking my live setup down for the duration. If you tell me exactly what you need — a diagnostics archive from the config entry, a debug log scoped to particular modules, or a specific sequence of actions to perform while logging — I can do it in one planned session rather than several partial ones.

Models I can test against besides the T9000: T8010, T8030, T8113, T8113-Z, T8122, T8134, T8161, T8170, T8210, T8214, T8400, T8410, T8419, T8423, T8425, T814X, T84A1, T8790, T8W11C.

max246 (approved):
LGTM — approving. Nice iteration; every thread here landed as an actual code change rather than a defense, and the evidence-first calls are the right ones.

**Verified locally at `edb071b`:**
- `npm run verify` fully green — 3594 tests / 198 files, all five guards, build, ESM, snippets, examples.
- Device type `27` is referenced nowhere else in `src/` (the lone grep hit is `"27%"` in a battery comment), so it fills the `26→28` gap cleanly.
- **The two station sets agree**, which was the core issue martijn raised: `STATION_9000` is in both `STATION_TYPES` (codec → arming/storage) and `HOMEBASE_TYPES` (hub gate → `reboot()` + siren + hub audio), and the new spec asserts `isHomeBase({ deviceType: STATION_9000 }) === true`, so they can't silently drift again.

**Good calls on the evidence discipline:**
- Naming the constant `STATION_9000` after the app's own model-registry family instead of the retail "S1 Pro" — the retail name was the one thing no source here actually supplied.
- Dropping the `registry.ts` row entirely rather than shipping an unsourced display name; `inferName` answering `T9000` until a real capture curates one is the honest state.
- Covering both routes (numeric type 27 *and* model `T9000`), since a record that states one doesn't always state the other.

The `isHomeBase` reads `deviceType`-only limitation martijn noted (a device known by model alone still misses the hub gate) is a pre-existing property of the two-layer design that applies to every hub, not something this PR introduces — no need to chase it here.

Ships an honest classification and correctly scopes out the connect-failure fix. 👍

PR #195, opened 2026-09-17
Stop the v2 thumbnail reconstruction costing 45 MB a picture, and say why one failed

martijnpoppen:
## What changed

Two commits, both in the push-thumbnail path.

**1. The v2 reconstruction reads the scan instead of decoding candidate frames, and re-encodes nothing.**

New `transport/http/jpeg-scan.ts` Huffman-walks the plaintext entropy-coded scan and throws every coefficient away — no IDCT, no component planes, no output image. It keeps the luma DC per MCU, which is enough for both unknowns the encrypted head hides:

- **subsampling** — it decides which Huffman table each block position in an MCU is coded with, so a wrong hypothesis diverges within a few blocks; the one that walks the scan to its last byte on a whole-MCU boundary is the answer.
- **width/height** — an MCU count is exactly `w × h`, so the candidates are its divisor pairs, picked by row continuity over the DC values. Divisor pairs, not a ladder of guessed sizes, so an off-ladder geometry stopped being a special case.

`jpeg-js` now decodes exactly one frame, as a probe: it proves the spliced JPEG decodes and measures how far the substitute quant tables fall short of the camera's. That shortfall is a single scale factor, so it is corrected by rewriting the DQT rather than the pixels — which is what removes the re-encode. `autoContrast()` goes with it (removed from the public surface).

**2. A failed candidate says why, in a vocabulary that is safe to log.**

`[stored-snapshot-cache] candidate failed { class: 'download-failed' }` was every outcome between observing a URL and holding bytes, and `downloadMediaResource` collapses them into three deliberately uninformative messages. Media failures now carry a `mediaFailure` tag from a closed vocabulary in `core/contracts.ts` — `url-not-allowed`, `address-not-public`, `redirect-not-allowed`, `http-status` (with `status`), `too-large`, `timeout`, `network`, `decode-failed` — and the cache reads the tag off the error, checks it against that vocabulary, and logs the term. A 404 now reads `cause: 'http-status', status: 404`.

Also in that commit: the per-device dedupe set is bounded to the last 64 URLs (it kept every URL a device ever sent, for the life of the process), and the guide's claim that v2 wrappers do not pass validation is corrected — they have been reconstructed on this path for some time, and now nothing is re-encoded.

## Why

A caller under a hard memory cap was killed by its watchdog shortly after a detection push. Measured here, on the committed fixtures, back to back:

| | before | after |
| --- | --- | --- |
| RSS per thumbnail | **+45.5 MB, then +39.5 MB** — permanent (glibc keeps the arenas) | **+0.4 MB** (+13 MB once per process, JIT) |
| blocked event loop | 176 ms / 207 ms | 6 ms |
| `jpeg-js` calls | ~30 decodes + 1 encode | 1 decode, 0 encodes |

Ordering the ladder by MCU count had already cut this from ~150 MB and it was still fatal: the per-decode cost, not the candidate count, was the bill — and the encoder turned out to be the larger half at ~40 MB of RSS on first use.

The second commit is the follow-up question that could not be answered from a log: one of those failures was observed in the field and there was no way to tell a 404 from a refused host from bytes that would not decrypt.

## Evidence

Measured, not reasoned about:

- `jpeg-js` cost, isolated: `encode` is +40.4 MB on first call and ~2.2 MB after; `decode` is +2.6 MB then ~0.4 MB. 30 decodes of a 320×180 frame move RSS +51 MB and it does not return after `global.gc()`.
- Geometry recovery verified against 10 geometries × 4:4:4 / 4:2:2 / 4:2:0 (30 synthetic blobs generated with the repo's own fixture generator, plus the 2 committed fixtures): exact in every case, up to the MCU alignment the old search was also limited to (the display size lives in the encrypted SOF).
- Contrast recovery verified by encoding the same synthetic image at several qualities: spliced at the reference table a quality-30 source spans 103..152 of 0..255; the correction names quality 31 for it, 52 for a quality-50 source, and 85 (i.e. unchanged) for a quality-95 one.
- Reconstruction output now matches the fixture's own original picture within a point of the mean on every channel — the fixtures' plaintext heads make that checkable, and the specs check it.
- The failure tag was wired end to end through the real cache and the real downloader against a stubbed 404: `{ class: 'download-failed', cause: 'http-status', status: 404, observedAt: …, retained: false }`.

n/a on a real device for the decoder itself: every fixture is synthetic, per repo policy. The crash it fixes was observed on real hardware.

## Verified how

- [x] `npm run verify` is green locally (3618 tests)
- [ ] Ran against a real device — the failing behaviour came from one; the fix is exercised by fixtures only
- [x] Added or updated specs (offline, synthetic fixtures) — two new fixtures (4:2:0, and a low-quality source for the fog), a scanner spec, a cost spec that pins "one decode, zero encodes", and tags per failure path
- [x] Docs updated (`docs/getting-started.md`, `docs/live-media.md`)

## Unverified or deferred

- **No retry, still.** `it("classifies downloader rejection without retrying")` pins that as intended, so a candidate that fails is lost until the next event. If the field failure comes back as `http-status 404` — the thumbnail not yet uploaded when the push landed — a single short retry is the fix, but that is a contract change and it is not in here.
- **The contrast correction cannot compress.** A thumbnail whose camera encoded *above* the reference quality reconstructs slightly over-contrasted and clipped; the stretch is clamped at 1. The old per-channel autocontrast could not undo clipping either. Real thumbnails fog, i.e. they err the other way.
- **A v1 wrapper whose payload is not a whole number of cipher blocks still throws.** It is now reported as `decode-failed` rather than looking like a network failure, but the decoder was left as it is: a sub-256-byte payload is not a real image.

https://claude.ai/code/session_01NffUQVpLqYxn7ZSi7JnWaC

---
_Generated by [Claude Code](https://claude.ai/code/session_01NffUQVpLqYxn7ZSi7JnWaC)_


src/transport/http/decodeImageV2.ts
```diff
@@ -144,127 +147,197 @@ interface Decoded {
...
-      }
-    }
-
-    let lo = 0;
-    while (lo < 256 && hist[lo] === 0) lo++;
-    let hi = 255;
-    while (hi >= 0 && hist[hi] === 0) hi--;
+    const { lo, hi } = trimmedRange(img.data, channel, total, cutoff);
+    // A channel with no range at all says nothing about scale, and amplifying it would only turn a
+    // flat picture into a black-and-white one. PIL's autocontrast skips it for the same reason.
```
max246: Inline body comment (`AGENTS.md` hard rule: *no inline body comments*). The channel-skip rationale belongs in the JSDoc of the function it sits in, not narrated in the body.

src/transport/http/jpeg-scan.ts
```diff
@@ -0,0 +1,321 @@
...
+    const dc = tables.get(assignment.dc);
+    const ac = tables.get(0x10 | assignment.ac);
+    if (!dc || !ac) return null;
+    components.push({ h: index === 0 ? lumaH : 1, v: index === 0 ? lumaV : 1, dc, ac, pred: 0 });
+  }
+
+  const reader = new BitReader(tail, scanStart);
+  // Grown as the scan turns out to be long rather than sized from the scan's byte length: a wrong
+  // hypothesis dies after a handful of MCUs, and provisioning three whole-scan arrays to discover that
+  // is exactly the kind of allocation this module exists to avoid.
```
max246: Inline body comment — `AGENTS.md` wants this in the enclosing function's JSDoc rather than in the body.

src/transport/http/jpeg-scan.ts
```diff
@@ -0,0 +1,321 @@
...
+      consumed = reader.consumed;
+      if (consumed >= tail.length) {
+        complete = true;
+        break;
+      }
+    }
+  } catch (error) {
+    if (!(error instanceof ScanEnd)) throw error;
+    // Clean when the divergence happened at the START of an MCU with nothing but a marker left: the
+    // scan handed over exactly whole MCUs. Anything else is a hypothesis reading the wrong tables.
```
max246: Inline body comment — and `cleanMcuBoundary` already carries a JSDoc at :65 this can fold straight into.

src/transport/stored-image-cache.ts
```diff
@@ -53,6 +69,12 @@ export class StoredImageCache {
     }
     if (state.seenUrls.has(url)) return;
     state.seenUrls.add(url);
+    // A Set iterates in insertion order, so the first entry is the oldest attempt.
```
max246: Inline body comment (hard rule). Move to the method's JSDoc or drop it — the `Set` insertion-order fact reads fine as a one-liner above the declaration.

max246 (commented):
Reviewed at `a056dbf60e80560c9077edc71c92ed349f28f6d6`. Verified locally — `npm run verify` green (199 files / 3618 tests), and I ran the v2 decoder against the fixtures directly.

**The decode is correct and well-tested — no issue there.** Same keyless head-splice method (marker `FF C4 00 1F 01`, plaintext tail), now recovering geometry + subsampling from the scan instead of assuming a fixed header. The round-trip block proves it pixel-level (within a point/channel) across 176×144, off-ladder **264×200**, and **320×240 4:2:0**; the Huffman walker handles byte-stuffing, `RSTn`, `DRI` and bounds; and it fails safe to `null` on a truncated scan rather than emitting a scrambled frame. Correcting the DQT scale instead of re-encoding is the right call and where the ~45 MB → tiny win comes from.

Two things against the repo's own rules, though:

**1. Inline body comments — hard rule.** `AGENTS.md`: *"no inline body comments … move it up, or delete it."* Four added (flagged inline). Each belongs in the enclosing declaration's JSDoc.

**2. `MediaFailure` placement.** It lives in `core/contracts.ts` but is consumed only within `transport/` — nothing in `model/`/`client/` reads it and it isn't on the public surface. `AGENTS.md` calls `contracts.ts` *"the one thing genuinely shared across layers."* Either it's caller-facing media-contract vocab beside `MediaProvider` (then it isn't surfaced yet — a gap) or it's transport-internal (then it belongs in `transport/`, since a transport module may import another). A one-line note on which would settle it.

Everything else is clean: consumer-agnostic throughout (no host names — nice), no real device data (the hex is standard JPEG tables), and the geometry heuristic fails safe. Solid work on the memory drop.

---
_Reviewed with [Claude Code](https://claude.ai/code)_

src/transport/http/decodeImageV2.ts
martijnpoppen: Moved into `contrastScale`'s JSDoc, folded into the paragraph that already explains why the tightest channel wins — a channel with no range says nothing about scale, and amplifying it would turn a flat picture into a black-and-white one. Fixed in 87b208e.

---

src/transport/http/jpeg-scan.ts
martijnpoppen: Moved into `scanEntropy`'s JSDoc — why the DC array grows rather than being sized from the scan's byte length (a wrong hypothesis dies after a handful of MCUs, so three whole-scan arrays would be exactly the allocation this module exists to avoid). Fixed in 87b208e.

---

src/transport/http/jpeg-scan.ts
martijnpoppen: Folded into the JSDoc at that spot — it's the doc on `EntropyScan.complete`, which is what the line was explaining, plus a sentence in `scanEntropy`'s own JSDoc saying where the walk stops. (There's no `cleanMcuBoundary` symbol; the flag is `complete`.) Fixed in 87b208e.

---

src/transport/stored-image-cache.ts
martijnpoppen: Moved into `observe`'s JSDoc, along with what the window is for: one event arrives as several pushes sharing a URL, and the `Set` iterates in insertion order, so the entry evicted when it is full is the oldest attempt. Fixed in 87b208e.

---

martijnpoppen:
Thanks — both addressed in 87b208e.

**1. Inline body comments.** All four are gone, each folded into the JSDoc of the declaration it belonged to (replied on the threads, resolved).

While in there I applied the same rule one line further down in `AGENTS.md` — *"a JSDoc does not narrate its own history"* — which the new module docs were breaking: what the search used to do, how much an earlier fix had already saved. What makes the design necessary is a present-tense fact about what a JPEG decoder costs per question, so it now says that, and the before-and-after stays in the commit that made it and in this PR's body.

**2. `MediaFailure` placement — it is transport-internal.** Settled that way rather than with a note: raised in `transport/http/media-download.ts` and `transport/http/mega-client.ts`, read in `transport/stored-image-cache.ts`, named by nothing in `model/` or `client/`. So it moves to `transport/media-failure.ts`, alongside `transport/ffmpeg.ts` and `transport/dp-preset.ts` as a transport module other transport modules import. It leaves the public surface with it — `MEDIA_FAILURE_REASONS` was published by the core barrel, which nothing asked for. 288 exports back to 287.

The reason it was NOT made caller-facing: the closed vocabulary exists so the cache can log a failure without quoting a signed URL or a response body, and that is a diagnostic, not a contract. The caller-facing gap you name is real but it is a different change — attaching the tag to `StoredSnapshotUnavailableError` so a caller reading `download-failed` can also read *why*. Happy to do it, in its own PR with its own specs, rather than widen this one.

`npm run verify` green (3618 tests), and the docs build too (TypeDoc clean, publication guard clean) — the earlier red check was a `{@link}` from a public comment to a module-private symbol, with `treatWarningsAsErrors` on.

---

max246 (approved):
Re-reviewed at `87b208e` — both of my findings are addressed, and cleanly:

- **`MediaFailure` placement:** moved out of `core/contracts.ts` into `transport/media-failure.ts`, consumed only by the cache and the downloader (both `transport/`) — so it's in the layer that owns it, not hoisted into the shared floor.
- **Inline body comments:** all four are gone; the rationale is in JSDoc now, and a scan of the diff shows no `//` body comments remaining in shipped `src/`.

The v2 decode was already correct and well-tested to begin with — same keyless head-splice at `FF C4 00 1F 01`, geometry + subsampling recovered from the scan, a pixel-level round-trip across three geometries (176×144, off-ladder 264×200, 320×240 4:2:0), and it fails safe to `null` on a truncated scan rather than emitting a scrambled frame. The DQT-scale de-fog instead of a re-encode is the right call and where the memory drop comes from.

Verified locally on `87b208e`: `npm run verify` green (3618 tests), and CI is green (including the docs build). Nothing else outstanding — approving.

---
_Reviewed with [Claude Code](https://claude.ai/code)_

PR #196, opened 2026-09-17
Solix Solarbank 4 (AE103): telemetry + ambient light / display timeout / SOC-limit controls

max246:
## Solix Solarbank 4 (AE103) support — telemetry + controls

Adds full support for the Anker Solix Solarbank 4 E5000 Pro (model AE103) on top of the existing
`SolixClient` account/MQTT stack. All wires are live-confirmed on real hardware; no guessed write paths.

### Telemetry (read)
- `param_info` ff09 decode for the Solarbank tag family (SOC, battery/charge/discharge power, PV strings,
  grid/home-load, AC-plug/socket, battery temperature from the `a4` BMS blob with a self-validating SOC check).
- `state_info` push decode with its own settings-tag table (operating/EMS mode, max home load).
- `getSiteScene()` + `solarbankSceneReadings()` as a low-rate backstop for battery temperature.
- App-side ambient-light and display-timeout changes read from the device `…/req` command channel.

### Controls (write, live-verified)
- **Ambient light** — `set_device_attrs` encrypted write.
- **Display timeout** — `…/req` cmd-17 (ff09 msgtype 0x68, tag `a5=[01,index]`) MQTT command, incl. Never (index 0).
- **SOC discharge/charge limits** — `set_site_device_param` (algo_ecdh, `param_type "27"`), read-modify-write via
  `get/setSafetySocParams`; recovered from a blutter decompile and confirmed by a read-back (discharge 20→15→20).

### Notes
- `b5` SOC-limit blob has a short settings variant (the limits) and a longer fast-frame variant (not the
  limits) on the same topic; the decoder accepts only the 4-byte settings blob.
- Tests use synthetic serials/ids only; offline + deterministic. Full `npm run verify` green (3632 tests).


src/transport/mqtt/solix-mqtt.ts
```diff
@@ -41,25 +41,129 @@ export interface SolixParamFrame {
...
 };
 
+/**
+ * Product-code prefix of the Smart Meter family that {@link SOLIX_METER_FIELD_NAMES} decodes. The table
+ * is meter-specific, so {@link solixReadings} applies its named fields ONLY to a frame from this family;
+ * a Solarbank (`AE103`) reporting the same `0xac` tag would otherwise be mislabelled `meterVoltageL1`
+ * with a nonsensical (negative-power) value. This is a product-code prefix used to select a decode
+ * table — not a model import — so the `transport ⊥ model` rule is untouched.
+ */
+export const SOLIX_METER_PRODUCT_PREFIX = "AE1X0";
```
martijnpoppen: **This branch carries a stale snapshot of #184/#186 and would revert three fixes landed there today.**

This PR merges `fix/solix-category-trim`, `fix/solix-meter-field-gating` and `feat/solix-power-station-and-site`, plus #184's *first* commit (`8621f8d`) — but not #184's later two (`9494aff`, `50bdbd3`) or today's `3f007e6`. Checked both trees:

| | this branch | #184 head (`3f007e6`) |
|---|---|---|
| meter prefix | `SOLIX_METER_PRODUCT_PREFIX` (string) | `SOLIX_METER_PRODUCT_PREFIXES` (array, JSDoc'd as the lockstep partner of `SOLIX_METER_MODELS`) |
| `solixReadings` | `productCode?` optional | `productCode` **required** (an optional arg silently drops names for existing callers) |
| meter L2/L3 provenance | `"apk"` ×6 | `"guessed"` ×6 |
| `LOAD_FRAME` fixture | absent | present — a real second AE1X0 capture |

The provenance one is the one I'd worry about: #184 downgraded L2/L3 from `apk` to `guessed` precisely because the phase assignment is block-ordering inference rather than something the app states, and that downgrade is the honest label. A merge resolution that takes this branch's side puts the overclaim back silently — no test fails, because the specs that pin it live in #184's version of the spec file.

Rebasing onto #184's current head before this goes further is the cheap fix. Worth doing before more review effort lands on it, since `solixReadings` is touched by both sides and will conflict.

Related, for sequencing rather than for this file: because the three merged branches are in the diff, GitHub shows 1565 insertions where only **941** are new here. Reviewing against `0d6bcd8` is what I did; anyone reading the PR page cold will re-review #185/#187 by accident.

---

src/transport/mqtt/solix-mqtt.ts
```diff
@@ -101,21 +205,76 @@ export function decodeSolixParamFrame(buf: Buffer): SolixParamFrame | null {
...
+    out.chargeLimit = b5[3]!;
+  }
+  // NOTE: tag `0xba` bit 0x20 does move with the ambient light, BUT only for OUR `set_device_attrs`
+  // write — an APP toggle changes the light via a separate command path (an `…/req` cmd-17 `a4`, see
+  // {@link SolixMqtt.handleCommand}) and does NOT update `ba` (proven live: the HA-switch write flips it,
+  // an app toggle leaves it stuck). So `ba` is STALE for app changes, and — arriving on every ~7s
+  // telemetry frame — it would clobber the correct value read from the command channel. It is therefore
+  // NOT emitted here: ambient-light state comes solely from the `…/req` `a4` command (which reflects the
+  // app's toggle AND echoes our own write), leaving the light unknown only until the first toggle.
+}
```
martijnpoppen, line 281: The protocol findings in this function are genuinely valuable — the `b5` fast-frame collision and the `ba`-is-stale-for-app-toggles discovery are exactly the kind of thing that gets rediscovered painfully later. But they're `//` blocks inside a function body, which `AGENTS.md` rules out ("Comments: JSDoc above the declaration, ground-truth, no inline body comments"), and the function's own JSDoc already covers SOC and temperature — so these two findings are the part that got left behind.

Both belong in that JSDoc as further bullets. The content transfers nearly verbatim:

```
 * - **SOC limits** (`dischargeLimit`/`chargeLimit`, %) come from tag `0xb5`'s SETTINGS-blob variant —
 *   type `0x04` with exactly 3 payload bytes, `[discharge, output cutoff, charge]`. Confirmed live:
 *   moving discharge 10%→5% moved `b5[1]` 0x0a→0x05 while charge held at `b5[3]`=0x64. The FAST
 *   telemetry frame (msgtype 0x05, ~7 s) also carries a `0xb5` type-`0x04` blob, but a 25-byte one
 *   whose bytes are not the limits — hence the exact length gate.
 * - **Ambient light** is NOT emitted here. Tag `0xba` bit `0x20` tracks only this SDK's own
 *   `set_device_attrs` write; an app-side toggle goes via an `…/req` cmd-17 `a4` and leaves `ba`
 *   unchanged, so on every ~7 s frame `ba` would clobber the correct value read from the command
 *   channel. State comes solely from {@link SolixMqtt.handleCommand}.
```

One line to drop rather than move, though — this bit:

> (An earlier note claimed the fast frame's b5 was a FLOAT — it is not; it is a longer 0x04 blob.)

That's "A JSDoc does not narrate its own history … not which guess was wrong". It's already recorded in `5ce907c`'s commit message, which is where `AGENTS.md` wants it.

---

src/transport/mqtt/topics.ts
```diff
@@ -143,8 +143,14 @@ export function parseSecureTopic(topic: string): ParsedTopic | undefined {
...
   paramInfo: string;
+  /**
+   * Settings/state the device pushes (SUBSCRIBE) — ff09 `state_info` frames. Same ff09 framing as
+   * `param_info` but the TAGS carry SETTINGS/targets (mode export limit, SOC limits, max_load, toggles),
+   * NOT live measurements — so it needs its own tag→name table, not the param_info one.
+   */
+  stateInfo: string;
   /** This device's command replies (SUBSCRIBE). */
   cmdRes: string;
   /** The device's requestDeviceInfo channel (PUBLISH only — the app arms reporting here). */
```
martijnpoppen: This JSDoc is falsified by this PR: `req` is no longer "PUBLISH only". `watch()` now adds `topics.req` to the subscribe list, and `onMessage` routes `topic.endsWith("/req")` into `handleCommand` — reading the app's own publishes back off the broker, which is the whole mechanism behind the ambient-light fix.

That's the more interesting half of the change and it deserves stating here, since this interface is where someone looks to find out what the topic is for:

```ts
/**
 * The device's requestDeviceInfo channel (cmd 17). PUBLISH to arm reporting; also SUBSCRIBE — the
 * broker copies the APP's publishes here to any co-subscriber, which is the only way to observe a
 * control the app changed that `param_info` does not reflect (ambient light, display timeout).
 */
```

Same note applies to `stateInfo`'s neighbours being accurate while this one isn't — worth a pass over the interface now that both directions are in play.

---

src/transport/mqtt/solix-mqtt.ts
```diff
@@ -41,25 +41,129 @@ export interface SolixParamFrame {
...
+ * Decode a `state_info` ff09 frame to named + raw settings values. Skips the header tags (`< 0xa5`:
+ * request marker, serial, timestamps). Each settings tag is emitted under `state_<hex>` (a plain number
+ * so it's watchable in a consumer while more tags get mapped) AND, when confirmed, under its name from
+ * {@link SOLIX_STATE_FIELD_NAMES}. Value is read type-aware: `0x05` float32, `0x02` u16, `0x01` u8, and
+ * `0x03` the whole-number settings byte (`payload[1]`).
+ */
+export function solixStateReadings(frame: SolixParamFrame): Record<string, number> {
+  const out: Record<string, number> = {};
+  for (const [tag, value] of frame.fields) {
+    if (tag < 0xa5 || !value || value.length < 2) continue;
```
martijnpoppen, line 164: Two sibling decoders over the same ff09 framing use different header cutoffs, and nothing says why:

- `solixStateReadings` (here): `if (tag < 0xa5) continue` — so `0xa5` **is** decoded.
- `solixReadings` (line 217): `if (tag < 0xa6) continue` — so `0xa5` is **not**.

If that's deliberate — `state_info` carries a real settings value at `0xa5` where `param_info` has a header tag there — it's worth one clause in the JSDoc, because the next person to touch either will assume the cutoffs should match and "fix" one of them. Your doc here says "Skips the header tags (`< 0xa5`: request marker, serial, timestamps)", which reads like the boundary is the same in both frames, and then the constant differs.

If it's not deliberate, `state_a5` is currently being emitted as a settings value from what is actually a header tag.

Cheapest way to pin whichever answer is right: a spec case asserting `0xa5`'s treatment in each decoder, so the two cutoffs can't silently drift apart.

---

martijnpoppen (commented):
Picked this up as a new PR while watching the others. Reviewed the **941 lines that are actually new here** (diffed against `0d6bcd8`, the last of the merged branches) rather than the 1565 GitHub shows — see the first inline comment on why those differ.

**Verified locally** on `5ce907c`: `tsc --noEmit` clean, `prettier --check` clean, all five guards pass, suite green at 200 files / 3632 tests.

The protocol work is strong, and two findings in particular are the kind that only come from real hardware: that `param_info`'s `0xba` tracks only *this SDK's* writes while an app-side toggle goes down a separate `…/req` path — so `ba` would clobber the correct value every 7 seconds — and that the fast telemetry frame carries a second, longer `0xb5` blob that isn't the SOC limits. Both are well evidenced and the length gate on `b5` is the right shape of fix. Splitting `SOLIX_SOLARBANK_FIELD_NAMES` and `SOLIX_STATE_FIELD_NAMES` into their own tables rather than extending the meter's is exactly the lesson from #186 applied, including keeping `0xab` distinct between the two.

### The thing I'd resolve before anything else

**This branch predates #184's last two rounds and would revert them on merge** — `SOLIX_METER_PRODUCT_PREFIX` (string, not the array), `solixReadings(productCode?)` (optional, not required), meter L2/L3 still `provenance: "apk"` (not `"guessed"`), and no `LOAD_FRAME` fixture. Table and detail inline. Rebasing onto #184's current head first would save re-reviewing this once the conflict is resolved, and `solixReadings` is touched on both sides so a conflict is certain.

### On sequencing

Stacking on three unmerged PRs makes this hard to land and hard to review. #185 and #187 both still have open findings from me (the category trim at ingest; `getSites`' unvalidated `site_id` cast; `solixProductFamily` re-implementing its own predicate) — all three are carried here unchanged, so they'd ride in with this. If the telemetry decode and the write controls can be separated, the decode half looks close to ready on its own, while the controls (six new write methods across HTTP and MQTT) are the part that wants the most scrutiny.

One question rather than a finding: the six controls (`setAmbientLight`, `setScreenOffTime`, `setPowerCutoff`, `setSafetySocParams`, `setDeviceAttrs`, `setDisplayTimeout`) all live in `transport/`, with nothing in the `solix.ts` members table. So a caller reaching them goes around the capability surface. Is that the intended end state for the Solix line, or a staging point before they're declared as members with `provenance`/`unverified` the way eufy writes are? Asking because `AGENTS.md`'s "a described write is one a caller can offer" and the unverified-write-throws rule both hang off the members table, and these currently sit outside it.

Three smaller things inline: the reserved-tag comments that want to be JSDoc (with one history-narration line to drop rather than move), the `req` topic doc that this PR falsifies, and a `0xa5`/`0xa6` cutoff mismatch between the two decoders.

---

src/transport/mqtt/solix-mqtt.ts
max246: Done (62ae01b) — both findings moved into `addSolarbankScalars`' JSDoc as bullets, near-verbatim from your text, and the "an earlier note claimed b5 was a FLOAT" history line dropped (it's in 5ce907c's message). Cleared the other `//` body comments in the function while I was in there.

src/transport/mqtt/topics.ts
max246: Fixed (62ae01b) — the `req` field JSDoc now says PUBLISH to arm + SUBSCRIBE to read the app's own publishes back (ambient light / display timeout), and I corrected the `solixDeviceTopics` builder doc that repeated "publish-only" too.

src/transport/mqtt/solix-mqtt.ts
max246: It's deliberate, now documented (62ae01b): `state_info` carries a settings value at `0xa5` (SOC) where `param_info` has a header tag, so the cutoffs are meant to differ. Added the clause to `solixStateReadings`' JSDoc and a spec that pins `0xa5`'s treatment in each decoder (a float at 0xa5 → `state_a5` present, `channel_a5` absent) so they can't silently drift.

src/transport/mqtt/solix-mqtt.ts
max246: Confirmed — this branch has the stale pre-review snapshot of #184 (string prefix, optional `productCode`, `apk` provenance, no `LOAD_FRAME`) and I don't want it reverting those on merge. Rather than duplicate #184's fixes here, the plan is to land #184/#185/#187 first (each now has its review addressed) and then rebase this onto beta-0.2.0 so it carries only the genuinely-new telemetry + controls — which also drops the ~624 lines of theirs that inflate the diff. Holding the rebase until they merge so I'm not resolving conflicts against moving heads; flagging it here so it's not merged before then.

max246:
On the members-table question (the six controls living in `transport/`, outside `solix.ts`): it's a **staging point**, not the end state. Every one of them is a **confirmed-live** write (each verified on a real AE103 — the SOC write via a read-back, the display timeout via the app reflecting it, etc.), so none are the unverified-throws case the members table guards. But you're right that they currently sit outside the capability surface, so a caller reaches them as transport methods rather than through `dev`.

The intended end state is to declare them as members with `provenance: "verified"` the same way eufy writes are — but the Solarbank's capability/members infrastructure isn't built yet (`solix.ts` only declares the `energyMeter` READ members; `battery` has no members table), and that migration is larger than this PR. So this ships the reads + confirmed controls now (reads ahead of the members surface, per "read paths may ship ahead of writes"), with the members-table migration as the follow-up that puts them behind `dev` with typed provenance. Happy to split the controls into their own PR if you'd rather review that migration as one unit — which also lines up with your "separate the decode from the controls" note.

martijnpoppen (commented):
**CI is red on `62ae01b` — it's the docs build, and it's been failing since `5ce907c` rather than being caused by this push.** Root-caused it; the fix is two `{@link}` targets.

There are two checks named `build` on this PR. The SDK one passes. The **docs** one (`docs/`: `typedoc && guard:docs && vitepress build`) fails at `typedoc`, which exits `4` on warnings:

```
[warning] Failed to resolve link to "SOLIX_ENDPOINTS.getSiteScene" in comment for SolixSiteScene
[warning] The comment for SolixMqtt.watch links to "onMessage" which was resolved but is not
          included in the documentation.
[warning] Found 0 errors and 2 warnings
##[error]Process completed with exit code 4.
```

Reproduced locally — `cd docs && npx typedoc` exits `4` on both `5ce907c` and `62ae01b`, and exits `0` on `beta-0.2.0`. So both warnings come from this branch:

- **`src/core/solix-types.ts:113`** — `SolixSiteScene`'s doc links `{@link SOLIX_ENDPOINTS.getSiteScene}`. `SOLIX_ENDPOINTS` isn't in the public surface, so the link can't resolve from a `core/` type. Naming the endpoint in prose, or linking `SolixClient.getSiteScene` (the exported method) instead, both fix it.
- **`src/transport/mqtt/solix-mqtt.ts:392`** — `SolixMqtt.watch`'s doc links `{@link onMessage}`, which is `private`, so it resolves but is never rendered. Worth noting `push-client.ts:96` uses the same `{@link onMessage}` without warning — the difference is that one sits on a private member whose doc isn't rendered, while `watch` is public and its doc is. So the fix is local to this comment, not a general ban: describe the filtering inline, or point at a public symbol.

### The part worth acting on beyond the two links

**`npm run verify` doesn't build the docs**, so the documented local gate cannot catch this class of failure:

```
format:check → typecheck → 5 guards → build → check:esm → check:snippets → typecheck:examples → test
```

That's why "verify green" on this branch and my own local runs both came back clean while CI was red — I ran the same chain and missed it for the same reason. Since `docs/` is generated from `src/` JSDoc, any `{@link}` added to shipped source can break it, and right now nothing local tells you. Adding the docs build to `verify` (or a `verify:docs` the contributing guide names) would close that gap — worth its own PR rather than widening this one.

Separately, thanks for the rebase plan on the stale-base thread — landing #184/#185/#187 first and then rebasing onto `beta-0.2.0` is exactly right, and holding the rebase until they merge so you're not resolving against moving heads makes sense. I've verified all three of those are now clear from my side, so they're not waiting on me.

---

martijnpoppen (commented):
That answers it — thanks. The staging-point reasoning holds: all six are confirmed-live writes, so the unverified-throws rule genuinely isn't in play, and "read paths may ship ahead of writes" covers shipping the reads before the members surface exists. The gap I was probing for isn't there.

**On splitting: yes, I'd split it** — as a reviewer's recommendation; the call is yours and the maintainer's.

Three reasons, in order of weight:

1. **The two halves have different risk profiles and different readiness.** The decode work is verifiable from fixtures and reasoning about frames, and it's close to done. The controls are six write paths whose correctness rests on live-device observation a reviewer can't reproduce — that's the part worth slow, focused review, and it reads very differently when it isn't sharing a diff with 900 lines of decoder.

2. **It makes the members migration reviewable as one unit.** If the controls land here as transport methods and then move behind `dev` later, that migration diff is "move six things and add a members table" with the original justification two PRs back. If the controls arrive *with* their members declaration, the provenance for each write and its typed surface land together, which is how the eufy side reads.

3. **It unwinds the stale-base problem faster.** The decode half rebases onto `beta-0.2.0` cleanly once #184/#185/#187 land; the controls can then stack on a stable base rather than everything waiting on one rebase.

If you'd rather not split, that's defensible too — the decode and the controls do share the `…/req` subscription discovery, and separating them means explaining that mechanism twice. In that case I'd just ask that the rebase come before the next review round, so the next reader isn't re-reading #185/#187 inside this diff.

Either way the docs-build failure is independent of the split and still wants those two `{@link}` fixes.

---

martijnpoppen (commented):
**CI is green on `f4f451c` — both checks pass, docs build included.** Confirmed rather than inferred: `cd docs && npx typedoc` now exits `0` with zero warnings here, where it exited `4` on `5ce907c` and `62ae01b`.

Both links took the right route:
- `core/solix-types.ts:113` → `{@link SolixClient.getSiteScene}`, the exported method, instead of the non-public `SOLIX_ENDPOINTS` constant. That keeps the pointer useful in the rendered docs rather than just silencing the warning.
- `solix-mqtt.ts` → the `{@link onMessage}` reference is gone.

Full gate green on this head too: `tsc --noEmit`, `prettier --check`, all five guards, 200 files / 3633 tests.

That clears the red. What's left on this PR is unchanged and both are yours to time rather than fix now: the **stale base** (you're holding the rebase until #184/#185/#187 land, which is the right call — all three are clear from my side and green, so they're only waiting on a maintainer), and the **split question**, where I've given my recommendation and the decision is yours.

---

martijnpoppen (commented):
**All three have landed — your rebase is unblocked.** `beta-0.2.0` is now `8ee6750`:

```
8ee6750 feat(solix): product-family classification + SolixSite system aggregate (#187)
9da1d9e fix(solix): detect a Solarbank's capabilities (category trailing space + newer models) (#185)
65f8bd2 feat(solix): AE1X0 meter bindings + gate the tag->name table to meter frames (folds #186) (#184)
```

Heads-up before you start: **the base moved more than you were expecting.** Two other PRs merged ahead of the Solix three this morning — `f3af454` (#195, the v2 thumbnail memory fix: new `jpeg-scan.ts` and `media-failure.ts`, a rewritten `decodeImageV2.ts`, `stored-image-cache.ts`, `store.ts`) and `20a0df1` (#190, the gtoken 401). Nine commits, 22 files, +1346/−332 in total.

`20a0df1` is the one that touches your area: it adds 5 lines to `solix-client.ts` recording that the same `ap_cloud_user_id` chain exists there and was deliberately *not* switched, with the Solix half left to you. Worth reading before the rebase — it turns what would have been a guess into a documented wire fact you can build on.

I test-merged the current base into this branch. The conflict set is exactly the stale-`#184` overlap, four files:

```
src/model/__tests__/solix-device.spec.ts
src/transport/http/__tests__/solix-client.spec.ts
```

In every one of those, **the base side is the newer work** — `SOLIX_METER_PRODUCT_PREFIXES` (array, lockstep-documented), `solixReadings(frame, productCode)` with the required parameter, L2/L3 at `provenance: "guessed"`, and the real `LOAD_FRAME` capture. This branch's copies predate all four. So resolving toward the base and re-applying only this PR's own additions on top (the Solarbank/`state_info` tables, `addSolarbankScalars`, the `…/req` subscription, the six controls) should be mechanical rather than a judgement call.

For reference, the other three open PRs (#180, #181, #192) all still merge clean against this base — this is the only one that needs the rebase.

Nothing else outstanding from me here beyond the split question, which is still yours to call.

---

martijnpoppen (approved):

PR #197, opened 2026-09-18
fix(p2p): let the session seal a set-payload, so a keyless standalone station can be sent one

martijnpoppen:
## What changed

`setPayload` without a `form` pins the `1350` envelope to level 2. Thirteen controls were in that
state on devices that can be their own keyless station; twelve now emit `form: "auto"` and leave the
seal to the session. Unchanged on a HomeBase, sendable at last on a standalone camera or doorbell.

| capability | members |
| --- | --- |
| `arming` | **mode** — the reported bug |
| `motion` | sensitivity, aiDetectType, humanOnlyAtNight, loiteringDetection |
| `camera` | nightVision, recordingQuality |
| `audio` | audioRecording |
| `ptz` | zoom |
| `light` | autoSpotlight |
| `battery` | powerSource |
| `doorbell` | dingdongVolume, dingdongRingtone |

The rule now has one home: `setPayload`'s own doc states the two conditions and the call sites just
satisfy them. Both have to hold.

1. **`mValue3` passed explicitly as 0.** The level-1 form writes `mValue3:0` itself while the level-2
   form defaults it to the sub-command, so a command passing 0 sends byte-identical JSON either way
   and `"auto"` changes only the seal. One that OMITS it would send an object nothing has captured —
   a new wire, not a downgrade.
2. **The device can be its own station.** A HomeBase, and an accessory whose session IS its
   HomeBase's, always holds a key. There, pinned is the honest behaviour: a keyless station is a real
   anomaly, and throwing says so where a silently-ignored level-1 frame would read as success.

**Eight stay pinned and now say why**, so the asymmetry reads as a decision rather than an oversight:
the doorbell status LED (1716), `ptz.setDefault` (6242) and the sensor test-mode write omit `mValue3`;
the HomeBase prompt volume is `available: isHomeBase`; the three siren writes and the entry sensor's
alarm volume are HomeBase-side. `ptz.setDefault`'s doc claimed the router picked its level by
topology, which was never true of a pinned set-payload — corrected, not merely amended.

**The level-1 branch now replays** `DIRECT_CMD_SENDS`× at 200ms, as `replayLevel2Send` and
`sendIntStringCommand` already do. It sent once, and twelve fire-and-forget writes just moved onto it:
unacknowledged datagrams, to the devices least able to spare one — no reply, no readback, nothing to
tell a caller the write was lost rather than refused.

## Why

Setting a guard mode on a standalone T8410 timed out every time, having sent nothing. From a caller's
log:

```
write eufy_arming_armingmode → arming.armingMode = 63
operation timed out after 20000ms: set armingMode
```

The level-2-only path does not fail on a station with no level-2 key — it WAITS for one:
`resolveSession({waitLevel2: true})` spends the full `LEVEL2_GRACE_MS` (25s), re-prompts, spends it
again, then throws `StationKeyUnavailableError` at ~50s. A caller bounding the call more tightly than
that reports a timeout and never learns the frame went nowhere; the 20s line above is such a bound,
firing partway through the first grace.

A T8410 never negotiates that key — its session is level-1, the fact `access.ts` already records
beside camera power 1035 ("verified live on standalone T8410 = L1"). So these writes were not slow,
they were unsendable, and the wait was the whole of the symptom.

## Evidence

- **The level-1 seal of this envelope is confirmed on hardware.** `rtsp` credentials (1287) already
  ship `"auto"` for exactly this reason, and `sendSetPayloadEnvelope`'s doc records "verified live: a
  standalone camera accepts the level-1 form".
- **The JSON is identical by construction, not by assumption.** Both senders build
  `{account_id, cmd, mChannel, mValue3, payload}` in that order; `sendSetPayload` hardcodes
  `mValue3: 0` and `sendControlLevel2` takes `mValue3 ?? cmd`. Every downgraded command passes 0, so
  the object the station parses is the captured one under either seal. That is condition 1, and it is
  why the four `mValue3`-omitting commands stayed pinned.
- **Nothing here changes a param's own evidence.** No wire id, value domain or polarity moved;
  `ARMING_MODE_WIRE` is untouched. The only device whose behaviour changes is the one that could not
  be sent the frame at all.
- **The HomeBase path is provably unchanged**: a session holding a key still takes the `l2` branch,
  pinned by a spec.

## Verified how

- [x] `npm run verify` is green locally — 202 files, 3673 tests
- [ ] Ran against a real device — **not yet; see below**
- [x] Docs updated — the rule in `setPayload`, the reason at each site that stays pinned

Three transport specs in `level2-wait-ownership.spec.ts` pin what the seal does per station kind: a
keyless own-session station is sent the level-1 frame, replayed, without spending the per-call grace;
the same frame with no `form` still refuses with `StationKeyUnavailableError` after spending both
graces (the behaviour the still-pinned writes keep); a keyed session still seals level-2. The replay
assertion is mutation-checked — reverting the loop fails it.

A downstream consumer of this SDK also typechecks and passes its own 619 tests against the built
package, so the `Command` shape change carries cleanly.

## Unverified or deferred

- **No level-1 hardware run for these twelve params.** The seal is confirmed on this envelope and the
  JSON is identical by construction, but nobody has watched *these* commands land on a keyless device.
  Stated plainly because the alternative is not "throw": today they hang for ~50s and then throw, so
  this trades a confirmed dead end for a well-grounded inference. The first real test is a T8410.
- **`light.ts`'s `TODO(L2-on-standalone)` is untouched** and still correct: brightness/color-temp are
  pinned `"direct-binary"`, a genuinely different frame whose level-1 wire is unconfirmed — not the
  same-JSON-different-seal case this PR is about. Its pointer to a README section
  ("Control-command encryption level") is stale; no such section exists in any README here.
- **The four `mValue3`-omitting commands stay unreachable on a keyless station** until something
  captures their level-1 form. Doorbell status LED and `ptz.setDefault` are the two a standalone
  device would actually want.

https://claude.ai/code/session_01RdfJxzv7cDWeFGstFr9eaw

---
_Generated by [Claude Code](https://claude.ai/code/session_01RdfJxzv7cDWeFGstFr9eaw)_


src/transport/p2p/__tests__/level2-wait-ownership.spec.ts
```diff
@@ -126,6 +128,55 @@ describe("resolving a session defers the level-2 wait to the session", () => {
   });
 });
 
+/**
+ * The `1350` SET_PAYLOAD envelope on a station that will never hold a level-2 key.
+ *
+ * Pinned to level 2 this frame is not slow on such a station, it is UNSENDABLE — and it does not say so:
+ * the required-key path spends the full grace, re-prompts, spends it again, and only then refuses, so
+ * every caller with a shorter bound reports a timeout instead. That is what reached a Homey user as a
```
max246: AGENTS.md consumer-agnostic hard rule — genericise the named consumer. e.g.

> …every caller with a shorter bound reports a timeout instead. That is what a caller with a short timeout bound sees as a guard-mode write that "times out" on a T8410.

(The model code `T8410` is fine — it's a public SKU, not PII; it's only the "Homey user" that needs to go.)

max246 (commented):
Reviewed against AGENTS.md — the fix itself is solid:

- Correct root cause: the required-key path doesn't fail on a keyless standalone, it *waits out* both graces and re-prompts, so callers with a shorter bound see a timeout rather than a refusal. `form: "auto"` hands the seal to the session.
- The rule has one home — `setPayload`'s two-condition JSDoc — and the call-sites just satisfy it. The 12-switch / 8-stay-pinned split is justified per member (explicit `mValue3:0` = byte-identical either seal, vs. omitted = an uncaptured new wire).
- The l1 branch reuses the existing `DIRECT_CMD_SENDS`×@200ms replay pattern rather than inventing one.
- Evidence-grounded (T8410 = L1 cross-referenced from `access.ts`), decorrelation preserved (`form` flows model→transport via the `Command`, no cross-import), and the `ptz.setDefault` doc is *corrected* (the false "router picks the level by topology") rather than amended — exactly what the JSDoc ground-truth rule wants.
- The new p2p spec covers the three behaviours that matter (l1 replay without a per-call wait, l2 still sealed, no-`form` still refuses).

One thing to fix before this can be approved — the **consumer-agnostic hard rule** (AGENTS.md: *"Source, JSDoc, guides, examples, commit messages **and PR bodies** never name a specific consumer or its stack … Say 'caller' or 'host'."*). `guard:consumer-agnostic` doesn't scan specs or the PR body, so CI is green, but there are three references to a specific consumer:

- the spec JSDoc (inline comment below);
- the PR description: *"a Homey app bound firing partway through the first grace"* and *"Downstream: the Homey app typechecks and its 619 tests pass"* — please genericise to "a caller" / "a downstream consumer".

Everything else looks great — happy to approve once the consumer references are generic.

src/transport/p2p/__tests__/level2-wait-ownership.spec.ts
martijnpoppen: Fixed in 018b86a, and you're right on the rule — it says "commit messages and PR bodies", so the spec's exclusion is the guard's scope, not the rule's.

Rewrote the JSDoc to state the behaviour rather than the incident, which also gets it off the "states the declaration, not its audience" rule it was brushing:

> the required-key path spends the full grace, re-prompts, spends it again, and only then refuses, so a caller bounding the call more tightly than that reports a timeout and never learns the frame went nowhere. A T8410 is such a station.

PR body done too — the "Why" section now attributes the log to "a caller's log" and calls the 20s figure "such a bound", and the verification line reads "a downstream consumer of this SDK". `grep -rniE '\bHomey\b' src/` is clean.

Two things you may want to weigh in on, both beyond what you flagged:

1. **A commit message on this branch still carries it.** bde372b's body ends "Homey app typecheck + 619 tests green against the synced build." The base branch merges rather than squashes, so that lands in history permanently. Fixing it means amending and force-pushing — cheap here (nothing to dismiss, CI is a minute), but it rewrites commits you've already reviewed, so I'd rather you call it than do it unasked. Say the word and I'll amend.

2. **Two pre-existing hits the guard also can't see**, in `shared-live-reuse-watch.spec.ts:13,16` — "a negotiated HomeKit session" and "30 s HomeKit timeout", both prose rather than eufy's own identifier-shaped wire vocabulary. Not this PR's to widen into, but worth a follow-up, and possibly worth dropping `--exclude-dir='__tests__'` from the guard so specs stop being a blind spot.

---

max246 (approved):
The consumer-agnostic note is resolved — the spec JSDoc now reads "a caller bounding the call more tightly than that" with no named consumer (T8410 stays, correctly a public SKU), and a broad scan of the diff is clean of stack names. Everything else held up on the first pass: correct root cause (the keyless station waits out both graces rather than failing, so a short-bound caller sees a timeout), the rule centralised in `setPayload`'s two-condition doc with the call sites just satisfying it, the `DIRECT_CMD_SENDS` replay reuse, the T8410=L1 cross-reference, the corrected `ptz.setDefault` doc, and the three-behaviour test coverage. CI green, mergeable. Approving.

PR #198, opened 2026-09-18
feat(client): let a caller pin the acting account name

martijnpoppen:
## What changed

`MegaClientConfig.accountName` pins the name commands attribute themselves to. Trimmed; a blank value counts as unset, so the login email's local-part stays the default and nothing changes for a caller that does not set it.

It reaches the three places `ctx.accountName` already went: the guard-mode write's `user_name` (`arming`), the HomeBase alarm's `user_name` (`siren`), and the lock's acting username (`A4`). JSDoc on `CommandContext.accountName` and `armingCommand` updated to say "acting name" rather than "email local-part", since the two are no longer the same thing.

## Why

A device records who sent a command, and the eufy app shows that name in its activity history. With the local-part hard-wired, every write from a non-phone caller landed in that history under the account holder's own name — indistinguishable from the same person doing it by hand, which is the one distinction the record exists to make.

It is a label, not an identity: the session still authenticates on the token and on the device record's member ids, and this touches neither.

## Evidence

n/a for the wire — the frame is unchanged. `SET_ARMING`'s `{mode_type, user_name}` payload and the lock's `A4` field are the captures already documented on `ARMING_CMD.SET_ARMING` and `LOCK_MEMBERS`; this only changes which string goes into a field that was already there.

## Verified how

- [x] `npm run verify` is green locally (3673 tests, all guards, build + ESM smoke)
- [ ] Ran against a real device — no, see below
- [x] Added or updated specs — `transport/http/__tests__/device-identity.spec.ts`: the local-part default, the no-`@` fallback, an override winning, and trim/blank handling

## Unverified or deferred

**Whether a device validates `user_name` is untested, and this makes that matter more.** The existing note on `armingCommand` says the real app's own capture showed a masked value (`"max***"`) where the SDK sends the unmasked local-part, and that the field is almost certainly attribution rather than something the device checks. That reasoning is unchanged, but a caller can now put an arbitrary string there — so if a station did validate the field, an override would be the thing that surfaces it. Nothing here is fire-and-forget in the unverified-write sense: the guard-mode write reports its own `MODE_SWITCH` push, so a rejected mode is observable rather than silent.

https://claude.ai/code/session_01SFuZTXW5EejGCecRgYnxSh

---
_Generated by [Claude Code](https://claude.ai/code/session_01SFuZTXW5EejGCecRgYnxSh)_


src/model/capabilities/arming.ts
```diff
@@ -175,9 +175,9 @@ function armingModeOf(v: boolean | number | string): ArmingMode | undefined {
 /**
  * Build the {@link ARMING_CMD.SET_ARMING} write intent, or throw if the context has no account
  * identity. `user_name` — the real app's own capture showed a MASKED value (`"max***"`, presumably
```
max246, line 177: Optional / pre-existing nit (not introduced by this PR): `"max***"` is a masked-but-recognisable fragment of a real account local-part. AGENTS.md — *"No real device data … account and user ids … Redact to synthetic placeholders."* Since you're already editing this JSDoc block, worth swapping to a synthetic like `"user***"`.

max246 (approved):
Reviewed against AGENTS.md — clean and well-tested. Transport config field + getter with model-side JSDoc-only touches (no layer violations); consumer-agnostic wording ("caller", "the client pins a different label"); the trimmed / blank-counts-as-unset semantics are fully covered by the new spec (default, no-`@`, override-wins, trim, blank, empty). The "an override is a different LABEL for the same account, not a different identity" note is a nice security clarification. Approving.

One optional, pre-existing nit inline.

Issue #201, opened 2026-09-18
Smart Lock: Param 6000 (state), Param 6001 (battery), and legacy P2P cmd 1961 actuation (T8520 vs T85L0)


matt1981m:
### Model code
T8520, T85L0

### Topology
Standalone (Wi-Fi)

### What is missing
Two related items for the `lock` capability:

1. **Stable State & Battery Parameters (Param 6000 & 6001)**:
   In `src/model/capabilities/lock.ts`, `locked` is declared with placeholder `param: 1200` marked `guessed` (*"no stable state param to key off; placeholder id pending verification"*).
   Live hardware reporting confirms:
   - **Param 6000** is the stable lock state parameter across models: `"4"` = locked, `"3"` = unlocked.
   - **Param 6001** is the lock battery percentage (e.g. `"31"` = 31%).

2. **Actuation Compatibility (`T85L0` vs `T8520`)**:
   `actuate()` in `src/model/capabilities/lock.ts` currently dispatches only `ff09-actuate`:
   - On modern locks like the **T85L0 (Smart Lock C33)**, `ff09-actuate` succeeds immediately.
   - On previous-generation locks like the **T8520 (Smart Lock Touch & Wi-Fi)**, `ff09-actuate` times out. These older Wi-Fi locks do not respond to `ff09` framed commands; they require legacy P2P command 1961 (`CMD_SMARTLOCK_REMOTE_CONTROL` / `APP_CMD_SMARTLOCK_REMOTE_CONTROL`).

#### Realtime `lockState` Event Payload (Captured live on T85L0):
```json
{
  "event": "lockState",
  "eventType": 257,
  "params": [
    {
      "param_type": 6000,
      "param_value": "3"
    },
    {
      "param_type": 6001,
      "param_value": "31"
    },
    {
      "param_type": 1101,
      "param_value": "31"
    }
  ]
}
```
- `6000`: `"3"` = unlocked, `"4"` = locked (matches `eventType` 257 = manual unlock, 264 = lock).
- `6001`: `"31"` = battery 31%.

#### Raw Parameter Snapshot (from Device Property Map):
```json
{
  "unknown_6000": "4",
  "unknown_6001": "31"
}
```
Currently, because 6000 and 6001 are missing from the param dictionary, they surface as `unknown_6000` and `unknown_6001`.

#### Actuation Behavior:
- **T85L0**:
  Dispatched `ff09-actuate` $
[...]

PR #202, opened 2026-09-19
feat(lock): map param 6000 (locked state) and 6001 (battery)

matt1981m:
Resolves #201

### Description
Replaces the placeholder param 1200 on \`LOCK_MEMBERS.locked\` with verified param 6000, and adds smart-lock battery param 6001 as a read-alias on \`LOCK_MEMBERS.battery\`.

### Changes
1. **Parameter Dictionary (\`src/model/param-dictionary.ts\`)**:
   - Registered \`6000\` (\`lockStatus\`, number, verified).
   - Registered \`6001\` (\`lockBattery\`, number, verified).
2. **Lock Capability (\`src/model/capabilities/lock.ts\`)**:
   - Updated \`LOCK_MEMBERS.locked\` to use verified param \`6000\` with \`coerce: (raw) => String(raw).trim() === "4"\` (wire values: \`4\` = locked, \`3\` = unlocked).
   - Updated \`LOCK_MEMBERS.battery\` with \`readAliases: [{ paramType: 6001 }]\` while retaining standard param \`1101\`.
   - Added \`6000\` to \`LOCK.detection.evidenceParams\` so smart locks self-reporting param 6000 resolve the capability.
3. **Tests (\`src/model/capabilities/__tests__/lock.spec.ts\`)**:
   - Asserted schema mapping for param 6000 and alias 6001.
   - Added \`Device\` integration tests verifying lock state decoding (\`4\` -> \`true\`, \`3\` -> \`false\`), battery decode over alias 6001, fallback to 1101, and lock capability detection.

### Verification
- \`npm run verify\` passed all checks (\`format:check\`, \`typecheck\`, all guards, \`build\`, \`check:esm\`, \`check:snippets\`, and full test suite across all 202 test files).
- No real device serial numbers, accounts, or PII included.


max246:
This is the one we're going with for the param-6000/6001 work 👍 — I reviewed it against the repo conventions and ran `npm run verify` on the same changes locally, all green. Two things you did that I'd since dropped from my own attempt (now closed, #204): `6000` in `evidenceParams`, and the `Device.fromRecord` integration tests — both better than what I had.

The only piece not covered here is the **push** side — making the `lockState` event itself carry a decoded `locked` boolean, so an external change (keypad/thumbturn/app) reflects without waiting for a poll. I split that into **#205** as a small follow-up on top of this (it only touches the event mapping, so the two are orthogonal and merge in either order).

Nice work on the hardware testing — this closes the poll-state half cleanly.

nit (non-blocking): CI doesn't seem to have reported on this branch yet — worth a push/re-trigger so the `build` check is green before merge.

max246:
@matt1981m — rather than a separate follow-up, it'd be cleaner if this PR carried the whole fix, so it's one complete thing under your name. The only missing piece is making the `lockState` event itself carry the decoded `locked` (the push side, so an external keypad/thumbturn/app change reflects without a poll). It's small and orthogonal to your param work — here's the exact drop-in. I'll close my #205 in favour of you folding this in.

**`src/model/capabilities/lock.ts`** — add two imports at the top:

```ts
import { LockPushEvent } from "../push-events.js";
// and add InboundSignal to the existing type import from "./types.js":
import type { AvailabilityContext, CapabilityModule, CommandContext, InboundSignal } from "./types.js";
```

Add this function just above `export const LOCK`:

```ts
/**
 * The `locked` boolean a `lockState` push carries, derived from which `LockPushEvent` fired: the seven
 * `*_LOCK` actions (262..268) → `true`, the six `*_UNLOCK` actions (257..261 + 269) → `false`. The rest
 * of the 257..771 range — alarms, low-power, offline/online, OTA/status — is not a lock transition and
 * contributes no `locked` field, so a consumer reads state only from an event that actually carries one.
 */
function decodeLockTransition(signal: InboundSignal): Record<string, unknown> {
  if (signal.source !== "push" || signal.eventType === undefined) return {};
  const e = signal.eventType;
  if (e >= LockPushEvent.MANUAL_LOCK && e <= LockPushEvent.TEMPORARY_PW_LOCK) return { locked: true };
  const unlocked =
    (e >= LockPushEvent.MANUAL_UNLOCK && e <= LockPushEvent.APP_UNLOCK) || e === LockPushEvent.TEMPORARY_PW_UNLOCK;
  return unlocked ? { locked: false } : {};
}
```

And add `derive` to the existing events mapping:

```ts
  events: [{ source: "push", match: [257, 771], emit: "lockState", derive: decodeLockTransition }],
```

**`src/model/capabilities/__tests__/decode-event.spec.ts`** — after the existing "lock events resolve via a RANGE" test:

```ts
    it("lockState carries a decoded `locked` boolean for the (un)lock actions", () => {
      expect(push(262)[0].payload).toMatchObject({ locked: true });  // MANUAL_LOCK
      expect(push(268)[0].payload).toMatchObject({ locked: true });  // TEMPORARY_PW_LOCK
      expect(push(257)[0].payload).toMatchObject({ locked: false }); // MANUAL_UNLOCK
      expect(push(269)[0].payload).toMatchObject({ locked: false }); // TEMPORARY_PW_UNLOCK
    });

    it("a non-transition lock event (alarm/status) still emits lockState but carries NO `locked`", () => {
      for (const et of [513, 769, 771]) {
        const p = push(et)[0];
        expect(p).toMatchObject({ event: "lockState" });
        expect(p.payload).not.toHaveProperty("locked");
      }
    });
```

That's the whole thing — `npm run verify` stays green with it. Once you push it here, this PR closes both halves (poll + push) and I'll close #205. Thanks again!

max246 (changes requested):
see the comments

max246:
To consolidate what's needed here (backing the changes-requested) — the param-6000/6001 work is solid and verified; just one required addition plus two small optional notes.

### Required — fold in the `lockState` push decode
The poll side (param 6000) is done here, but the **push side** is missing: the `lockState` event still fires by name only, so a consumer can't tell a lock from an unlock without re-deriving the range itself. That's the piece that reflects an external change (keypad/thumbturn/app) without waiting for a poll. I posted the exact drop-in [in the comment above](https://github.com/mega-yfue/eufy-sdk/pull/202#issuecomment-5740136842) — the `decodeLockTransition` function, the two imports, the one-line `derive` on the events mapping, and two `decode-event.spec.ts` tests. Verify stays green with it. (I closed my own #205 so this lands as one complete PR under your name.)

### Optional (non-blocking)
- **Confirm param 6000 is lock-exclusive.** `evidenceParams: [6000]` means *any* device reporting 6000 resolves as a lock — reasonable, since the whole 60xx block is smart-lock vocabulary, but worth a sanity check against your fleet that nothing non-lock reports it.
- **Dict names vs property names** (`lockStatus`/`lockBattery` vs `locked`/`battery`): purely cosmetic — a resolved lock's `spec.name` always wins in `applyParams`, so the dict names are never surfaced. Your distinct names are actually the safe choice (naming 6001 `battery` would duplicate 1101's dict entry). Leave as-is; just noting it so it's not a surprise later.

Once the push decode is in and CI is green, this is good to go — nice work on the hardware verification.

matt1981m:
> To consolidate what's needed here (backing the changes-requested) — the param-6000/6001 work is solid and verified; just one required addition plus two small optional notes.
> 
> ### Required — fold in the `lockState` push decode
> The poll side (param 6000) is done here, but the **push side** is missing: the `lockState` event still fires by name only, so a consumer can't tell a lock from an unlock without re-deriving the range itself. That's the piece that reflects an external change (keypad/thumbturn/app) without waiting for a poll. I posted the exact drop-in [in the comment above](https://github.com/mega-yfue/eufy-sdk/pull/202#issuecomment-5740136842) — the `decodeLockTransition` function, the two imports, the one-line `derive` on the events mapping, and two `decode-event.spec.ts` tests. Verify stays green with it. (I closed my own #205 so this lands as one complete PR under your name.)
> 
> ### Optional (non-blocking)
> * **Confirm param 6000 is lock-exclusive.** `evidenceParams: [6000]` means _any_ device reporting 6000 resolves as a lock — reasonable, since the whole 60xx block is smart-lock vocabulary, but worth a sanity check against your fleet that nothing non-lock reports it.
> * **Dict names vs property names** (`lockStatus`/`lockBattery` vs `locked`/`battery`): purely cosmetic — a resolved lock's `spec.name` always wins in `applyParams`, so the dict names are never surfaced. Your distinct names are actually the safe choice (naming 6001 `battery` would duplicate 1101's dict entry). Leave as-is; just noting it so it's not a surprise later.
> 
> Once the push decode is in and CI is green, this is good to go — nice work on the hardware verification.

Folded in the push-event decoding and test cases from #205. npm run verify passes cleanly with all checks green!

max246 (approved):
Approving — the required change is in and correct. 👍

**The push decode is folded in properly** (`81fddfa`):
- `decodeLockTransition` derives `locked` from the event: `MANUAL_LOCK`(262)..`TEMPORARY_PW_LOCK`(268) → `true`, `MANUAL_UNLOCK`(257)..`APP_UNLOCK`(261) + `TEMPORARY_PW_UNLOCK`(269) → `false`, everything else in 257..771 → no `locked` field. I checked the enum: those ranges are clean and contiguous (257–261 unlock, 262–268 lock, 269 the lone trailing unlock), so nothing is mis-bucketed, and it correctly guards on `source === "push"`.
- Wired through the supported `derive` hook on the event mapping, so `lockState` now carries the decoded boolean — an external keypad/thumbturn/app change reflects on the push without waiting for a poll, which was the missing piece.
- Both tests I asked for are there: one asserting `locked` true/false on the (un)lock actions, one asserting a non-transition event (LOW_POWER / STATUS_CHANGE / LOCK_ONLINE) still emits `lockState` but carries no `locked`.

**Verified locally against `beta-0.2.0`:** `npm run verify` green (3680 tests, 202 files) **and** `cd docs && npm run build` green (no typedoc link breakage).

The two optional notes stay optional — the `evidenceParams: [6000]` fleet sanity-check and the cosmetic dict-name point, neither blocking. Nice work on the hardware-verified param mapping; good to merge.

PR #204, opened 2026-09-19
feat(lock): read lock state from param 6000 + carry `locked` on the lockState event

max246:
Closes the two read-path gaps @matt1981m found on live hardware (T85L0 / C33) while testing the HA lock entity ([ha-eufy-sdk#41](https://github.com/mega-yfue/ha-eufy-sdk/pull/41)), tracked in #201. Doing the decode **here** keeps device truth in the SDK — the alternative was param ids and push-code sets hardcoded into the HA integration's coordinator.

## What

1. **`locked` reads the stable state param 6000** (`4` = locked, `3` = unlocked) via a `coerce`, replacing the guessed placeholder `1200` — which never reported and, it turns out, collided with `language` in the param dictionary. The coerced boolean is the *stored* value, so a cloud poll surfaces `locked` directly instead of a raw `unknown_6000`; the getter and the broadcast state agree. Evidence-gated as ever — a lock that reports no 6000 leaves the getter uninstalled and falls back to the event below.

2. **The `lockState` event carries a decoded `locked` boolean.** A `derive` on the existing `257..771` range maps the seven `*_LOCK` actions (262..268) → `true` and the six `*_UNLOCK` (257..261 + 269) → `false`. Alarms, low-power and status/OTA events stay `lockState` with **no** `locked` field, so a consumer reads state only from an event that actually carries a transition.

3. **Param 6001 (lock battery %) as a read alias** for `battery`, so a modern lock reporting battery on 6001 rather than the fleet-wide 1101 still resolves `dev.lock().battery`.

Params 6000/6001 added to the security param dictionary (both `verified` — live on a T85L0).

## Why this shape

- `coerce` (ingest-time) not `decode` (per-read): the HA integration reads the **stored** `state.locked` the bridge broadcasts, so the correction has to land on the stored value, not only in the getter.
- `derive` returns `{}` for non-transition events, so nothing is invented — same evidence bar as the rest.

## Impact for the integration

With this, [ha-eufy-sdk#41](https://github.com/mega-yfue/ha-eufy-sdk/pull/41)'s "state reverts to unknown after the 20s reconcile" disappears (reconcile reads a real `locked`), and reflecting an external change (keypad/thumbturn/app) shrinks to one glue line that writes `evt.locked` into coordinator state — no param ids in HA.

## Out of scope

Write-path parity for legacy locks (P2P cmd 1961, e.g. T8520) stays a separate task in #201 — unverified until someone can actuate one, so per the never-guess-a-write rule it doesn't ship here.

## Tests

`npm run verify` green (3673 tests, all guards). New coverage: the 6000 coerce (4→true, 3→false, other→false), the 6001 battery alias, and the `lockState` derive (lock/unlock → `locked`, alarm/status → none).

Refs #201.


max246:
Closing in favour of @matt1981m's #202, which landed first and does the param 6000 + 6001 work on real hardware (and adds `6000` to `evidenceParams` + `Device`-level integration tests, both nicer than what I had here).

The one piece #202 doesn't cover is the **`lockState` event carrying a decoded `locked` boolean** — the bit that reflects an external change (keypad/thumbturn/app) without a poll. I've split that out as a small follow-up so it stacks cleanly on top of #202 rather than duplicating it: **#205**.

PR #205, opened 2026-09-19
feat(lock): carry a decoded `locked` boolean on the lockState event

max246:
Follow-up to @matt1981m's #202 — the **push half** of the smart-lock state fix (#202 does the poll/param-6000 half). Split out so it stacks cleanly on #202 rather than duplicating it; the two are orthogonal (this touches only the event mapping, not the `locked`/`battery` members or the param dictionary), so they merge in either order.

## What

The `lockState` push already fires for every `LockPushEvent` in 257..771, but only by *name* — a consumer had to re-split the range itself to tell a lock from an unlock. This adds a `derive` that decodes it at the source:

- `*_LOCK` actions (262..268) → `locked: true`
- `*_UNLOCK` actions (257..261 + 269) → `locked: false`
- alarms / low-power / status / OTA → `lockState` with **no** `locked` field (not a transition, so nothing invented)

## Why it matters

With this, a consumer reflects an external lock change (keypad, thumbturn, auto-lock, app) from the event alone — no push-code table on its side. Together with #202's param-6000 read, it closes both state-tracking gaps [ha-eufy-sdk#41](https://github.com/mega-yfue/ha-eufy-sdk/pull/41) hit on live hardware: the poll surfaces `locked`, and the push carries transitions live.

## Tests

`npm run verify` green (3671 tests). New coverage in `decode-event.spec.ts`: the (un)lock actions carry `locked`, and a non-transition event (alarm/status) carries none.

Refs #201, #202.


max246:
Closing so this lands as part of @matt1981m's #202 instead — one complete PR under the original author rather than a maintainer-side follow-up. Handed the exact drop-in over on #202; the code + tests live in that comment.

PR #206, opened 2026-09-19
docs(solix): cross-reference field names to Anker's official Modbus registers

max246:
Anker published an official HA integration for the newer **Modbus-TCP** Solix generation, whose per-device YAML register maps name every quantity. Those registers **independently corroborate** the field meanings this SDK reverse-engineered from the cloud-MQTT `ff09` telemetry — and one of their configs is literally a *Solarbank 4 E5000 Pro* (the same product as our `AE103`, a newer hardware rev).

This annotates each of our meter/Solarbank field names with its equivalent Anker register (comment-only, **non-breaking** — no rename), so each reverse-engineered name has the vendor's own naming as evidence rather than just assertion.

## Meter (Smart Meter Gen 2)
| ours | Anker register |
|---|---|
| `meterPowerL1/L2/L3` | `primary_phase_1/2/3_active_power` |
| `meterPowerTotal` | `primary_total_active_power` |
| `meterVoltageL{1..3}` | `primary_phase_{1..3}_voltage` |
| `meterCurrentL{1..3}` | `primary_phase_{1..3}_current` |
| `meterImportEnergy` | `primary_total_forward_active_energy` |
| `meterExportEnergy` | `primary_total_reverse_active_energy` |

## Solarbank
`photovoltaicPower`↔`pv_power`, `chargePower`↔`battery_charging_power`, `dischargePower`↔`battery_discharging_power`, `gridInputPower`↔`grid_import_power`, `homeLoadPower`↔`load_power`, signed `batteryPower`↔`battery_charging_power − battery_discharging_power`.

## New public export: `SOLIX_MODBUS_EMS_MODES`
Anker's Modbus `operating_mode` register enumeration (`{0 self_consumption, 1 tou, 3 third_party, 4 custom, 5 socket_overlay, 6 smart, 7 dynamic_pricing}`, value 2 unassigned) exposed as a reference table. **It is NOT a decoder for this SDK's ff09 `mode`** (state_info `0xa9`): the AE103 ff09 numbering is different on every value (1 custom, 2 self-consumption, 4 rapid charge, 7 smart, 8 dynamic tariff), recorded on the `0xa9` field comment. The public name says which hardware rev it belongs to so a caller can't mistake it for the ff09 decode; the two fold together only if a live `0xa9` correlation capture proves the numbers match.

## Unnamed load slots
Anker's meter also exposes `reactive_power`, `power_factor`, per-phase energy, and a **second CT channel** — none carried by our single-channel `ff09` frame. For the still-raw `channel_b2` / `channel_b7` (they read ~0 at idle, non-zero under load): `0xb2` (0.009 under S≈328 VA) is dimensionally consistent with **power factor** and rules **reactive power** out (a VAR slot would read in the hundreds); `0xb7` stays fully open. Both left raw regardless — a single frame doesn't pin them.

Note: their devices are a different, newer generation (local Modbus) than our cloud-MQTT `AE103`/`AE1X0`, so this is a semantic cross-reference, not the same hardware. `npm run verify` **and** `cd docs && npm run build` (typedoc + guard:docs + vitepress) both green.


src/transport/mqtt/solix-mqtt.ts
```diff
@@ -140,12 +154,43 @@ export const SOLIX_SOLARBANK_FIELD_NAMES: Readonly<Record<number, string>> = {
...
+ * Solarbank 4 E5000 Pro (register `operating_mode`, gated by the `0x8006` capability mask). Value `2`
+ * is unassigned there, so there are seven modes across values `{0,1,3,4,5,6,7}`, not eight.
+ *
+ * EVIDENCE NOTE: the enumeration is authoritative for Anker's newer Modbus-TCP hardware rev; this SDK
+ * decodes the OLDER cloud/MQTT `AE103` over ff09, and its `0xa9` was not independently confirmed to use
+ * these same numbers (an earlier reverse-engineered read suggested a different set). Adopting Anker's
+ * official numbering here on that basis; a live "set each mode in the app, read `0xa9`" capture should
+ * confirm it, and if a value disagrees this map is the one place to correct.
+ */
+export const SOLIX_EMS_MODES: Readonly<Record<number, string>> = {
```
martijnpoppen: This map disagrees with the numbering this repo previously recorded for `0xa9` — on **every single value**, not just some.

The line this PR rewrites (`60874a8`, the AE103 telemetry commit) said:

> `mode` — current operating (EMS) mode (1 custom, 2 self-consumption, 4 rapid charge, 7 smart, 8 dynamic tariff)

Running the old table's values through the new map, on this branch:

| observed value | what this repo recorded for AE103 | `SOLIX_EMS_MODES[v]` |
|---|---|---|
| 1 | custom | `timeOfUse` |
| 2 | self-consumption | **`undefined`** |
| 4 | rapid charge | `custom` |
| 7 | smart | `dynamicTariff` |
| 8 | dynamic tariff | **`undefined`** |

Not one value survives. The JSDoc is honest about this ("an earlier reverse-engineered read suggested a different set"), but the conclusion it draws — adopt Anker's numbering anyway — is the risky direction. If the old read was right for `AE103`, then every consumer that labels an ff09 `mode` with this map gets a **confidently wrong label**, silently. There is no in-band way for them to notice: `0` is a valid key here, so a self-consumption device reporting the old `2` returns `undefined` while a device reporting `0` reads as `selfConsumption`.

A wrong label is worse than no label. Two options that keep the (genuinely useful) vendor enumeration without asserting it over the wire this SDK actually decodes:

1. **Name it for the hardware it's from** — `SOLIX_MODBUS_EMS_MODES`, documented as the Modbus-TCP rev's `operating_mode` register. Then nobody reaches for it to decode an ff09 `mode`, and it's still there when someone does the correlation capture.
2. **Keep it unexported** until `0xa9` is confirmed on an AE103, same bar the decode tables are held to (`index.ts` keeps `SOLIX_SOLARBANK_FIELD_NAMES` module-internal for less than this).

Either way the old numbering is worth keeping recorded somewhere rather than deleted — it's the only AE103-side datum here, and it's what a correlation capture would be checked against.

---

src/transport/mqtt/index.ts
```diff
@@ -9,3 +9,6 @@ export * from "./biz-stream.js";
 // SOLIX_METER_FIELD_NAMES stay module-internal; their specs import them from the module directly.
 export { SolixMqtt } from "./solix-mqtt.js";
 export type { SolixMqttOptions, SolixMqttDevice, SolixReading, SolixParamFrame, SolixChannel } from "./solix-mqtt.js";
+// The operating-mode enumeration IS consumer-facing (a caller labels the `mode` value with it), so unlike
+// the decode tables it is exported.
+export { SOLIX_EMS_MODES } from "./solix-mqtt.js";
```
martijnpoppen: This export is what turns the numbering question above from an internal note into a support commitment — and it's also what breaks the docs build.

`typedoc` reports (reproduced locally on `6ea6ba1`, exit 4):

```
[warning] The comment for SOLIX_EMS_MODES links to "SOLIX_STATE_FIELD_NAMES" which was
resolved but is not included in the documentation.
[warning] Found 0 errors and 1 warnings
```

`SOLIX_EMS_MODES` is now exported from the package entry, so typedoc documents it and follows its `{@link SOLIX_STATE_FIELD_NAMES}` — but that symbol is deliberately module-internal (the comment three lines up says so). A `{@link}` from a documented symbol to an undocumented one is the warning, and the docs workflow treats it as failure.

Worth flagging why `npm run verify` said green: `verify` does not run the docs build. The failing `build` job is `cd docs && npm run build` → `typedoc && guard:docs && vitepress build`. So a broken `{@link}` in shipped `src/` JSDoc is invisible to the local gate and only shows up in CI. (Same trap as #196.)

The fix follows whichever way you go above: if the map stays unexported, the link resolves the same way it does today for the other tables; if it stays exported, the link needs to point at something that's also exported — or be plain backticks rather than `{@link}`.

---

martijnpoppen (commented):
Reviewed `6ea6ba1` (2 commits) against `beta-0.2.0`.

The cross-reference commit (`463c1f5`) is the good half and I'd take it as-is — getting the vendor's own register names next to each reverse-engineered name is exactly the kind of evidence this file has been short of, and it's annotation-only. The second commit (`6ea6ba1`) is where the problems are, and they're separable.

**CI is red**, and it's this PR: the `build` job fails at `typedoc` with exit 4 over the new `{@link}` in `SOLIX_EMS_MODES`. Reproduced locally. Details inline on `index.ts`; the short version is that `verify` doesn't run the docs build, so "`npm run verify` green" and "CI green" aren't the same claim here.

**The mode enumeration contradicts the AE103 numbering this repo already recorded, on every value** — and it's exported, so a consumer labelling an ff09 `mode` with it gets a wrong label rather than no label. Table and two suggested ways out are inline on the declaration. This is the one I'd want settled before merge; the rest is minor.

Two smaller things:

**The body undersells the diff.** It says "comment-only, **non-breaking** — no rename", which is true of `463c1f5` but not of `6ea6ba1`: that commit adds a new public export to the package entry. Anyone reading the description and skimming the diff would approve a new API surface thinking they were approving comments. Worth a line in the body, or splitting the enum into its own PR — which would also let the cross-reference land today instead of waiting on the numbering question.

**The PR's own fixture now sits in a hole the new map defines.** `463c1f5` rewrites the spec's `mode = 2` comment to "label it with `SOLIX_EMS_MODES`" — but `SOLIX_EMS_MODES[2]` is `undefined` by design (the JSDoc says value 2 is unassigned). Following the comment's own instruction on the frame directly below it produces nothing. Either the fixture should use a value the map assigns, or it's a small piece of evidence that AE103 really does number these differently.

---

**One thing the PR gives away too cheaply.** The JSDoc now names `reactive_power` and `power_factor` as the leading candidates for `channel_b2` / `channel_b7`, jointly. The committed `LOAD_FRAME_HEX` already discriminates them for `b2` — no new capture needed:

```
a8 meterPowerL1   =   2.7000 W
ac meterVoltageL1 = 239.3000 V
af meterCurrentL1 =   1.3710 A
b2 (unnamed)      =   0.009000

apparent power S = V·I = 328.08 VA
power factor P/S       = 0.00823      <-- b2 reads 0.009
reactive power Q       = 328.07 var   <-- four orders of magnitude off b2
```

So for `b2`, **reactive power is ruled out** (it would read ~328, not 0.009) and a dimensionless 0…1 quantity is the only one of the two that fits — `b2` lands within ~9% of the frame's own computed power factor, on a value quantised to three decimals against inputs given to 2–4 significant figures. That's not enough to *name* it `powerFactor`, and I wouldn't; but "b2 is dimensionally consistent with power factor and inconsistent with reactive power, per the load frame's own P/V/I" is a stronger and fully-grounded statement than listing both candidates for both tags. `b7` (0.1) isn't explained by either and stays open.

Gate on this branch: `tsc --noEmit` clean, `prettier --check` clean, all 5 guards pass, 3674 tests pass. Only the docs build fails.

Review and comments only — not pushing anything to this branch.

---

src/transport/mqtt/solix-mqtt.ts
max246: Fixed in 7622a9a. You're right on every value — the AE103 ff09 numbering (1 custom / 2 self-consumption / 4 rapid charge / 7 smart / 8 dynamic tariff) is not the Modbus map. I've restored that AE103 numbering directly on the `0xa9` field comment, and the Modbus enum's JSDoc now states plainly it is **not** a decoder for ff09 `mode` — labelling an ff09 value with it would be confidently wrong. The two are only folded if a live `0xa9` correlation capture ever proves the numbers match.

src/transport/mqtt/index.ts
max246: Fixed in 7622a9a. Renamed to `SOLIX_MODBUS_EMS_MODES` so the public name says which hardware rev it's from (a caller can't mistake it for the ff09 decoder), and dropped the internal `{@link SOLIX_STATE_FIELD_NAMES}` for plain backticks — that link from the exported symbol to an undocumented one was the typedoc exit 4. `npm run verify` doesn't run the docs build so it was invisible to it; confirmed green now with `cd docs && npm run build` (typedoc + guard:docs + vitepress all pass).

martijnpoppen: Verified on `7622a9a` — docs build is fixed, and CI agrees (both `build` jobs green).

Reproduced locally with the same chain CI runs:

```
typedoc      exit 0   (was exit 4 on 6ea6ba1)
guard:docs   exit 0   docs guard OK — no method references or internal paths in api/
```

Plus the SDK gate on this head: `tsc --noEmit` clean, `prettier --check .` clean, all 5 guards pass, 202 files / 3674 tests pass.

The rename is the right shape — `SOLIX_MODBUS_EMS_MODES` carries the hardware rev in the name, so the mistake it guards against isn't available to make by accident, and the comment here now says outright that it isn't the ff09 decoder. Dropping the `{@link}` for backticks is also the better of the two fixes, since it keeps the map exported without dragging an internal symbol into the generated docs.

---

src/transport/mqtt/solix-mqtt.ts
martijnpoppen: Verified on `7622a9a`. This resolves it, and the JSDoc is now stronger than what it replaced — "labelling an ff09 `mode` value with this Modbus map would be confidently wrong" states the trap directly, and keeping the map as "the thing an AE103 `0xa9` correlation capture would be checked against" gives it a job that doesn't require it to be right about ff09. The spec fixture comment tracks it too (`mode raw value 2 = self-consumption (AE103 numbering)`).

You also took the `b2` point further than I'd expected — the writeup is correct and I re-derived it: 239.3 V × 1.371 A = 328.08 VA, P/S = 0.00823 against `b2` = 0.009, and reactive power would sit at ~328 var. Leaving it raw despite the magnitude tell is the right call on one frame.

One loose end, not a blocker and not something this PR introduced. The whole fix now rests on the AE103 numbering being right, and the JSDoc attributes it as "correlated against the app" — but the commit that first recorded it (`60874a8`) doesn't say where it came from, and there's nothing in the repo to point at. `AGENTS.md` asks wire claims to be grounded in evidence you can point at, and this one has become load-bearing in a way it wasn't when it was a passing parenthetical.

If it *was* an app correlation you did, a line saying how (set each mode in the app, read `0xa9`) would close it permanently — the value of that is mostly that it stops the next reader from re-litigating exactly what I did here. If it was inference rather than a capture, then the honest state is two competing unevidenced tables, which is still strictly better than before (an unevidenced table is not an exported *decoder*), but worth the JSDoc saying so rather than asserting correlation.

Either way the PR is green and I'd merge it as-is; the provenance line can ride any later commit.

---

martijnpoppen (approved):
Approving `7622a9a`.

Every blocker I raised is resolved, and I re-verified rather than going on the replies. Note the base moved since the last CI run (`beta-0.2.0` is now `9dc7e92` — #207, #208, #210 landed), so I test-merged this branch against **that** base and ran the gate on the merged result, not just the branch in isolation:

```
merge against 9dc7e92   CLEAN
tsc --noEmit            clean
prettier --check .      clean
5 guards                all pass
vitest                  202 files / 3707 tests pass
typedoc                 exit 0
guard:docs              exit 0
```

Plus GitHub CI green on both `build` jobs, and `mergeable_state: clean`.

What changed since my first review:

- `SOLIX_EMS_MODES` → `SOLIX_MODBUS_EMS_MODES`, so the public name carries the hardware rev and the ff09 mistake isn't available to make by accident.
- The `{@link SOLIX_STATE_FIELD_NAMES}` → plain backticks, which was the actual typedoc exit 4.
- The AE103 ff09 numbering restored on the `0xa9` field comment, with the Modbus map's JSDoc stating outright that it is not the ff09 decoder.
- The spec's `mode = 2` fixture comment no longer points at a map that leaves 2 unassigned.
- The body now documents the new public export and the `b2`/`b7` analysis, which closes the "body undersells the diff" point too.

**What I did and didn't verify.** Everything above is machine-checked. I can't independently confirm the vendor-side facts — that Anker's published integration really names those registers, or that the AE103 ff09 numbering is right — and I'm not vouching for them. What makes that acceptable to approve is that the PR no longer *asserts* them in a way that can mislead code: the Modbus table is labelled as the Modbus rev's and disclaimed as a decoder, and the cross-reference annotations are comments beside names that were already there. The one place the code would have acted on an unverified claim was the exported enum, and that's exactly what got fixed.

The provenance question on thread [#discussion_r4054383451](https://github.com/mega-yfue/eufy-sdk/pull/206#discussion_r4054383451) stays open as a follow-up — the JSDoc attributes the AE103 numbering as "correlated against the app" while `60874a8` doesn't record where it came from. Not a merge blocker, as I said there; worth a line on any later commit.

For the record: this approval was posted by Claude Code on @martijnpoppen's instruction, not a separate human read. @lenoxys is still a requested reviewer.

---

PR #207, opened 2026-09-19
feat(model): describe a detection only where the device evidences it

martijnpoppen:
## What changed

`EventMapping` gained an optional `claim` — evidence resolved per device when the manifest is built, so `CapabilityDescriptor.events` stops being the capability module's static list and becomes what *this unit* is expected to emit. `emitsOf(m)` → `emitsOf(m, reads, ctx)`.

Three fields, AND rather than `DetectionSpec`'s OR, because a claim rules a family **out** rather than finding one more reason to say yes:

| Field | Evidence | Claimed by |
| --- | --- | --- |
| `codecs` | the families whose devices issue the id | all five AI rows (`camera`) |
| `reads` | a member whose getter installed — the device reported the parameter | `vehicleDetected` → `aiDetectType` |
| `homeBaseAttached` | the topology an id belongs to | `dogDetected` → a station parent |

`codecs` aside, a fact the context does not carry rules nothing out. `codecs` is the exception for the reason `AvailabilityContext.codec` already gives: absent means a device outside the eufy device model, not one whose family is merely unresolved, so it matches no vocabulary and cannot be issuing that vocabulary's ids.

**This narrows the description only.** The dispatch index is untouched — a device that sends the push still gets the event. Not a judgement call: `decodeEvent` already refuses to gate a single-claimant push on capabilities, *"detection is evidence-based and can under-report, so gating every push on it would silently drop real events."*

Two smaller things ride along:

- `homeBaseAttached` moves from `CommandContext` to `AvailabilityContext`, where its own JSDoc already said it belonged — "a DEVICE fact, not a transport one" — and is as true of a described device as of a commanded one. `Device.describe` states it from `stationSn`. Non-breaking: `CommandContext` still has it by inheritance, and `siren`/`rtsp` read it unchanged.
- `PET_DETECTION` (3106) moves from `doorbell` to `motion`.

## Why

Every AI detection is emitted by `motion` or `person_detection`, and both bind on the `camera` codec alone. So binding `motion` at all put `vehicleDetected` and `dogDetected` in the manifest — the unit was never asked. **An evidence gate on a capability is not an evidence gate on its events**, and detections are where that bites: one module carries a dozen independent features and a camera has some of them. A caller offering a control per described event offers vehicle and dog classification on hardware that does neither.

Pet was the inverse gap. 3106 was mapped only in `doorbell`, whose detection spec is `{ modelHints: [/doorbell/i] }` — so a camera that is not a doorbell decoded **nothing** for it, and the push reached no caller at all.

Moved rather than added: a doorbell holds both capabilities, so a contested id would match twice and emit one push as two events. `decode-event.spec.ts` holds that, and the existing "contested ids" spec still expects an empty set.

Also fixed in passing: a standalone motion sensor binds `motion` — it *is* motion detection by device type — and was described as emitting crying, sound, pet, vehicle and dog. It announces itself under `CusPushEvent.MOTION_SENSOR_PIR` and reports no AI classification of any kind. It now describes exactly that one event.

## Evidence

**The push id cannot be the signal.** The AI ids are declared verbatim across the families:

| Id | `DoorbellPushEvent` | `IndoorPushEvent` | `HB3PairedDevicePushEvent` |
| --- | --- | --- | --- |
| 3106 pet | ✓ | ✓ | ✓ |
| 3107 vehicle | ✓ | ✓ | ✓ |
| 3108–3110 dog | — | — | ✓ |

The integer is the vendor's meaning for a classification, never a statement that a unit performs it. `IndoorPushEvent` listing 3107 does not make an indoor camera a vehicle classifier. Only the dog ids are genuinely family-exclusive — which is exactly what `DOG_CLAIM` rests on, and why it gates on topology rather than on a parameter.

**Vehicle rests on 1298.** `MOTION_MEMBERS.aiDetectType`, provenance `verified`, bits decoded live and confirmed against the app: `0x30000` enabled base, bit0 human recognition, bit1 human detection, **bit2 vehicle**, bit3 pet. It is the one signal a unit reports about itself that names vehicles at all, and claiming on it is the same bar the typed read of that bitmask already answers to. A camera that starts reporting the parameter starts announcing the event with it.

**Dog gates on any station, not on an HB3 specifically.** A camera on an older base is not in the population that issues 3108–3110 either, but nothing here establishes the generation split, so the coarser test is the one that cannot lose a real detection.

**No parameter was invented.** 6045 `indoorSetMotionDetectionType` is the obvious companion to 1298 and was deliberately **not** modelled: its name is the app's own constant (provenance `apk`) and its presence in `SECURITY_PARAMS` means a real device reported it, but nothing establishes what its *value* means. The gate does not need it — absence of 1298 is the evidence.

## Verified how

- [x] `npm run verify` is green locally — format, typecheck, all five guards, build, ESM check, 91 doc snippets, examples, **3679 tests**
- [ ] Ran against a real device — **no.** See below.
- [x] Added or updated specs (offline, synthetic fixtures) — claim resolution in `manifest.spec.ts`, pet decode in `decode-event.spec.ts`

  Synthetic records through `Device.fromRecord` + `bindActions`, reading `describe().details[].events`:

  | Device | Described detections |
  | --- | --- |
  | standalone camera, no 1298 | motion, pet, sound, crying |
  | standalone camera, reports 1298 | + vehicle |
  | station-attached camera, reports 1298 | + vehicle, dog |
  | station-attached camera, no 1298 | + dog |
  | standalone motion sensor | motion |

  And `decodeEvent` on push 3106 answers exactly one `petDetection` for a camera, for a doorbell, and for a device holding both capabilities.
- [x] Docs updated, or n/a — n/a; the surface is documented in the JSDoc the site generates from.

## Unverified or deferred

**Whether an indoor camera reports 1298 is not established here.** The vehicle claim turns on it and no indoor unit was available to ask. If one does not report it, its vehicle event is correctly withdrawn. If one does, the claim passes and the description is unchanged from today — the gate never over-claims, but it may under-deliver on this model, and that is the case to check on hardware first.

The honest next step if so is 6045: establish what its value means on a real unit, then claim `vehicleDetected` on the absence of a vehicle bit in whichever selector the unit answers on. That is a wire question, not a refactor, so it is deliberately not in this PR.

No write path changed, so nothing here ships a frame that was not already confirmed.

https://claude.ai/code/session_01Y2Xprsr6qytfQb6bhR8Lpw

---
_Generated by [Claude Code](https://claude.ai/code/session_01Y2Xprsr6qytfQb6bhR8Lpw)_


max246 (approved):
Reviewed against AGENTS.md — rigorous and correct. Verified:

- **The core safety property holds:** `claim` narrows the manifest DESCRIPTION only; the dispatch index stays unclaimed, so a device that sends the push still gets the event. Under-describing is recoverable, dropping a real detection isn't — and the existing `decode-event` "refuses to gate a single-claimant push on capabilities" invariant is untouched.
- **`holds()` is right:** AND semantics (a claim rules a family OUT), codec-absent rules out (the one documented exception, because an absent codec means outside the eufy model), a missing required read rules out, and `homeBaseAttached` is ignored when the context lacks it — so a fact the context doesn't carry withdraws nothing.
- **The claims are evidence-grounded:** vehicle gates on `aiDetectType` (1298, the bitmask that actually has a vehicle bit — same bar as its typed read), dog on station attachment (3108–3110 are HB3-paired-exclusive; the coarser topology test can't lose a real detection), all AI on the `camera` codec. `PET_DETECTION` moved (not added) to `motion` to avoid a doorbell double-matching one push into two events. "No parameter invented" (the 6045 non-modelling call) is the honest choice.
- **Non-breaking `homeBaseAttached` move** to `AvailabilityContext` — `CommandContext` keeps it by inheritance, `siren`/`rtsp` read it unchanged, and the JSDoc already said it belonged there.
- **Tests nail it,** including the subtle one I most wanted to see: `aiDetectType` reported → `vehicleDetected` described / absent → not; dog present/absent/coarse; and the standalone sensor now describing exactly `["motion"]` instead of crying/sound/pet/vehicle/dog. `npm run verify` green (3679 specs).

Nothing to address in the diff. Approving.

PR #208, opened 2026-09-19
fix(http): put the token-replacement rate on the rejection, so a host cannot loop past it

martijnpoppen:
## What changed

`SessionExpiredError` now carries the rate the transport has already worked out:

- **`retryAfterMs`** — the remainder of the hold-off on replacing the token, from the same arithmetic `recoveryDue()` gates on. One function (`holdOffRemainingMs()`) now serves both, so the number a caller is handed and the number this client observes cannot disagree.
- **`contended`** — true when the rejection landed *inside* that wait, which is the shape repeated displacement has.
- A `login()` that follows a rejection counts against the hold-off exactly as `recoverRejectedSession()` does (`noteTokenReplacement()`), so the wait escalates whoever spent the session.
- The message's hint leads with the mechanism — something else signed in on this account, and the cloud holds about one session per account and device identity — and offers the `openudid` remedy for the case it applies to.
- `EufyMega`'s `sessionExpired` event is typed as `SessionExpiredError` rather than `Error`, so a host can read the two fields.

## Why

The hold-off exists because repeated logins on one account are what make the cloud start demanding captchas: first replacement immediate, then a minute, doubling to half an hour. It bounded only the replacements this client spent on itself.

A host holds the other half. It gets a `SessionExpiredError`, finds no number in it, and re-drives `login()` on whatever ladder its login state machine has for a *failed* login — seconds. So a displaced session became a login every few seconds from outside a hold-off built to prevent exactly that, and the `holding off 60s` warning named a wait nothing was observing. Reported from a Homey install as a stream of

```
/app/house/get_house_list failed (401/401): token error — another client may be signed in with …
```

where the host was already giving itself its own `openudid` — so the only remedy the message named was one it had applied, and the part that was actionable (something else is signing in on this account) was not in a form it could act on.

## Evidence

Wire behaviour unchanged — no new codes, headers or endpoints. The 401 wording and code classification (`tokenRejected`) are untouched; this changes only what the client does about a rejection it already recognises, and what it reports about it.

## Verified how

- [x] `npm run verify` is green locally (format, typecheck, all five guards, build, ESM, doc snippets, examples, 3681 specs)
- [ ] Ran against a real device — n/a, no wire path changes
- [x] Added or updated specs (offline, synthetic fixtures) — four in `session-expiry.spec.ts`: the wait and `contended` on a held-off rejection, the wait after a replacement that was spent and failed, no wait when none has been spent, and a host-driven `login()` escalating the hold-off from 60s to 120s. The last one was checked against the unfixed client: it fails there with the wait stuck at 60s.
- [x] Docs updated — `getting-started.md` (the contention bullet), `events.md` and `troubleshooting.md` (the `sessionExpired` snippet now honours `retryAfterMs`, plus a row for the symptom)

## Unverified or deferred

**Who displaces the session is not settled by this PR.** The hint's premise — another client on the same account — is the likeliest reading and this client's own, but the report does not prove it, and a replacement token rejected for a reason of its own looks identical from here (the `gtoken` mismatch in `homey-eufy-app`'s `login-gtoken-401.md` did exactly that, producing this same hold-off warning while having nothing to do with contention). What this PR fixes is the loop and the message; the cause of the underlying 401 needs a debug log. The consumer side is mega-yfue/homey-eufy-app#24, which records both candidates and what would tell them apart.

Also worth a reviewer's eye: `SESSION_KICKED`'s own note says the cloud keeps *about one session per account*. If that is per account rather than per account-and-device, the `openudid` half of the hint isolates nothing — the wording now states the mechanism as "about one session per account and device identity", which is the same hedge the code comment already carried.

https://claude.ai/code/session_01VRXN5fHarHZE3fKUDw6JWs


max246 (approved):
Solid fix, and well-reasoned. Verified:

- **Correct root cause + fix.** The hold-off only ever bounded the client's own replacements; a host that re-drove `login()` on a bare `SessionExpiredError` looped past it every few seconds. Surfacing `retryAfterMs`/`contended` and counting a host-driven login against the same hold-off (`noteTokenReplacement()` on a `login()` that follows a rejection) closes exactly that loop.
- **DRY is the right call.** `holdOffRemainingMs()` as the single source for both `recoveryDue()` and `retryAfterMs` is what guarantees the client and the host can't disagree about the rate — the stated invariant holds.
- **Single-count is sound.** I traced it: the recovery path calls `noteTokenReplacement()` (clearing `rejectedTokenPending`) *before* its own login, so a replacement is counted once whether the client or the host spent it; `noteSessionWorking()` also clears the flag so a later unrelated login can't wrongly count a long-since-resolved rejection.
- **Type change is safe.** Narrowing `sessionExpired` to `SessionExpiredError` (an `Error` subclass) is additive for existing handlers and lets a host read the two fields.
- **JSDoc is ground-truth** and honestly hedged — the "who displaces the session" uncertainty and the per-account-vs-per-account-and-device question are called out in the doc and the PR rather than asserted. Consumer-agnostic scan of the diff (source + the three docs) is clean.
- Tests cover the four cases including the adversarial "host-driven login escalates 60s→120s, checked against the unfixed client." CI green.

Approving. One non-blocking nit (not in the diff, so no inline): the PR **body** names the specific consumer ("a Homey install" + the `homey-eufy-app#24` link). We genericised exactly this on #197 per the consumer-agnostic rule (which covers PR bodies too) — worth trimming here for consistency, but it doesn't touch the shipped artifacts, so it's not a blocker.

PR #209, opened 2026-09-19
feat(solix): expose Solarbank batteryHealth (SOH) from the a4 BMS blob

max246:
The `0xa4` BMS status blob already yields **temperature** (byte `length-8`) and a self-validated **SOC** (byte `length-6`). The byte one after SOC (`length-5`) reads **100** on a captured pack — a state-of-health percentage, matching the app's own `bmsHealth` field. Emit it as **`batteryHealth`**.

- **Self-validated:** only read when the blob's SOC byte equals tag `0xa3` (same gate as temperature), so it's withheld on the empty realtime-frame blob variant rather than read from the wrong offset.
- **CANDIDATE:** the offset is confirmed and the value fits SOH, but that it's SOH *specifically* isn't hardware-correlated yet — clearly noted in the JSDoc. Read-only.

Found by static analysis of the app decompile (the `bmsHealth*` libapp vocabulary + our own `SOLIX_BATTERY_FIELDS_LIVE` blob notes). `npm run verify` green.


src/transport/mqtt/solix-mqtt.ts
```diff
@@ -129,6 +129,9 @@ export const SOLIX_SOLARBANK_FIELD_NAMES: Readonly<Record<number, string>> = {
   0xc7: "pv2Power",
   0xc8: "pv3Power",
   0xc9: "pv4Power",
+  // CANDIDATE (not yet live-correlated): 0xb1 reads 0 at idle and ~5.7 while the pack cycles
+  // (~5.7 A × ~50 V ≈ 280 W), which fits the app's `batteryCurrent` magnitude. Named tentatively.
+  0xb1: "batteryCurrent",
```
martijnpoppen: The sanity check in this comment doesn't hold against the fixture the same commit adds it to.

The comment reasons `~5.7 A × ~50 V ≈ 280 W`. But `0xb1 = f32(5.7)` goes into `FRAME`, and that frame already asserts:

```ts
expect(v.batteryPower).toBeCloseTo(510, 0);
expect(v.chargePower).toBeCloseTo(510, 0);
```

510 W ÷ 5.7 A = **89.5 V**, not ~50 V. So the one frame in the repo carrying this tag describes a pack that can't satisfy the comment's own arithmetic. Either the fixture's `5.7` should be the current that matches 510 W (~10.2 A at 50 V), or the pack voltage assumption is off, or `b1` isn't battery current. As it stands the comment and the fixture contradict each other, and whichever is wrong, a reader checking the reasoning hits the discrepancy immediately.

Second thing, and the bigger one: "reads 0 at idle and ~5.7 while the pack cycles" is an on-wire observation, but there's no capture in the repo that shows it. Every Solarbank frame in this spec is built by `buildFrame` — the serials are literally `AE1BMSYNTH00001` / `AK7DSYNTH00000001`. The only real captures committed are the two AE1X0 **meter** frames (`FRAME_HEX`, `LOAD_FRAME_HEX`). So `expect(v.batteryCurrent).toBeCloseTo(5.7, 1)` passes because the test wrote 5.7 into the frame two lines earlier — it exercises the tag→name map, which is worth testing, but it can't corroborate the naming.

This is the same shape as the `0xb2`/`0xb7` evidence question on #184, which got resolved by committing a real second capture with the load actually varying. An idle/load AE103 pair would do the same job here and would settle `b1`, `bf` and the SOH byte in one go.

---

src/transport/mqtt/solix-mqtt.ts
```diff
@@ -271,6 +280,7 @@ function addSolarbankScalars(frame: SolixParamFrame, out: Record<string, number>
     const body = frame.fields.get(0xa4)?.subarray(1);
     if (body && body.length >= 8 && body[body.length - 6] === soc) {
       out.batteryTemperature = body[body.length - 8]!;
+      out.batteryHealth = body[body.length - 5]!; // SOH % — candidate; byte after the validated SOC
```
martijnpoppen: Inline body comment — `CODING_STANDARDS.md:164`:

> **Comments: JSDoc above the declaration, ground-truth, no inline body comments.** Put the rationale in a JSDoc block **above** the function […] do not narrate inside the body with `//` lines.

Everything this trailing comment says is already in the function's JSDoc (the "CANDIDATE", the `%`, and "the byte one after it"), so by the rule's own test — "a body comment is a smell that the JSDoc is incomplete — move it up, or delete it if the code already says it" — this one can just go. The line above it (`out.batteryTemperature`) carries no trailing comment for the same reason.

---

martijnpoppen (commented):
Reviewed `2d2f731` (2 commits) against `beta-0.2.0`. CI is green and I reproduced it: `tsc --noEmit` clean, `prettier --check` clean, all 5 guards pass, 3673 tests pass, and `cd docs && npx typedoc` exits 0 (worth checking explicitly on this file — it's outside `npm run verify` and it's what's failing #206 right now).

**The `batteryHealth` half is well built.** Reusing the existing `body[length-6] === soc` gate rather than adding a second one is the right call: the empty realtime-frame `0xa4` variant withholds both values together, so there's no path where health is read from an offset temperature wouldn't trust. The offset is internally consistent with the documented struct (`[… TEMP 01 SOC SOH 00 01 00 02]` → `length-8`, `length-6`, `length-5`), CANDIDATE is stated plainly in the JSDoc, and it's read-only. No objection to the mechanism.

**One thing about the evidence, though, that CANDIDATE doesn't quite cover.** The PR rests on the byte reading `100` on a captured pack. `100` is the single least discriminating value that byte could have returned: a healthy pack's SOH is 100, and so is a constant, a padding byte, a percent-scale ceiling, or a `0x64` that means something else entirely. Every hypothesis predicts 100 here, so the observation doesn't favour SOH over any of them — it's consistent with the claim without being evidence for it. What would discriminate: a second pack with any wear on it, or the same pack read across a long enough span to move. Until then "the offset is confirmed" is fair, but "fits SOH" is doing less work than it reads like. Worth saying so in the JSDoc in those terms — the current wording invites the next reader to treat it as corroborated.

**The description covers one of the three fields this PR adds.** Title and body are entirely about `batteryHealth`; `2d2f731` also names `0xb1 → batteryCurrent` and `0xbf → faultStatus`. Those are two further unconfirmed tag→name bindings, and they're the weaker two — a reviewer going by the description would approve them without knowing they were there. They'd also be fine as their own PR: `batteryHealth` is self-validated and ready, the other two are waiting on the same capture.

Details on `b1` inline — short version, its comment's own arithmetic (`~5.7 A × ~50 V ≈ 280 W`) contradicts the fixture it was added to (which asserts 510 W at that current ⇒ 89.5 V), and the "0 at idle / ~5.7 under load" observation isn't in the repo: every Solarbank frame here is synthesised, so the new assertion passes on bytes the test itself wrote.

**On `faultStatus` (`0xbf`)** — the app debug string `updateDeviceStateInfo----faultStatus` establishes that the app *has* a field by that name, which is good; it doesn't establish that `0xbf` is the tag carrying it. That step is still inference, and the comment reads as though the debug line settled both. A fixture where the bitfield is anything other than `0` would help a lot — right now `expect(v.faultStatus).toBe(0)` is indistinguishable from the tag being absent-and-defaulted, and `0` is what you'd see from almost any unrelated status tag on a healthy unit. If a fault state is reproducible on the hardware at all (pull a PV string, over-temp it), one frame in that state pins the tag and gives the bitfield a known bit.

One nit inline on the trailing `//` at line 283 (`CODING_STANDARDS.md:164`).

Nothing here is a correctness bug — the decode is sound and the gating is right. It's about how much the names claim relative to what's been shown, which given `SOLIX_SOLARBANK_FIELD_NAMES` is the file's whole point.

Review and comments only — not pushing anything to this branch.

---

src/transport/mqtt/solix-mqtt.ts
max246: Both fair — fixed in 83e8c6c by dropping `batteryCurrent` (0xb1) **and** `faultStatus` (0xbf) entirely. You're right on both counts: the sanity math contradicted the fixture (510 W ÷ 5.7 A = 89.5 V, not ~50 V), and more importantly a synthetic `buildFrame` can't corroborate a naming — the assertion just re-reads bytes the test wrote two lines up. Naming a float/state slot from position alone with no real capture is the 'never ship guessed params' rule, so they come out until a real idle/load AE103 pair pins b1, bf (and confirms SOH) in one capture, exactly as #184 did.

I kept **only** `batteryHealth` (SOH), because it isn't the same shape of guess: the byte is read at a fixed offset immediately after the SOC byte, and that SOC byte self-validates against tag `0xa3` — the parse is withheld unless the blob is the right variant. So its OFFSET is structurally anchored (not a bare slot), even though 'it is SOH specifically' stays CANDIDATE in the JSDoc pending the same capture. If you'd rather SOH also wait for the capture, say so and I'll pull it too.

src/transport/mqtt/solix-mqtt.ts
max246: Done in 83e8c6c — deleted the trailing body comment. As you said, the JSDoc above already carries the CANDIDATE / % / 'byte after SOC' points, so per CODING_STANDARDS.md:164 the inline line just goes.

src/transport/mqtt/solix-mqtt.ts
martijnpoppen: **No — keep SOH.** Your distinction is the right one, and I'd have drawn the line in the same place.

The difference is real, not a judgement call. `b1` and `bf` were *positional* claims: nothing but the slot's position linked the tag to a quantity, so a wrong guess decodes silently and forever. `batteryHealth`'s offset is anchored to a byte the frame itself validates — the parse only runs when `body[length-6] === soc`, so the blob has to agree about where SOC is before SOH is read relative to it. A wrong-variant blob is rejected rather than misread, which is exactly the failure mode the other two had no defence against.

Three more reasons it clears the bar the other two didn't:

- **The rule you invoked is about the typed surface.** "Never ship a guessed parameter as a typed getter" — `solixReadings` answers `Record<string, number>`, and `SolixDevice` keeps its own private `values` bag behind `telemetry()`. No typed getter appears and nothing in `model/capabilities/` gains a property, so there's no compile-time claim being made about this byte.
- **`batteryTemperature` is the precedent, one line up.** Same blob, same gate, same class of evidence, in the tree since `60874a8`. Pulling SOH would hold it to a standard its immediate neighbour doesn't meet.
- **Read-only**, so the unverified-write rule never engages.

What's left is purely semantic — is this byte SOH, or something else that happens to read 100 — and `CANDIDATE` in the JSDoc is the instrument the repo already has for that. My original point still stands and the JSDoc handles it correctly by saying the value *fits* SOH rather than confirms it. Worth knowing for whenever the capture happens: the discriminating observation is a pack **not** at 100. A second device at 100 adds nothing; one at 97 settles it outright.

Verified `83e8c6c`: the removals are clean, no stale references to either name anywhere in `src/`, and the trailing `//` on the `batteryHealth` line went with them. CI green (both `build` jobs), and the full gate reproduced locally — `tsc` clean, `prettier --check .` clean, all 5 guards, 202 files / 3673 tests, `typedoc` exit 0, `guard:docs` exit 0.

Title and body now describe the whole diff too, which was the other half of that finding. Nothing outstanding from me on this PR.

---

martijnpoppen (approved):
Approving `83e8c6c`.

Re-verified against the base as it stands now — `beta-0.2.0` moved to `9dc7e92` (#207, #208, #210) after this branch's CI ran, so I test-merged and gated the merged result rather than the branch alone:

```
merge against 9dc7e92   CLEAN
tsc --noEmit            clean
prettier --check .      clean
5 guards                all pass
vitest                  202 files / 3706 tests pass
typedoc                 exit 0
guard:docs              exit 0
```

GitHub CI green on both `build` jobs, `mergeable_state: clean`.

`83e8c6c` dropped `batteryCurrent` (`0xb1`) and `faultStatus` (`0xbf`) — the two positional guesses — leaving only `batteryHealth`. I checked the removals are complete: no references to either name survive anywhere in `src/`, the fixture tags are gone with them, and the trailing body comment went too. Title and body now describe the whole diff.

What remains is a single read-only byte whose offset is anchored to a self-validating SOC byte, labelled `CANDIDATE`, creating no typed getter — the reasoning is on [#discussion_r4054393036](https://github.com/mega-yfue/eufy-sdk/pull/209#discussion_r4054393036) and I won't repeat it. Dropping the other two rather than defending them was the right call and made this an easy approval.

**What I did and didn't verify.** The decode logic, the gating, the removals and the whole gate are machine-checked. I have no way to confirm the hardware-side claim — that the byte reads 100 on a real pack, or that it is SOH — and I'm not vouching for it. The JSDoc says the value *fits* SOH rather than confirms it, which is the honest framing, and the discriminating observation whenever a capture happens is a pack **not** at 100.

For the record: this approval was posted by Claude Code on @martijnpoppen's instruction, not a separate human read. @lenoxys is still a requested reviewer.

---

PR #210, opened 2026-09-19
feat(vacuum): let a caller state the cleaning type, extent and mop level

martijnpoppen:
## What changed

`vacuumClean().setCleanParam(cleanType, cleanExtent, mopLevel)` — the write side of DP 154.

- `encodeCleanParam` builds `CleanParamRequest.clean_param` from the same `CLEAN_PARAM_FIELD` table the reads decode through, via `RawDpWriter`.
- `cleanParamWireValue` inverts the existing `CLEAN_TYPE` / `CLEAN_EXTENT` / `MOP_LEVEL` maps, so a write and the read that observes it cannot come to disagree about which integer a name means.
- The member publishes `args` with all three enum domains, so a caller can render three pickers without inventing a value set.
- `docs/vacuums.md` gains a "What a run does with a surface" section and the Controls preamble now names the third evidence tier this write sits in.

`clean_times`(7) is never written, so the robot keeps its configured pass count. `fan`(6) is not written either — suction has its own data point and `setSuctionLevel` already owns it.

One call takes all three settings rather than three setters taking one each: the device's `CleanParam` carries all three, so a per-setting write would be the same message with the others silent, and what a robot makes of a half-stated one has not been observed. This is the same reasoning the existing spec states as "grows no setters — the whole message would have to be re-encoded to write one field".

## Why

`cleanType`, `cleanExtent` and `mopLevel` were readable and not settable, so a caller could show what a robot is set to and not change it. A vacuum's cleaning mode is one of the things a host is most often asked to automate.

## Evidence

The frame is field 1 of `CleanParamRequest`, carrying the very `CleanParam` this module decodes out of field 1 of the reports a live T2351 sends. Everything the encoder needs is already read off that capture and shipped:

- `CLEAN_PARAM_FIELD` — `clean_param`(1), `clean_type`(1), `clean_extent`(3), `mop_mode`(4), `level`(1), `value`(1).
- the single-field wrapper shape (`CleanType{value}`, `CleanExtent{value}`), live-proven for `clean_type`.
- `MOP_LEVELS` — `provenance: "verified"`, captured setting the water level to High and reading back `mop_mode { level: 2 }`.
- the proto3 default rule already relied on by `encodeModeCtrl` and by `decodeCleanParamValue`'s "a present-but-empty wrapper answers 0".

So `encodeCleanParam` writes what `decodeCleanParamValue` reads, and the specs hold the two to each other: byte-exact expectations built from the same `int`/`sub` fixture helpers (which encode the proto3 rules independently), plus round-trips back through `decodeCleanType` and `decodeCleanParamValue`.

## Verified how

- [x] `npm run verify` is green locally (3696 tests, 202 files)
- [ ] Ran against a real device — **not done**, see below
- [x] Added or updated specs (offline, synthetic fixtures) — `encodeCleanParam` byte-exactness, the zero-member wrapper, `clean_times` absence, round-trip through the reads, DP-154 dispatch on the AIoT line and absence on the Tuya one
- [x] Docs updated

## Unverified or deferred

**The write direction is not separately captured.** The message shape is proven by the read; nobody has watched a `CleanParam` write land on hardware.

It ships as a `method()` rather than an `unverified` write on the judgement that the hazard the unverified-write rule answers does not arise on this data point. That rule is about a write that "is wrong looks exactly like success" — true of a fire-and-forget `ModeCtrlRequest`, where a wrong method number is a different command. DP 154 is the robot's own settings report: a frame it does not accept leaves `cleanType`, `cleanExtent` and `mopLevel` exactly where they were, so the write answers for itself through the three reads beside it.

Worth one run on a T2351 to close it. If a reviewer reads the rule the other way, the change to make is `writeOnly`/`unverified` on three value members instead — say so and I will.

https://claude.ai/code/session_019EJQDC24Nkh9rNKQw2xCnM

---
_Generated by [Claude Code](https://claude.ai/code/session_019EJQDC24Nkh9rNKQw2xCnM)_


martijnpoppen:
Merged `beta-0.2.0` in (it had moved to 3ac6dbb) so CI proves the actual merge result rather than the base as it was at open. `npm run verify` green locally against the new base: 3706 tests, 202 files.

Worth flagging on timing: the consumer side merged first — mega-yfue/homey-eufy-app#27 is in `main` there — so that repo's CI is red on `gen:flow --check` (`changed: vacuum`) until this lands. Its workflow checks the SDK out at `beta-0.2.0`, so the committed Flow cards name a method the checked-out SDK does not have yet. Nothing to fix on either side; merging this clears it.

Two things unchanged from the description and still worth a reviewer's eye: the write direction has not been watched on hardware, and the call is deliberately all-three-at-once. Both are argued in **Unverified or deferred** above.

---

max246 (approved):
Approving — this is careful, well-grounded work.

**Verified locally against `beta-0.2.0`:**
- Constants line up exactly: `CLEAN_PARAM_FIELD` `{CONFIGURED:1, CLEAN_TYPE:1, CLEAN_EXTENT:3, MOP_MODE:4, MOP_LEVEL:1, VALUE:1}` matches both the encoder and the spec's byte fixtures, and `cleanParamWireValue` inverts the same `CLEAN_TYPE`/`CLEAN_EXTENT`/`MOP_LEVEL` maps the reads decode through — so a write and its observing read can't drift.
- Proto3 handling is right: wrapper always emitted, inner varint only when non-zero, so an empty wrapper is the enum's zero member — and the specs assert that byte-for-byte (`sub(TYPE, [])`) plus round-trip back through `decodeCleanType`/`decodeCleanParamValue`.
- `clean_times`(7) and `fan`(6) correctly left unwritten; `setCleanParam` present on the AIoT line and absent on Tuya, both asserted.
- `npm run verify` green (3706 tests), **and** `cd docs && npm run build` green — the `{@link VACUUM_CLEAN_MEMBERS.setCleanParam}` references resolve, so no repeat of the typedoc-exit-4 docs-build trap.

**On method-vs-unverified:** agree with shipping it as a `method()`. The reasoning holds — DP 154 is the robot's own settings report, so a rejected frame leaves the three reads beside it unchanged rather than silently doing the wrong thing, which is the specific hazard the unverified-write rule guards. The one residual assumption is that `CleanParamRequest` field 1 carries the same `CleanParam` shape the *response* field 1 does; it's the common proto pattern and clearly disclosed, and the adjacent reads would expose a no-op, so it's an acceptable ship.

**One non-blocking note:** `setCleanParam` accepts `sweepThenMop`, which is itself a `⚠️ unverified` read value (CLEAN_TYPE value 3, vendor enum, not yet observed). A caller picking it inherits that uncertainty — but it's self-revealing through `cleanType` beside it, consistent with the PR's own write-direction caveat, so nothing to change.

Merging this leaves only the "one run on a T2351" to close the write direction. I've got no T2351 on the bench here, but if a real device is available I'm happy to script the capture and confirm the settings actually move.

PR #212, opened 2026-09-20
fix(p2p): use the device's retransmissions instead of discarding them

sapireli:
A Wi-Fi camera on a lossy link showed one frame and then froze in every player, while the official app played it correctly from the same network at the same time.

The DATA transport is acknowledged: we ACK every datagram and the device resends anything unacknowledged. Reassembly is forward-only, so on a missing datagram `onData`:

1. advances the high-water mark past the hole,
2. discards the half-assembled frame,
3. drops the resend when it arrives, because it now reads as behind and within `STALE_RETRANSMIT_DEPTH`.

One lost datagram therefore costs the whole frame, permanently, even though the device resent the missing piece. Decoders report this as `Could not find ref with POC` / `Error constructing the frame RPS`. The existing doc comment describes the behaviour: a retransmission "is acknowledged, then ignored".

A packet capture of the official app streaming the same camera shows 414 ACKs against 662 data packets, so it depends on these retransmissions.

## Change

A forward gap now means the datagram has not arrived yet rather than that it is lost. Datagrams that overtook it are held, delivery stays in sequence, and the resend completes the frame.

Once `REORDER_WAIT_MS` passes with no resend, or `REORDER_MAX_DATAGRAMS` accumulate behind the hole, it is treated as lost and the previous behaviour applies: trace the gap, drop the pending frame, resume from the earliest held datagram. A genuinely lost datagram cannot stall the stream.

Sequence restart and duplicate handling are unchanged in intent. The "behind" branch now means a resend of something already delivered, since holes are waited on rather than skipped.

Measured on that camera over 10s:

- frames decoded: frozen on one frame, now 250/250
- decoder frame-reference errors: several hundred, now a few dozen

## Tests

The two cases that asserted a gap is dropped immediately now advance the timer to the point the datagram is treated as lost. A new case covers the fix: a datagram overtaken by its successor and then resent completes the frame with no gap traced.

Typecheck clean. Suite passes, apart from a pre-existing failure in `check-doc-snippets.spec.ts` that is also present on `beta-0.2.0`.


src/transport/p2p/p2p-session.ts
```diff
@@ -1672,20 +1692,102 @@ export class P2PSession extends EventEmitter {
...
+    }
+    if (!held.size) this.clearReorderTimer(dataType);
+  }
+
+  /**
+   * The missing datagram was not repeated in time: report the gap, drop the frame it belonged to, and
+   * resume from the earliest datagram still held so the stream keeps moving.
+   */
+  private abandonHole(dataType: number): void {
+    this.clearReorderTimer(dataType);
```
lenoxys: Blocker: this clears the data type's only timer, and nothing re-arms it while datagrams stay held behind a second hole. Feed 40, 42, 43, 45 and let it fire: 42-43 are delivered, 45 sits with zero timers until another out-of-order datagram arrives, which on a quiet data type like control is never. The overflow return in holdForRetransmit ends the same way. Re-arm whenever anything is still held after a drain, and add a spec for this sequence that fails on this head.

src/transport/p2p/p2p-session.ts
```diff
@@ -1672,20 +1692,102 @@ export class P2PSession extends EventEmitter {
...
+    if (!held?.size) return;
+    for (;;) {
+      const next = ((this.lastSeqByType.get(dataType) ?? 0) + 1) & 0xffff;
+      const body = held.get(next);
+      if (body === undefined) break;
+      held.delete(next);
+      this.lastSeqByType.set(dataType, next);
+      this.reassemble(dataType, body);
+    }
+    if (!held.size) this.clearReorderTimer(dataType);
```
lenoxys: Blocker: filling a hole here leaves the old hole's timer running when a later hole is still open, so that hole only gets what is left of the 250ms (fill at 240ms and the next hole is abandoned 10ms later). Make drain own the timer: clear it, then arm a fresh one if anything is still held. Cover it with a spec that advances just short of a full wait after the fill.

src/transport/p2p/p2p-session.ts
```diff
@@ -366,6 +378,9 @@ export class P2PSession extends EventEmitter {
   /** Last datagram sequence number seen per dataType — used to detect a lost/reordered datagram
    * mid-frame and drop the (now unrecoverable) partial frame instead of splicing wrong bytes. */
   private readonly lastSeqByType = new Map<number, number>();
+  /** Datagrams that arrived past a hole, held per data type until the missing one is retransmitted. */
+  private readonly reorderByType = new Map<number, Map<number, Buffer>>();
```
lenoxys: Cut: this adds two more per-data-type maps beside lastSeqByType and pendingByDataType, and every path keeps them in step by hand (clearReorder, clearReorderTimer, the loop in resetInboundSequencing). Fold held and timer into one record per data type, drop clearReorder, which has one caller, and reset through the same helper. The lastSeqByType JSDoc above now means last delivered; update it.

lenoxys (commented):
The new retransmit spec fails on the base, but the timer belongs to the data type, not the hole: datagrams held behind a second hole get stranded, and a later hole inherits whatever is left of the first hole's wait. Fix both with specs that fail on this head, then collapse the new state.

Fix the PR body too. 414 ACKs against 662 data packets does not show the app relies on retransmission; count repeated sequence numbers in the capture instead. The full suite, check-doc-snippets included, passes on this head, so drop the note about a pre-existing failure.

src/transport/p2p/p2p-session.ts
```diff
@@ -1672,20 +1692,102 @@ export class P2PSession extends EventEmitter {
     this.send(addr, RequestMessageType.ACK, buildAckPayload(this.ackTypeHeader(dataType), seqNo));
 
     const prevSeq = this.lastSeqByType.get(dataType);
-    const advance = prevSeq === undefined ? 1 : (seqNo - prevSeq) & 0xffff;
+    if (prevSeq === undefined) {
```
lenoxys: Cut: base computed `prevSeq === undefined ? 1 : ...`, which sends the first datagram through the advance-1 branch. Restore that and delete this branch. Drain does nothing when nothing is held.

src/transport/p2p/__tests__/data-reassembly.spec.ts
```diff
@@ -13,6 +13,8 @@ const STATION_SN = "T8000P0000000000";
 const P2P_DID = "XXXXXXX-000000-XXXXX";
 const ADDRESS = { host: "127.0.0.1", port: 1 };
 const VIDEO_DATA_TYPE = 1;
+/** Mirrors REORDER_WAIT_MS in p2p-session.ts: how long a hole waits for the device to repeat it. */
+const REORDER_WAIT_MS = 250;
```
lenoxys: Cut: this restates REORDER_WAIT_MS, so the two drift apart silently. Use vi.advanceTimersToNextTimer() in both cases and delete the constant. One fact, one name.

src/transport/p2p/p2p-session.ts
```diff
@@ -1672,20 +1692,102 @@ export class P2PSession extends EventEmitter {
...
-    if (advance > SEQUENCE_LOOKBACK && !restarted) return;
-    this.lastSeqByType.set(dataType, seqNo);
-    if ((advance > 1 || restarted) && this.pendingByDataType.has(dataType)) {
-      if (this.tracedDatagramGaps++ < MAX_TRACED_DATAGRAM_GAPS) {
-        this.trace({ phase: restarted ? "sequence-restart" : "datagram-gap", dataType });
-      }
+    if (advance > SEQUENCE_LOOKBACK) {
+      // Numbered behind the last datagram delivered. Because a hole is now waited on rather than skipped
+      // past, this is a repeat of something already reassembled — unless the numbering itself restarted.
+      if (!(0x10000 - advance > STALE_RETRANSMIT_DEPTH)) return;
```
lenoxys: Nit: write `<=` instead of negating `>`. The next line is a behaviour change: base traced a restart only when it dropped a pending frame, this traces every restart, and abandonHole now traces a gap with nothing pending too. Both spend the MAX_TRACED_DATAGRAM_GAPS budget. Keep the pending guard or say in the PR why it goes.

src/transport/p2p/p2p-session.ts
```diff
@@ -1672,20 +1692,102 @@ export class P2PSession extends EventEmitter {
...
     if (advance === 0) return;
-    const restarted = advance > SEQUENCE_LOOKBACK && 0x10000 - advance > STALE_RETRANSMIT_DEPTH;
-    if (advance > SEQUENCE_LOOKBACK && !restarted) return;
-    this.lastSeqByType.set(dataType, seqNo);
-    if ((advance > 1 || restarted) && this.pendingByDataType.has(dataType)) {
-      if (this.tracedDatagramGaps++ < MAX_TRACED_DATAGRAM_GAPS) {
-        this.trace({ phase: restarted ? "sequence-restart" : "datagram-gap", dataType });
-      }
+    if (advance > SEQUENCE_LOOKBACK) {
+      // Numbered behind the last datagram delivered. Because a hole is now waited on rather than skipped
```
lenoxys: Nit: delete this body comment. The JSDoc above already covers the restart case, and "now waited on rather than skipped past" is history. AGENTS.md: "no inline body comments" and "A JSDoc does not narrate its own history."

src/transport/p2p/p2p-session.ts
```diff
@@ -1661,9 +1675,15 @@ export class P2PSession extends EventEmitter {
...
-   * for that data type is discarded, and the frame is rebuilt from the next header.
+   * A forward gap means a datagram has not arrived YET, which is not the same as lost. Because the device
+   * repeats whatever goes unacknowledged, the missing datagram is usually about to arrive again — so the
+   * ones that overtook it are held and delivery stays strictly in sequence, letting the repeat slot into
+   * place and complete the frame. A logical frame's payload spans datagrams that carry no header of their
+   * own, so reassembling around a hole is impossible; waiting for it is what keeps the frame whole.
+   *
+   * Only once {@link REORDER_WAIT_MS} passes, or {@link REORDER_MAX_DATAGRAMS} pile up, is the datagram
+   * treated as truly lost: the gap is traced, whatever was pending for that data type is discarded, and the
+   * frame is rebuilt from the next header — the behaviour that previously applied to every gap.
```
lenoxys: Nit: drop "the behaviour that previously applied to every gap". AGENTS.md: "A JSDoc does not narrate its own history." That belongs in the commit message.

src/transport/p2p/p2p-session.ts
```diff
@@ -197,6 +197,18 @@ const SEQUENCE_LOOKBACK = 0x8000;
...
+/**
+ * How long a data type waits for a missing datagram to be retransmitted before giving up on it, and how
+ * many later datagrams it will hold while waiting.
+ *
+ * The transport is acknowledged: the device repeats what we do not acknowledge, so a datagram missing from
+ * the sequence is usually about to arrive again rather than gone. Holding the ones that overtook it for a
+ * moment is what lets the repeat be used; without that wait a single lost datagram costs the whole logical
+ * frame it belonged to. The bound keeps a datagram that really is lost from stalling the stream.
+ */
+const REORDER_WAIT_MS = 250;
```
lenoxys: Nit: 250 and 128 come with no stated basis. Put the observed device resend delay that 250ms covers in the JSDoc (AGENTS.md: "the protocol fact that makes it so"), and give REORDER_MAX_DATAGRAMS its own line saying what the bound protects.

src/transport/p2p/p2p-session.ts
sapireli: Fixed. The 40, 42, 43, 45 test fails on the old head: after the first timer delivers 42 and 43, 45 never drains. The new per-type record re-arms a full timer whenever held datagrams remain. A second test fills the first hole at 240 ms and proves the later hole still gets its own 250 ms.

src/transport/p2p/p2p-session.ts
sapireli: I do not have a measured resend-delay distribution, so I have not presented 250 ms as device timing. The JSDoc and PR now call it a bounded SDK wait policy. The 128-datagram limit has its own JSDoc: it bounds retained memory and forces progress when traffic outruns the timer. The unsupported ACK-count inference is removed from the PR.

src/transport/p2p/p2p-session.ts
```diff
@@ -1674,18 +1694,102 @@ export class P2PSession extends EventEmitter {
...
+
+  /** Hold a datagram that arrived past a hole, and arm the wait for the missing one to be repeated. */
+  private holdForRetransmit(dataType: number, seqNo: number, body: Buffer): void {
+    let reorder = this.reorderByType.get(dataType);
+    if (!reorder) {
+      reorder = { held: new Map<number, Buffer>() };
+      this.reorderByType.set(dataType, reorder);
+    }
+    if (!reorder.held.has(seqNo)) reorder.held.set(seqNo, body);
+    if (reorder.held.size > REORDER_MAX_DATAGRAMS) {
```
lenoxys: Blocker: nothing tests this path. Deleting the abandon call leaves the suite green. Add a spec that holds 129 datagrams behind one hole and checks they all arrive with no timer advance and no timer left armed. The wrap-aware sort in abandonHole has the same gap: a plain a - b also passes. Add a hole that straddles 0xffff.

src/transport/p2p/p2p-session.ts
```diff
@@ -1674,18 +1694,102 @@ export class P2PSession extends EventEmitter {
...
-    if ((advance > 1 || restarted) && this.pendingByDataType.has(dataType)) {
-      if (this.tracedDatagramGaps++ < MAX_TRACED_DATAGRAM_GAPS) {
-        this.trace({ phase: restarted ? "sequence-restart" : "datagram-gap", dataType });
-      }
+    if (advance > SEQUENCE_LOOKBACK) {
+      if (0x10000 - advance <= STALE_RETRANSMIT_DEPTH) return;
+      if (this.pendingByDataType.has(dataType) && this.tracedDatagramGaps++ < MAX_TRACED_DATAGRAM_GAPS)
+        this.trace({ phase: "sequence-restart", dataType });
+      this.clearReorderTimer(dataType);
+      this.reorderByType.delete(dataType);
```
lenoxys: Blocker: deleting these two lines, or the two in resetInboundSequencing, leaves the suite green. Either way an old-numbering datagram stays held, and its timer later moves lastSeqByType back onto the old numbering. In the restart and reconnect specs, hold a datagram behind a hole first, then advance timers and assert nothing else arrives.

src/transport/p2p/p2p-session.ts
```diff
@@ -1674,18 +1694,102 @@ export class P2PSession extends EventEmitter {
...
+    this.clearReorderTimer(dataType);
+    const reorder = this.reorderByType.get(dataType);
+    if (!reorder?.held.size) return;
+    const held = reorder.held;
+    if (this.pendingByDataType.has(dataType) && this.tracedDatagramGaps++ < MAX_TRACED_DATAGRAM_GAPS)
+      this.trace({ phase: "datagram-gap", dataType });
+    this.pendingByDataType.delete(dataType);
+    const last = this.lastSeqByType.get(dataType) ?? 0;
+    const earliest = [...held.keys()].sort((a, b) => ((a - last) & 0xffff) - ((b - last) & 0xffff))[0];
+    const body = held.get(earliest) as Buffer;
```
lenoxys: Cut: these lines repeat one pass of the drainReorder loop. Set lastSeqByType to earliest - 1 and call drainReorder, which already delivers, re-arms and deletes. That also fixes a leak: when abandoning empties held, drainReorder returns early, so an empty record stays in reorderByType.

src/transport/p2p/p2p-session.ts
```diff
@@ -197,6 +197,18 @@ const SEQUENCE_LOOKBACK = 0x8000;
...
  */
 const STALE_RETRANSMIT_DEPTH = 1024;
+
+/**
+ * Maximum time a data type waits for a missing datagram before resuming from the next held one.
+ *
+ * The transport is acknowledged: the device repeats what we do not acknowledge, so a datagram missing from
+ * the sequence is usually about to arrive again rather than gone. Holding the ones that overtook it for a
+ * moment is what lets the repeat be used; without that wait a single lost datagram costs the whole logical
+ * frame it belonged to. The 250 ms limit bounds the added delay if no repeat arrives.
```
lenoxys: Cut: this repeats the onData JSDoc and uses "we". Cut it to one line: what the wait bounds, and that 250 ms is a chosen bound, not a measured resend delay. The PR body says that but the JSDoc does not. "Usually about to arrive again" is also unmeasured, so drop it.

lenoxys (commented):
Both timer blockers are fixed and the two new specs fail on 7e0f5ed. The overflow, restart and reconnect paths still have no spec: deleting any of them leaves the whole suite green. Add those specs, then shrink abandonHole and the wait JSDoc.

Both commits are unsigned. CONTRIBUTING.md requires signed commits, so re-sign both and force-push.

src/transport/p2p/p2p-session.ts
sapireli: The current revision keeps held datagrams and their timer together in one reorderByType record per data type, uses clearReorderTimer during reset, and documents lastSeqByType as the last delivered sequence number. The extra maps and clearReorder helper from the first revision are gone.

src/transport/p2p/p2p-session.ts
sapireli: The wait JSDoc is now one line and explicitly says 250 ms is a chosen bound, not a measured device resend delay. REORDER_MAX_DATAGRAMS has its own JSDoc stating that 128 bounds retained memory and delay. I have not inferred a device resend distribution from ACK counts.

src/transport/p2p/p2p-session.ts
sapireli: Added a spec that holds 129 complete datagrams behind a hole across 0xffff, then checks immediate delivery of all 129 in modular sequence order and zero remaining timers. I verified the spec fails if the overflow abandon call is removed and also if the wrap-aware comparator is replaced by a plain numeric sort. Full verify passes with 3,710 tests.

src/transport/p2p/p2p-session.ts
sapireli: The restart and reconnect specs now each hold an old-numbering datagram behind a hole before the reset, advance timers afterward, assert no old frame arrives, and assert no timer remains. I checked that removing either held-state cleanup or the reconnect timer cleanup makes the focused spec fail.

src/transport/p2p/p2p-session.ts
sapireli: Addressed in b48aefc: abandonHole sets lastSeqByType to earliest - 1 modulo 16 bits and calls drainReorder. That path delivers the earliest and consecutive held datagrams, re-arms if needed, and deletes the record when empty. The overflow spec checks that no timer or held record behavior remains after all 129 drain.

src/transport/p2p/p2p-session.ts
sapireli: Addressed in b48aefc: the wait JSDoc is one line, states the bound, and explicitly says 250 ms is not a measured device resend delay. The longer sequence and frame behavior remains in onData.

sapireli:

- Replaced the two unsigned commits with signed, GitHub verified equivalents and added signed fix commit [`b48aefc`](https://github.com/sapireli/eufy-sdk/commit/b48aefc695f8078f4b835362650d320d5b7d3d42). The PR still changes only `p2p-session.ts` and its data-reassembly spec, and the final tree matches the locally tested tree exactly.
- Added the 129-datagram overflow test across 16-bit wraparound. It requires immediate in-order delivery and no remaining timer. It fails if overflow abandonment is removed or sorting becomes plain numeric order.
- Strengthened restart and reconnect tests by first holding an old datagram behind a hole, then advancing timers and checking that no old frame or timer remains. Removing the corresponding cleanup makes these tests fail.
- Simplified abandonment to resume through `drainReorder` and reduced the wait JSDoc to the chosen 250 ms bound without claiming measured resend timing.
- Full `npm run verify` passed locally: 3,710 tests in 202 files, plus format, typecheck, guards, build, ESM, snippets, and examples. GNU `paste` and `sed` were supplied on macOS for the guard script.

I replied to every open review thread and GitHub now shows 0 unresolved out of 13. The fork reports no check runs; the local full-gate result is the available verification here.

lenoxys (commented):
Close to ready: 12 of 13 old threads are fixed, verify is green on b48aefc and all three commits are signed. Still open: the wait-JSDoc thread asked to drop "usually about to arrive again" as unmeasured, and it still sits in the onData JSDoc.
The diff since my last review is the three specs I asked for and a smaller abandonHole. Nothing new in scope. One cut and two nits below cover code that can never run.

src/transport/p2p/p2p-session.ts
```diff
@@ -1674,18 +1687,99 @@ export class P2PSession extends EventEmitter {
...
+  }
+
+  /**
+   * The missing datagram was not repeated in time: report the gap, drop the frame it belonged to, and
+   * resume from the earliest datagram still held so the stream keeps moving.
+   */
+  private abandonHole(dataType: number): void {
+    this.clearReorderTimer(dataType);
+    const reorder = this.reorderByType.get(dataType);
+    if (!reorder?.held.size) return;
```
lenoxys: cut: delete this guard and the clearReorderTimer call above it. Both callers get here only with datagrams held (the timer is armed only when something is held, and overflow means more than 128), and drainReorder clears the timer first anyway.

src/transport/p2p/p2p-session.ts
```diff
@@ -1674,18 +1687,99 @@ export class P2PSession extends EventEmitter {
...
+    const held = reorder.held;
+    if (this.pendingByDataType.has(dataType) && this.tracedDatagramGaps++ < MAX_TRACED_DATAGRAM_GAPS)
+      this.trace({ phase: "datagram-gap", dataType });
+    this.pendingByDataType.delete(dataType);
+    const last = this.lastSeqByType.get(dataType) ?? 0;
```
lenoxys: nit: lastSeqByType is always set when anything is held, because a datagram is only held once prevSeq is defined. Drop the `?? 0` fallback here and in drainReorder instead of inventing a sequence number.

src/transport/p2p/__tests__/data-reassembly.spec.ts
```diff
@@ -130,33 +216,51 @@ describe("P2P data reassembly", () => {
...
-    feed(dataPacket(40_000, commandFrame(40_000, 1300, payload)));
-    await close();
-    feed(dataPacket(5, commandFrame(5, 1300, payload)));
+      feed(dataPacket(40_000, commandFrame(40_000, 1300, payload)));
+      feed(dataPacket(40_002, commandFrame(40_002, 1300, Buffer.from([9]))));
+      await close();
+      expect(vi.getTimerCount()).toBe(0);
+      feed(dataPacket(5, commandFrame(5, 1300, payload)));
+      feed(dataPacket(6, commandFrame(6, 1301, Buffer.from([1]))));
+      vi.advanceTimersByTime(250);
```
lenoxys: nit: this literal 250 repeats REORDER_WAIT_MS, and the constant was removed from the spec for exactly that reason. Use vi.runAllTimers() here and at the same line in the restart spec.

src/transport/p2p/p2p-session.ts
sapireli: Removed the redundant timer clear and empty-held guard. Both call paths reach abandonHole with held datagrams; drainReorder owns timer cleanup.

src/transport/p2p/p2p-session.ts
sapireli: Removed the invented zero fallbacks in abandonHole and drainReorder. Held datagrams imply lastSeqByType was initialized, and the code now uses that invariant.

src/transport/p2p/__tests__/data-reassembly.spec.ts
sapireli: Replaced the literal timer advance with vi.runAllTimers() in both specs, so the tests follow the policy constant.

sapireli:
Final update: the per-data-type reorder record gives each remaining gap a fresh bounded wait, drains on overflow in 16-bit sequence order, and clears held datagrams and timers on restart or reconnect. The latest review cuts remove redundant timer cleanup, a dead guard, invented zero sequence fallbacks, and duplicated 250 ms test literals. The 250 ms bound is an SDK policy rather than a measured device retransmit interval. Full npm run verify passed with 3,710 tests across 202 files; all current PR commits are GitHub verified signed.

PR #214, opened 2026-09-20
feat(p2p): let a caller require a camera's media path to stay on the LAN

sapireli:
I stream several cameras at once to a video wall on my own network, and I want that video to stay on the network. There is currently no way to require it.

A connect races the local and cloud lookup channels and keeps whichever peer answers first. That is a reasonable default, but it means a session can settle on a relay or on the camera's public address while the camera is on the same LAN. With several continuous feeds that is a large amount of traffic leaving and re-entering the network unnecessarily.

## Change

`lanOnlyForStation` names the CIDR a peer must fall inside:

```ts
const eufy = new EufyMega({ email, password, lanOnlyForStation: () => "192.168.1.0/24" });
```

A peer outside the block is refused and the lookup continues, so a local peer answering later is still used. Returning `undefined`, the default, keeps current behaviour.

It is consulted per session rather than fixed at construction, so a caller can prefer the LAN and keep a relay as a fallback: restrict the first attempt, return `undefined` on the next. That is how a P2P-only toggle is built on top of it.

The refusal sends `END` rather than dropping the peer. The device opens a session as soon as it answers `CAM_ID`, so ignoring it leaves one stranded. Standalone cameras have few session slots and begin refusing new connects once they are used up.

Also adds `ipv4InCidr` to `lan-ip.ts`. A malformed address or block matches nothing, so a typo cannot widen what counts as local.

## Tests

No behaviour change unless the option is set. Typecheck clean. Suite passes, apart from a pre-existing failure in `check-doc-snippets.spec.ts` that is also present on `beta-0.2.0`. `lan-only-sessions.spec.ts` covers keep, refuse and `END`, recovery onto a later local peer, the unrestricted default, and the CIDR matcher's edge cases.

Naming: I followed the existing `stationSn` / `parentSn` vocabulary. Happy to rename if you prefer something else.


src/transport/p2p/p2p-session.ts
```diff
@@ -847,6 +858,15 @@ export class P2PSession extends EventEmitter {
 
   private onConnected(addr: Address): void {
     if (this.connected) return;
+    if (this.cfg.lanOnlyCidr && !ipv4InCidr(addr.host, this.cfg.lanOnlyCidr)) {
```
lenoxys: Blocker: this check runs only at CAM_ID. sendLookups asks the cloud again every second, and each LOOKUP_ADDR hole-punches the public/relay address on 7 ports. The station opens a session, then gets END, about 15 times per connect within CONNECT_TIMEOUT_MS, on the same slot-limited devices the PR body worries about.
Filter in beginCheckCam so an out-of-block candidate is never punched. Keep this guard only as a backstop for TURN_SERVER_CAM_ID, and add a spec that feeds a LOOKUP_ADDR and asserts no CHECK_CAM goes out.

src/transport/p2p/lan-ip.ts
```diff
@@ -47,3 +47,32 @@ export function freshestLanIp(raw: unknown): string | undefined {
...
+/**
+ * Whether an IPv4 address falls inside a CIDR block (e.g. `192.168.1.0/24`).
+ *
+ * Used to decide whether a peer that answered a connect is on the caller's own network. Anything that is
+ * not a well-formed address and prefix is treated as no match, so a malformed block cannot silently widen
+ * what counts as local.
+ */
+export function ipv4InCidr(ip: string, cidr: string): boolean {
+  const [network, bitsText] = cidr.split("/");
+  const bits = Number(bitsText);
```
lenoxys: Blocker: `192.168.1.0/` and `192.168.1.0/ ` parse as /0 because Number("") and Number(" ") are 0, so every address counts as local. That contradicts the JSDoc promise above. The cut on line 58 removes this parser. If you keep one, reject a non-digit prefix and add both cases to the malformed-block spec.

src/transport/p2p/lan-ip.ts
```diff
@@ -47,3 +47,32 @@ export function freshestLanIp(raw: unknown): string | undefined {
...
 }
+
+/**
+ * Whether an IPv4 address falls inside a CIDR block (e.g. `192.168.1.0/24`).
+ *
+ * Used to decide whether a peer that answered a connect is on the caller's own network. Anything that is
+ * not a well-formed address and prefix is treated as no match, so a malformed block cannot silently widen
+ * what counts as local.
+ */
+export function ipv4InCidr(ip: string, cidr: string): boolean {
```
lenoxys: Cut: delete ipv4InCidr. The barrel's `export * from "./lan-ip.js"` makes it permanent public API. AGENTS.md: "Search before creating... including the primitive APIs it uses". isPrivateIpv4, a few lines up, already defines LAN for freshestLanIp, so filter on that.
If a caller-named block turns out to be a real need, use node:net BlockList, which transport/http/media-download.ts already imports. It validates the prefix and throws instead of widening.

src/client/types.ts
```diff
@@ -66,6 +66,25 @@ export interface EufyMegaOptions extends MegaClientConfig {
...
+   * expires.
+   *
+   * It is consulted per session, so a host can prefer LAN and fall back to a relay by returning the CIDR
+   * for a first attempt and `undefined` for the next.
+   *
+   * ```ts
+   * const eufy = new EufyMega({ email, password, lanOnlyForStation: () => "192.168.1.0/24" });
+   * ```
+   */
+  lanOnlyForStation?: (stationSn: string) => string | undefined;
```
lenoxys: Cut: with isPrivateIpv4 this becomes `lanOnly?: (stationSn: string) => boolean`. The current name reads as a boolean but returns a CIDR string, and the session calls the same fact lanOnlyCidr. One fact, one name. In the PR body, list what the RFC-1918 check skips (CGNAT, overlay networks) and add a block option only when a caller needs one.

src/client/types.ts
```diff
@@ -66,6 +66,25 @@ export interface EufyMegaOptions extends MegaClientConfig {
    * where the record's IP is wrong/blocked (AP isolation, a stale `ip_addr`).
    */
   localAddresses?: Record<string, string>;
+  /**
+   * Require a station's media path to stay on your own network, by returning the CIDR a peer must fall
```
lenoxys: Cut: the same rationale appears three times (here, P2PRouterDeps, P2PSessionConfig) plus the spec header, about 30 lines of prose for 10 lines of logic. Keep one short paragraph here. AGENTS.md: "Second person (`you`, `your host`) never appears", and a JSDoc states the declaration, "NOT who will call it, what a caller might do with it". Drop "your own network", the privacy/bandwidth/metered-uplink motives and the host-fallback sentence.

src/transport/p2p/command-router.ts
```diff
@@ -223,6 +223,14 @@ export interface P2PRouterDeps {
...
   /** LAN address overrides for direct P2P, keyed by parent-station serial (host or host:port). */
   localAddresses?: Record<string, string>;
+  /**
+   * Restrict a station's session to a direct-LAN peer, by returning the CIDR it must fall inside.
+   *
+   * Consulted per session, so a caller can turn the restriction on or off for a station between connects
+   * (a host that prefers a local path but keeps a relay as a fallback, say). Return undefined to accept
+   * whichever peer answers first, which is the default.
+   */
+  lanOnlyForStation?: (stationSn: string) => string | undefined;
```
lenoxys: Cut: make this JSDoc one line with a {@link} to the EufyMegaOptions field. The "a host that prefers a local path" clause describes the audience, not the declaration.

src/transport/p2p/p2p-session.ts
```diff
@@ -249,6 +250,16 @@ export interface P2PSessionConfig {
...
+  /**
+   * Restrict this session to a peer inside this CIDR (e.g. `192.168.1.0/24`) — a direct-LAN path only.
+   *
+   * A lookup races the local and cloud channels and the session keeps whichever peer answers first, which
+   * may be a relay or the station's public address. Where a caller requires the media path to stay on its
+   * own network — for privacy, for bandwidth, or because it is paying for the uplink — a peer outside the
+   * block is refused: the station is told END (it opened a session the moment it answered) and the session
+   * keeps looking until a local peer answers or the connect expires.
+   */
+  lanOnlyCidr?: string;
```
lenoxys: Cut: one line is enough here. "for privacy, for bandwidth, or because it is paying for the uplink" is caller motive, which AGENTS.md keeps out of JSDoc.

src/transport/p2p/__tests__/lan-only-sessions.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+    answerFrom("192.168.1.50");
+    expect(session.isConnected).toBe(true);
+  });
+
+  it("refuses a peer outside it, and tells the station to drop the session it opened", () => {
+    const { session, sent, answerFrom } = harness(LAN);
+    answerFrom("203.0.113.9");
+
+    expect(session.isConnected).toBe(false);
+    expect(sent.some(({ addr, type }) => addr.host === "203.0.113.9" && type.equals(RequestMessageType.END))).toBe(true);
```
lenoxys: Nit: `npm run verify` stops at format:check on this file. Run prettier --write on it. The PR body's "pre-existing failure in check-doc-snippets" does not reproduce: check:snippets and the full suite (about 3700 tests) pass on this head.

lenoxys (commented):
Two blockers. An empty CIDR prefix widens locality to every address, and the refusal fires only at CAM_ID, so the out-of-block peer gets re-punched and sent END on every lookup tick.
Ponytail: the new public ipv4InCidr duplicates what isPrivateIpv4 in the same file or node:net BlockList already answer. This +165/-0 should shrink to a boolean option plus a filter before the hole-punch. `npm run verify` is red on prettier.

src/transport/p2p/p2p-session.ts
sapireli: Fixed at the lookup candidate, before CHECK_CAM. The new spec feeds a public LOOKUP_ADDR and asserts zero CHECK_CAM sends, then feeds a private candidate and sees the handshake. The CAM_ID check remains a backstop for unsolicited replies. I also replaced the CIDR parser with the existing isPrivateIpv4 helper and made the public option a per-station boolean.

src/transport/p2p/command-router.ts
```diff
@@ -494,6 +496,7 @@ export class P2PCommandRouter {
       p2pDid: did,
       cloudAddresses: conn ? decodeP2PCloudIPs(conn) : undefined,
       localAddress,
+      lanOnly: this.deps.lanOnly?.(stationSn),
```
lenoxys: Nit: no test covers this line or the matching one in eufy-mega.ts. Delete either and the option silently does nothing, and the full suite still passes. Add one spec that opens a session through the router with lanOnly returning true, feeds it a public LOOKUP_ADDR and asserts no CHECK_CAM goes out.

lenoxys (commented):
The eight threads are addressed. The filter in beginCheckCam covers LOOKUP_ADDR, LOCAL_LOOKUP_RESP and the port loop, and the new specs fail on base. Verify is green.
Blocker: e7cc2e7 is unsigned and CONTRIBUTING.md requires signed commits. Re-sign it with -S and force-push. Add a commit body saying what changed and why while you are there.

src/transport/p2p/command-router.ts
sapireli: Addressed in signed commit 0adbf48. The new client-level spec constructs EufyMega with lanOnly returning true, opens a station session through its P2P router, feeds a public LOOKUP_ADDR, and asserts that no CHECK_CAM is sent. I verified the spec fails if either the facade-to-router line or the router-to-session line is removed. Full npm run verify passes with 3,712 tests in 204 files.

sapireli:

- Replaced the unsigned commit with GitHub verified signed commit [`0adbf48`](https://github.com/sapireli/eufy-sdk/commit/0adbf48f64c8dc544aa28e21b40aacd2c4a2cdd1), including a commit body explaining the private-peer policy and cloud lookup behavior.
- Added a client-level spec that opens a station session through `EufyMega` and `P2PCommandRouter`, then feeds a public `LOOKUP_ADDR` and proves no `CHECK_CAM` is sent. It fails if either the facade-to-router or router-to-session `lanOnly` wiring is removed. The existing session spec covers a private candidate being punched and a public `CAM_ID` being refused.
- Full `npm run verify` passed locally: 3,712 tests in 204 files, plus format, typecheck, guards, build, ESM, snippets, and examples. GNU `paste` and `sed` were supplied on macOS for the guard script.

I replied to the remaining open review thread and GitHub now shows 0 unresolved out of 9. The fork reports no check runs, so the local full-gate result is the available verification here.

src/transport/p2p/__tests__/lan-only-sessions.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+    const { lookupAddress, sent } = harness(true);
+    lookupAddress("203.0.113.9");
+    expect(sent.some(({ type }) => type.equals(RequestMessageType.CHECK_CAM))).toBe(false);
+    lookupAddress("192.168.1.50");
+    expect(
+      sent.some(({ addr, type }) => addr.host === "192.168.1.50" && type.equals(RequestMessageType.CHECK_CAM)),
+    ).toBe(true);
+  });
+
+  it("keeps a private peer", () => {
```
lenoxys: cut: delete this test. "still settles on a local peer that answers after one was refused" already asserts that a private peer is kept, and this one passes without the restriction.

src/transport/p2p/p2p-session.ts
```diff
@@ -837,6 +840,7 @@ export class P2PSession extends EventEmitter {
   }
 
   private beginCheckCam(addr: Address): void {
+    if (this.cfg.lanOnly && !isPrivateIpv4(addr.host)) return;
```
lenoxys: cut: this drop is silent, and a debug log is off by default. When lanOnly refuses every candidate, make the connect timeout error say so: name the refused hosts and the fix, set localAddresses for this station or turn lanOnly off for it.

src/client/types.ts
```diff
@@ -66,6 +66,11 @@ export interface EufyMegaOptions extends MegaClientConfig {
    * where the record's IP is wrong/blocked (AP isolation, a stale `ip_addr`).
    */
   localAddresses?: Record<string, string>;
+  /**
+   * Require an RFC-1918 IPv4 peer for a station session. The callback is read for each session;
+   * `false` or absence accepts any peer. CGNAT and overlay addresses outside RFC-1918 are excluded.
+   */
+  lanOnly?: (stationSn: string) => boolean;
```
lenoxys: nit: add a lanOnly row to the options table in docs/connectivity.md, next to the existing localAddresses row.

lenoxys (commented):
No blockers left. All 10 earlier points are fixed (8 first-round threads, the wiring-spec nit, the unsigned commit), with none disputed. 0adbf48 is signed and verify is green.
Since the last review the PR grew by only the 53-line client spec that was asked for, and it adds nothing just-in-case. Two cuts and a nit below.
The PR body says what the change does but not why. Name the station you ran lanOnly against and whether a private LOOKUP_ADDR or LOCAL_LOOKUP_RESP candidate connected.

src/transport/p2p/__tests__/lan-only-sessions.spec.ts
sapireli: Removed the redundant private-peer spec. The remaining test exercises a refused candidate followed by a private peer that answers.

src/transport/p2p/p2p-session.ts
sapireli: The connect timeout now names non-private refused hosts and advises setting localAddresses for the station or disabling lanOnly. A focused timeout spec checks this diagnostic.

src/client/types.ts
sapireli: Added lanOnly next to localAddresses in the connectivity options table.

sapireli:
Final update: lanOnly filters non-private lookup candidates before CHECK_CAM, retains the CAM_ID backstop, and gives a refused-host timeout diagnostic with localAddresses and policy options. I removed the redundant test, added a diagnostic spec, and documented lanOnly in the connectivity options table. Full npm run verify passed with 3,712 tests across 204 files; the latest commit is GitHub verified signed. Read-only checks on a HomeBase 3 and standalone camera timed out on this PR head alone; with #211 also present, both connected to private peers and delivered a video frame. Both LOOKUP_ADDR and LOCAL_LOOKUP_RESP advertised the winning peer, so this evidence does not isolate which response initiated the handshake.

lenoxys:
I'll be direct about scope: I don't think lanOnly belongs in the SDK. Keeping P2P off public and relay peers is network policy, and a firewall rule on the host already does it: drop UDP to non-private addresses. The SDK's job is to turn eufy's protocol into an API that's easy to use, not to re-implement blocking rules, and the recommendation stays not to expose the host running it.

Nothing in the PR shows a need a firewall can't cover. If there's a eufy-specific case, describe the scenario here before we go further. 

#211 is separate and still under review.

sapireli:
I disagree with the scope objection. The proposed host firewall rule is not equivalent to `lanOnly`.

There is a concrete eufy-specific sequence:

1. A station record supplies UDP cloud lookup servers through `p2p_conn`/`app_conn`. A session can send `LOOKUP_WITH_KEY` or `LOOKUP_WITH_KEY2` to a **public** lookup server.
2. The reply can present a private direct candidate alongside public or relay candidates. `lanOnly` allows the broker request, then refuses a non-private candidate **before `CHECK_CAM`** and refuses a public peer at `CAM_ID` as a backstop.
3. A blanket `drop UDP to non-private addresses` rule also drops step 1. It can prevent discovery of the very private peer this option is meant to allow. On a host with multiple SDK clients, it also blocks unrelated accounts and stations whose sessions are allowed to use remote peers. The callback applies the choice to each station session.

I replaced the existing candidate test with one that exercises this distinction: with `lanOnly` enabled, the public broker still receives a lookup, a public candidate receives no `CHECK_CAM`, and a private candidate is punched. Signed commit `3caa0a5`; full `npm run verify` passes with 3,712 tests in 204 files. The PR description now explains the scenario.

Advice about exposing a host does not answer which **outbound P2P peer** the SDK accepts. A more elaborate firewall with broker exceptions can complement this policy, but the blanket rule proposed here cannot express it or the per-station choice. #211 is a separate source-port reliability change; I am not presenting this PR alone as a fix for the observed connection timeout.

lenoxys:
What's the actual need? Who runs into a problem today that lanOnly solves, and what happens to them without it?

sapireli:
The concrete need is a caller that requires a **private P2P peer for one local station** and wants the connect to fail visibly when only public or relay peers answer. Without this option, `onConnected()` adopts the first `CAM_ID` regardless of address. If a public peer answers first, the SDK can carry that session over the public path even though the caller required a private one. `lanOnly` makes that requirement expressible; it does not claim to improve reachability or prove the peer shares the physical subnet.

The choice must be per station. The same host can have another SDK client, even another account, whose remote station needs public peers. A blanket host rule would block that session and the public UDP broker used to discover private candidates for the local station. This is why the protocol-level distinction and callback matter.

The read-only integration check with #211 present connected a HomeBase 3 and a standalone camera to private peers and obtained video; #214 alone timed out in that check. I have kept both facts in the PR body and do not claim that #214 independently fixes connection success. The code and specs establish the policy behavior: broker lookup remains allowed, public candidates are not punched, public `CAM_ID` is refused, and a private peer may be accepted. I added the use case to the PR description.

PR #216, opened 2026-09-20
feat(solix): expose Solarbank batteryCycles (candidate) from the 0xde BMS blob

max246:
## What

Decode a **battery cycle count** (`batteryCycles`) from the Solarbank's `param_info` telemetry — tag `0xde`, a type-`0x04` blob whose payload is a `uint32` LE. Found by a live MQTT capture: the frame carries ~15 BMS blobs the SDK discards, and `0xde` read **31**, constant across the capture, in the tag range where a cumulative counter belongs.

## Why it's cycles (and not something else)

The decisive disambiguation is the pack's **age**: the captured unit is only a **few days old**. A *days-since-install* counter would therefore read single digits — but `0xde` reads **31**. So it's a cumulative count carried over from before install: **charge/discharge cycles** (factory/QA cycles + a few days of use), which is exactly what a BMS reports here. `0xe0` (pack-voltage-ish + max-load) and `0xdf` (rated power `2500` + max-load `800`) were checked and are *not* cycles; `0xdb` climbs too fast (energy/runtime), so it isn't either.

## Honesty / status

- **CANDIDATE**, same bar as `batteryHealth` (#209): the wire position + type are confirmed and the magnitude/age argument is strong, but it is **not yet hardware-labelled**.
- The clincher is a next-day read: `0xde` incrementing **~+1/day** (or +2 on a two-cycle day) against the app's own cycle figure. The device owner is checking the app tomorrow; I'll post the follow-up read here.
- **Not** claiming a capacity crack: the E5000's 5000 Wh is the nameplate (the integration already uses it), and the wire's `2500` is rated power, not capacity — so nothing new to assert there.

## Change

- `addSolarbankScalars` reads `0xde` (guarded on type `0x04` + length) as `uint32`, emitted as `batteryCycles`.
- Solarbank-family only, like the other `a3`/`a4`/`b5` scalars — asserted **absent** on a meter frame.
- Specs: the value in the main Solarbank fixture + a `>255` case proving the `uint32` read (not a byte).

`npm run verify` green (3708 tests) and `cd docs && npm run build` green (no typedoc link breakage).


max246:
Closing without merging — we can't stand behind labelling `0xde` as a cycle count.

Two things settled it:
1. **It didn't increment.** Read `0xde` = 31 two days running (2026-09-20 → 2026-09-21) with no change. That rules out a days/calendar counter (good), but it also means we have no evidence it's a *live* counter vs a static factory value.
2. **The app doesn't decode it as cycles.** The disassembly showed the app sources `batteryCyclesNumber` from HTTP sub-package JSON (null for an integrated AE103), not from this ff09 byte — so nothing corroborates `0xde` = cycles.

`0xde` is a real, stable value on the wire, but "cycles" is an unconfirmed guess, and we don't ship guessed field names (same bar that dropped `batteryCurrent`/`faultStatus`). If it ever ticks 31→32 in step with a real full charge/discharge cycle, we'll revisit with actual evidence.

PR #217, opened 2026-09-20
feat(solix): expose grid power limits + backup reserve (write-readback confirmed)

max246:
Three Solarbank **settings** decoded from `param_info`, each **confirmed live by write-readback** (change it in the app, watch the exact bytes follow):

| field | wire | now | evidence |
|---|---|---|---|
| `gridImportLimit` | `0xdf` u16 @3 | 2500 W | app import limit 2500→1650→2500 tracked `df@3` |
| `gridExportLimit` | `0xdf` u16 @5 | 750→800 W | app export limit 800→750 tracked `df@5` |
| `backupReserve` | `0xb5` (25-byte) `[1]` | 15 % | app reserve 0→5→15 % tracked `b5[1]`, while `b5[2]` (min-SOC) held → distinct field |

No candidates — every byte followed a deliberate app change, several in both directions. Ships as **confirmed**.

## Notes
- `0xdf` stores import at offsets 3 & 13, export at 5 & 9 (each echoed); we read the first of each. Length gate ≥7.
- `0xb5` has two variants: the **4-byte** SETTINGS blob `[discharge, cutoff, charge]` (existing `dischargeLimit`/`chargeLimit`, untouched) and the **25-byte** FAST-frame blob that leads with `[backupReserve, discharge, charge, …]`. This corrects an earlier code note that said the 25-byte bytes "aren't the limits" — they are. Only the write-confirmed `b5[1]` is asserted from that variant.
- Solarbank-family only, like the other `a3`/`a4`/`b5` scalars.

`npm run verify` green (3708 tests) and `cd docs && npm run build` green.

## Follow-up
Read-only for now. The write path (set these from HA) is the same cmd-17 `…/req` channel the SOC-limit / display-timeout writes already use.


lenoxys (commented):
One blocker: the `b5` gate is looser than the variant the evidence covers. One design question on `backupReserve` now having two sources that may not mean the same thing. The rest is house-style cuts. `0xdf` read is fine.

src/transport/mqtt/solix-mqtt.ts
```diff
@@ -329,6 +338,13 @@ function addSolarbankScalars(frame: SolixParamFrame, out: Record<string, number>
   if (b5 && b5[0] === 0x04 && b5.length === 4) {
     out.dischargeLimit = b5[1]!;
     out.chargeLimit = b5[3]!;
+  } else if (b5 && b5[0] === 0x04 && b5.length >= 11) {
```
lenoxys: Blocker. The evidence and the JSDoc are about the 25-byte variant, the gate takes anything from 11 up. The branch above uses an exact length for exactly this reason (different `b5` layouts, same offsets), and this runs for every `AE10*`, not just the AE103 you tested. Make it `=== 25`, and add a spec that a non-25 type-`0x04` blob emits no `backupReserve` so the gate is actually tested.

src/transport/mqtt/solix-mqtt.ts
```diff
@@ -329,6 +338,13 @@ function addSolarbankScalars(frame: SolixParamFrame, out: Record<string, number>
...
+    out.backupReserve = b5[1]!; // % — the FAST-frame b5 leads with the backup reserve
```
lenoxys: Same name as the HTTP `getSafetySocParams().backupReserve`, which comes paired with `backupReserveSwitch`. Is `b5[1]` the configured value or the effective one? Your 0 -> 5 -> 15 run doesn't tell: if 0 was switch-off, MQTT reads 0 where HTTP reads 15, and two sources disagree under one name. Worth one toggle of the switch with the value held. Either way the HTTP JSDoc still says it's the only source for `backupReserve`, that needs updating in this PR.

src/transport/mqtt/solix-mqtt.ts
```diff
@@ -329,6 +338,13 @@ function addSolarbankScalars(frame: SolixParamFrame, out: Record<string, number>
...
+  }
+  const df = frame.fields.get(0xdf);
+  if (df && df[0] === 0x04 && df.length >= 7) {
+    out.gridImportLimit = df.readUInt16LE(3); // max power drawn FROM the grid (W)
```
lenoxys: Cut: inline body comments on 342/346/347 break the AGENTS.md rule, and nothing else in this function has one. The JSDoc already states all three, drop them.

src/transport/mqtt/solix-mqtt.ts
```diff
@@ -304,11 +304,20 @@ export function solixReadings(frame: SolixParamFrame, productCode: string): Reco
...
- *   not the limits — hence the exact length gate.
+ * - **SOC limits** (`dischargeLimit`/`chargeLimit`, %) come from tag `0xb5`'s 4-byte SETTINGS-blob
+ *   variant — type `0x04`, payload `[discharge, output cutoff, charge]`. Confirmed live: moving discharge
+ *   10%→5% moved `b5[1]` 0x0a→0x05 while charge held at `b5[3]`=0x64.
+ * - **Backup reserve** (`backupReserve`, %) comes from tag `0xb5`'s longer FAST-frame variant (the 25-byte
+ *   type-`0x04` blob on the ~7 s telemetry frame), which leads with `[backupReserve, discharge, charge, …]`.
+ *   CONFIRMED live by write-readback: setting the app's backup reserve 0→5→15 % moved `b5[1]` 0x00→0x05→0x0f
+ *   each time, while `b5[2]` (min-SOC) held at 5 — so it is a distinct field, not the discharge floor. Only
+ *   `b5[1]` is asserted from this variant; the exact-length-4 gate keeps the SETTINGS-blob path unchanged.
+ * - **Grid power limits** (`gridImportLimit`/`gridExportLimit`, W) come from tag `0xdf`'s type-`0x04`
```
lenoxys: Nit: 11 lines of JSDoc for 7 lines of code, mostly the write-readback log. Keep the layout and "confirmed by write-readback", the 2500 -> 1650 -> 2500 story belongs in the commit body. Same for the backup reserve bullet above.

max246:
Thanks — all addressed in the latest push:

- **Blocker (b5 gate):** tightened `b5.length >= 11` → `=== 25`. You're right that the type-`0x04` layouts share the leading offsets, so an intermediate blob could be misread on any `AE10*`. Added a spec asserting a 12-byte type-`0x04` `b5` emits **no** `backupReserve` (and no SOC limits), so the gate is actually tested.
- **Two sources under one name:** good catch, and it's honestly unresolved. I can't disambiguate configured-vs-effective from the 0→5→15 run — if `0` was switch-off, MQTT would read 0 where HTTP reads the held value. Rather than overclaim, both JSDocs now say so: the MQTT `b5[1]` is the *realtime/effective* read, `getSafetySocParams().backupReserve` is the *configured* value (+ its switch), and "whether they agree while the switch is OFF is not yet captured." Dropped the HTTP "only source for backupReserve" line. I'll do the switch-toggle-with-value-held capture on the device to settle it and upgrade the claim in a follow-up.
- **House-style:** cut the three inline body comments and trimmed the write-readback play-by-play out of the JSDoc (kept the "confirmed by write-readback" note; the 2500→1650→2500 story is in the commit body). `0xdf` read left as-is per your note.

src/transport/mqtt/solix-mqtt.ts
```diff
@@ -304,11 +304,20 @@ export function solixReadings(frame: SolixParamFrame, productCode: string): Reco
...
- *   frame (msgtype 0x05, ~7 s) also carries a `0xb5` type-`0x04` blob, but a 25-byte one whose bytes are
- *   not the limits — hence the exact length gate.
+ * - **SOC limits** (`dischargeLimit`/`chargeLimit`, %) come from tag `0xb5`'s 4-byte SETTINGS-blob
+ *   variant — type `0x04`, payload `[discharge, output cutoff, charge]`. Confirmed live: moving discharge
+ *   10%→5% moved `b5[1]` 0x0a→0x05 while charge held at `b5[3]`=0x64.
+ * - **Backup reserve** (`backupReserve`, %) comes from tag `0xb5`'s 25-byte FAST-frame variant (type
+ *   `0x04`), which leads with `[backupReserve, discharge, charge, …]`. Confirmed by write-readback (it
+ *   tracked the app setting 0→5→15 while the min-SOC held), so it is a distinct field from the discharge
+ *   floor. The gate is exact-length-25: the 4-byte SETTINGS blob shares these offsets and is read above,
+ *   so other `b5` layouts must not fall through here. This is the realtime/effective value; the CONFIGURED
```
lenoxys: "Realtime/effective" vs "configured" is the distinction you say isn't captured yet, but both JSDocs state it as fact. Until the switch-toggle capture exists, say what is known: same name, two sources, agreement while the switch is off not yet verified. Same fix on the `getSafetySocParams` JSDoc.

lenoxys (commented):
Gate and spec fixed. The last point is the wording.

src/transport/mqtt/solix-mqtt.ts
max246: Reworded both JSDocs to what's known: same name, two sources (MQTT b5 / HTTP param 27), agreement while the switch is OFF not yet verified. No effective/configured claim.

lenoxys (approved):
All threads addressed. The b5 gate is exact, the spec pins it, and the backupReserve docs state only what is known.

